PR #2, opened 2020-07-07
Add a basic GitHub Actions CI workflow

paride:


OddBloke:
This LGTM, but I do have one request before I approve: can you push up a failing commit so we can confirm that tox failure will be appropriately surfaced when it occurs?

paride:
@OddBloke sure, but maybe there is one thing to change. The workflow now has `runs-on: ubuntu-20.04`, but maybe we should use `ubuntu-18.04` for broader compatibility, e.g. our Jenkins nodes all run Bionic at the moment. What do you think?

paride:
Job failed successfully :P

If it LGTY I'll drop the commit and force-push with `runs-on: ubuntu-18.04`.

PR #3, opened 2020-07-07
Add oracle

TheRealFalcon:
Also includes an additional commit for some small key handling refactors that I don't want squashed. Might make sense to review the two commits separately.

The API purposely doesn't completely line up with the base class (like many of the other implementing classes) but is taking the form that I would like the base class to use (in a separate soon-to-come PR). It should work as-is though.

### DIFF
--- a/docs/clouds/oci.md
+++ b/docs/clouds/oci.md
@@ -1,0 +1,63 @@
+# OCI
+
+## Credentials
+### Easy way
+Run:
+```bash
+$ pip install oci-cli
+$ oci setup config
+```
+When prompted:
+```
+location for your config: use default
+user OCID: enter your user id found on the Oracle console at Identity>>Users>>User Details
+tenancy OCID: enter your tenancy id found on the Oracle cnosole at Administration>>Tenancy Details
+region: Choose something sensible
+API Signing RSA key pair: use defaults for all prompts
+* Note this ISN'T an SSH key pair
+Follow instructions in your terminal for uploading your generated key
+```
+
+### Hard way
+
+Construct your config file manually by filling in the appropriate entries
+documented here: <br>
+https://docs.cloud.oracle.com/en-us/iaas/Content/API/Concepts/sdkconfig.htm
+
+### Compartment id
+In addition to the OCI config, the API also requires you provide the
+compartment id upon initialization. This can be found in the OCI console
+from the menu at Identity>Compartments> <your compartment>
+
+## SSH Keys
+
+OCI does not require any special key configuration. See the SSH Key page for more details
+
+## Image Lookup
+
+OCI doesn't have a concept of releases vs daily images, so both API calls refer
+to the same thing. To get the list for a release of Ubuntu:
+
+```python
+oci.released_image('focal')
+'ocid1.compartment.oc1..aaaaaaaanz4b63fdemmuag77dg2pi22xfyhrpq46hcgdd3dozkvqfzwwjwxa'
+```
+
+The returned image id can then be used for launching instances.
+
+##  Instances
+
+Launching instances requires at minimum an image_id, though instance_type
+(shape in Oracle terms) can also be specified, in addition to the other
+parameters specified by the base API.
+
+## Snapshots
+
+A snapshot of an instance is used to generate a new backing image. The generated image can in turn get used to launch new instances. This allows for customization of an image and then re-use of that image.
+
+```python
+inst = oci.launch(image_id)
+inst.execute('touch /etc/foobar')
+snapshot = oci.snapshot(instance.id)
+inst_prime = oci.launch(snapshot)
+```
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -81,6 +81,7 @@ class OCI(BaseCloud):
 
         Returns:
             string, id of latest image
+
         """
         if operating_system == 'Canonical Ubuntu':
             if not re.match(r'^\d{2}\.\d{2}$', release):  # 18.04, 20.04, etc
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -1,0 +1,226 @@
+# This file is part of pycloudlib. See LICENSE file for license information.
+"""OCI Cloud type."""
+
+import os
+import re
+import oci
+
+from pycloudlib.cloud import BaseCloud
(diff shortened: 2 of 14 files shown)


TheRealFalcon:
tox failure is mostly me forgetting how to format docstrings. I'll fix that up soon but I don't think it should block review.

pycloudlib/oracle/cloud.py
```diff
@@ -0,0 +1,218 @@
...
+        Args:
+            tag: Name of instance
+            compartment_id: A compartment found at
+                https://console.us-phoenix-1.oraclecloud.com/a/identity/compartments
+            config_path: Path of OCI config file
+        """
+        super(OCI, self).__init__(tag)
+        self.compartment_id = compartment_id
+        self.config_path = config_path
+        config = oci.config.from_file(str(config_path))
```
blackboxsw, line 34: Can we do a pre-flight if not os.exists(config_path) here and raise a RuntimeError saying "get your OCI client setup: see <LAUNCHING_ORACLE_URL>?

pycloudlib/oracle/cloud.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def delete_image(self, image_id):
+        """Delete an image.
+
+        Args:
+            image_id: string, id of the image to delete
+        """
+        self.compute_client.delete_image(image_id)
+
+    def released_image(self, release, operating_system='Canonical Ubuntu'):
```
blackboxsw, line 48: Do you have current use-cases in mind to have call-sites currently where we are supporting non-canonical Ubuntu?
Are you planning for centOS integration testing for cloud-init specifically?

pycloudlib/util.py
```diff
@@ -280,3 +280,13 @@ def _safe_int(possible_int):
         return int(possible_int)
     except (ValueError, TypeError):
         return None
+
+
+def ubuntu_name_to_number(name):
+    """Convert to version number based on the version name."""
```
blackboxsw, line 286: Any objection to a global dict? The function seems a bit over the top of a dict lookup. /me was tempted to suggest distro-info, but that's an external dep dependency. 

UBUNTU_RELEASE_VERSION_MAP = {
    'focal': '20.04',
    'bionic': '18.04',
    'xenial': '16.04',
}

pycloudlib/oracle/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    Returns:
+        The updated version of the current_data
+    """
+    for _ in range(sleep_seconds):
+        current_data = func(current_data.id).data
+        if current_data.lifecycle_state == desired_state:
+            return current_data
+        time.sleep(1)
+    raise Exception(
+        "Still not ready. "
```
blackboxsw, line 23: Static exception messages give me the grumpies. Can we parameterize the message a bit to say desired state <state> not met. Still <current_state> after waiting <sleep_seconds>"

pycloudlib/oracle/cloud.py
TheRealFalcon: No, I initially hardcoded things to be Ubuntu, but thought for a library that might be used outside of just the cloud-init project, we might need greater flexibility. Do you have a concern with doing it this way, or just curious?

pycloudlib/util.py
TheRealFalcon: Not familiar with distro-info. I'm willing to entertain an external dependency, as that can be automatically updated. I'm not really a big fan of what I did here because it means it will need to be updated after every new release. Global dict makes sense though.

TheRealFalcon:
@blackboxsw I believe I've addressed all of your comments and there are 2 open questions.

docs/clouds/oci.md
```diff
@@ -0,0 +1,63 @@
...
+```python
+oci.released_image('focal')
+'ocid1.compartment.oc1..aaaaaaaanz4b63fdemmuag77dg2pi22xfyhrpq46hcgdd3dozkvqfzwwjwxa'
+```
+
+The returned image id can then be used for launching instances.
+
+##  Instances
+
+Lauching instances requires at minimum an image_id, though instance_type
```
blackboxsw, line 50:
```suggestion
Launching instances requires at minimum an image_id, though instance_type
```

pycloudlib/util.py
blackboxsw: No need for this. If it becomes a maintenance burden we can look at using an external dynamic utility for this, but python-proper libraries or modules are probably limited here I think as most modules would want to know "what environment I'm running in" instead of "what is the known set/map of all releases available".

pycloudlib/oracle/cloud.py
blackboxsw: This was only me being curious, thanks. I wasn't sure if you already has another specific consumer in mind.

pycloudlib/oci/cloud.py
```diff
@@ -0,0 +1,227 @@
...
+
+        Args:
+            release: string, Ubuntu release to look for
+            operating_system: string, Operating system to use
+
+        Returns:
+            string, id of latest image
+
+        """
+        if operating_system == 'Canonical Ubuntu':
```
blackboxsw, line 86: Not a blocker to release of this PR, What in your mind differentiates 'Canonical Ubuntu' from just 'ubuntu'? I'm thinking about things in terms of when your followup refactor-api tries to generalize methods up in pycloudlib.cloud parent class, and ease of use when folks try to run 'ubuntu' vs. 'opensuse' 'debian' or others. 

I realize this might just be an Oracle CLI need, for which I'd say we'd probably like to standardize on something simple in the parent class, that get's translated to "Canonical Ubuntu" in the oracle api call

TheRealFalcon: This is just a consequence of the Oracle API. If I want to list Ubuntu images with the `self.compute_client.list_images(...)` call, I have to pass 'Canonical Ubuntu' as the `operating_system` argument. In my head, this was an Oracle-specific thing. In my refactor-api branch, I was expecting the method to look like `def daily_image(self, release, **kwargs)`. Since the default is defined here, that's how the API will normally function, but if you have a reason to make some Oracle-specific calls, you can override which operating_system selection.

TheRealFalcon:
@blackboxsw I fixed the typo, rebased and squashed (I don't want any additional squashing). Since I made more changes I need another +1 before it'll let me merge.

PR #4, opened 2020-07-14
ec2: add list_keys method

blackboxsw:
Add ec2.list_keys operation to allow for checking if keys exist.

It allows us to look before we leap before calling either delete_key
or upload_key.

### DIFF
--- a/pycloudlib/cloud.py
+++ b/pycloudlib/cloud.py
@@ -117,6 +117,15 @@ class BaseCloud:
         """
         raise NotImplementedError
 
+    def list_keys(self):
+        """List ssh key names present on the cloud for accessing instances.
+
+        Returns:
+           A list of strings of key pair names accessible to the cloud.
+
+        """
+        raise NotImplementedError
+
     def use_key(self, public_key_path, private_key_path=None, name=None):
         """Use an existing key.
 
--- a/pycloudlib/ec2/cloud.py
+++ b/pycloudlib/ec2/cloud.py
@@ -194,6 +194,7 @@ class EC2(BaseCloud):
         return instance
 
     def list_keys(self):
+        """List all ssh key pair names loaded on this EC2 region."""
         keypair_names = []
         for keypair in self.client.describe_key_pairs()["KeyPairs"]:
             keypair_names.append(keypair["KeyName"])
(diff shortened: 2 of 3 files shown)


paride:
@blackboxsw LGTM, but can you rebase it on master to get the CI running on it?

pycloudlib/ec2/cloud.py
```diff
@@ -193,6 +193,12 @@ def launch(self, image_id, instance_type='t2.micro', user_data=None,
 
         return instance
 
+    def list_keys(self):
```
OddBloke, line 196: Should `list_keys` this also be added to the `BaseCloud` class, as part of the standard interface that clouds are expected to implement?

OddBloke, line 196: Missing a docstring here.

paride:
Happy to see that Dan and CI agree!

pycloudlib/ec2/cloud.py
blackboxsw: @OddBloke it probably should, @TheRealFalcon and I were deciding intially to cowboy in implementations for the specific clouds in our first iterations and look to consolidated common methods after we have ec2, azure and oci in place. But, this is a fairly straightforward  method that I believe we need on most clouds.

blackboxsw: Although, that said, lxd wouldn't need this feature. But, LXD/C will be an outlier for any of the key-related methods

blackboxsw: done. pushed up a BaseCloud.list_keys. It makes sense for most clouds to support this

pycloudlib/cloud.py
```diff
@@ -117,6 +117,14 @@ def snapshot(self, instance_id, clean=True, wait=True):
         """
         raise NotImplementedError
 
+    def list_keys(self):
+        """List ssh key names present on the cloud for accessing instances.
+
+        Returns:
+           A list of strings of key pair names accessible to the cloud.
+        """
+        return NotImplementedError
```
OddBloke, line 126:
```suggestion
        raise NotImplementedError
```

paride:
Dan's approval is required because of his "request changes" review.

PR #7, opened 2020-07-21
oci: add __init__ to include oci submodule in pkging

blackboxsw:
Add __inti__.py so oci submodule is recognized in the pycloudlib package

### DIFF
--- a/pycloudlib/oci/__init__.py
+++ b/pycloudlib/oci/__init__.py
@@ -1,0 +1,2 @@
+# This file is part of pycloudlib. See LICENSE file for license information.
+"""Main OCI module __init__."""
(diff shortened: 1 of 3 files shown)


TheRealFalcon:
This shouldn't be necessary as of python 3.3:
https://www.python.org/dev/peps/pep-0420/

Is there a different reason we would need it?

blackboxsw:
> This shouldn't be necessary as of python 3.3:
> https://www.python.org/dev/peps/pep-0420/

TIL: unneeded, thanks for the link.


> Is there a different reason we would need it?

hrm something in tox virtualenv setup for our behave tests in uaclient are balking here and checking the pycloudlib package installed in the .tox env shows no pycloublib/oci subdir.
https://paste.ubuntu.com/p/VTrb2Bz5v5/

TheRealFalcon:
Interesting. Not sure why that is. I normally don't need __init__.py's anymore, but if it fixes it here, then it makes sense to add it.

PR #8, opened 2020-07-21
ec2: add wait_until_available and reload after calling create_vpc

blackboxsw:
Before we can add subnets, we need to wait until the boto3 vpc
resource is available and reloaded before trying to setup subnets as
AmazonProvidedIpv6CidrBlock=True takes a little while to get
allocated to the VPC, and subnet creation relies on the dynamic
IPv6Cidr created.


powersj:
Fix tox and I'm +1 :P

PR #9, opened 2020-07-22
Refactor api

TheRealFalcon:
The goal here is to have the base class be consistent with the implementing classes.

I've made the base cloud and instance classes officially inherit from ABC and added abstract decorators to all of the relevant methods/properties. This will give us more much detailed type/linting info and errors if we have a mismatch or unimplemented method in an implementing class.

I also changed all of the method signatures across the implementations to use the same API as the base class. Where certain APIs might request more to be passed in, the base class provides `**kwargs` that can be overridden.

Since none of the remaining TODOs are code-focused, I think this can be reviewed now.

TODO:
 * Test for every cloud. As of right now I've only verified my changes on LXD
 * Update `boot-speed/clouds/measure-cloud.py` in `server-test-scripts` right before/after this change lands.

### DIFF
--- a/examples/base_api.py
+++ b/examples/base_api.py
@@ -1,0 +1,86 @@
+"""A run through of base API operations on all supported clouds."""
+
+import os
+from contextlib import suppress
+
+import pycloudlib
+from pycloudlib.cloud import BaseCloud
+
+
+cloud_config = """#cloud-config
+runcmd:
+  - echo 'hello' > /home/ubuntu/example.txt
+"""
+
+
+def exercise_api(client: BaseCloud):
+    """Run through supported functions in the base API."""
+    try:
+        image_id = client.released_image('focal')
+    except NotImplementedError:
+        image_id = client.daily_image('focal')
+    print('focal image id: {}'.format(image_id))
+    print('launching instance...')
+    instance = client.launch(
+        image_id=image_id,
+        user_data=cloud_config
+    )
+    assert repr(instance) == repr(client.get_instance(instance.name))
+
+    print('instance name: {}'.format(instance.name))
+    with suppress(NotImplementedError):
+        print('instance ip: {}'.format(instance.ip))
+
+    print('starting instance...')
+    instance.start()
+    print('waiting for cloud-init...')
+    instance.execute('cloud-init status --wait --long')
+    with suppress(NotImplementedError):
+        instance.console_log()
+    assert instance.execute('cat /home/ubuntu/example.txt').stdout == 'hello'
+
+    print('restarting instance...')
+    instance.execute('sync')  # Prevent's some wtfs :)
+    instance.restart()
+    assert instance.execute('cat /home/ubuntu/example.txt').stdout == 'hello'
+
+    print('shutting down instance...')
+    instance.shutdown()
+    print('starting instance...')
+    instance.start()
+    print(instance.execute('cat /home/ubuntu/example.txt').stdout)
+    snapshot_id = None
+    with suppress(NotImplementedError):
+        print('snapshotting instance...')
+        snapshot_id = client.snapshot(instance)
+        print('snapshot image id: {}'.format(snapshot_id))
+    if snapshot_id:
+        assert snapshot_id != image_id
+        print('deleting snapshot...')
+        client.delete_image(snapshot_id)
+
+    print('deleting instance...')
+    instance.delete()
+
+
+clouds = {
+    pycloudlib.Azure: {},
+    pycloudlib.EC2: {},
+    # pycloudlib.GCE: {
+    #     'project': os.environ.get('PROJECT'),
+    #     'region': 'us-central1',
+    #     'zone': 'a',
+    # },
+    pycloudlib.OCI: {
+        'compartment_id': os.environ.get('COMPARTMENT_ID')
+    },
+    pycloudlib.KVM: {},
+    pycloudlib.LXD: {},
+}
+
+if __name__ == '__main__':
+    for cloud, kwargs in clouds.items():
+        print('Using cloud: {}'.format(cloud.__name__))
+        client_api = cloud(tag='base-api-test', **kwargs)
+        exercise_api(client_api)
+        print()
--- a/pycloudlib/cloud.py
+++ b/pycloudlib/cloud.py
@@ -3,13 +3,14 @@
 
 import getpass
 import logging
+from abc import ABC, abstractmethod
 
 from pycloudlib.key import KeyPair
 from pycloudlib.streams import Streams
-from pycloudlib.util import get_timestamped_tag
+from pycloudlib.util import get_timestamped_tag, validate_tag
 
 
-class BaseCloud:
+class BaseCloud(ABC):
     """Base Cloud Class."""
 
     _type = 'base'
@@ -26,8 +27,9 @@ class BaseCloud:
         self.key_pair = KeyPair(
             '/home/%s/.ssh/id_rsa.pub' % _username, name=_username
         )
-        self.tag = get_timestamped_tag(tag)
+        self.tag = validate_tag(get_timestamped_tag(tag))
 
+    @abstractmethod
     def delete_image(self, image_id):
         """Delete an image.
 
@@ -36,7 +38,8 @@ class BaseCloud:
         """
         raise NotImplementedError
 
-    def released_image(self, release):
+    @abstractmethod
+    def released_image(self, release, **kwargs):
         """ID of the latest released image for a particular release.
 
         Args:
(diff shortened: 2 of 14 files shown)


paride:
@TheRealFalcon by "update the server job" you mean the bootspeed script in the `server-test-scripts` repo?

TheRealFalcon:
@paride , the `measure-cloud.py` script, yes

pycloudlib/kvm/cloud.py
```diff
@@ -1,6 +1,7 @@
 # This file is part of pycloudlib. See LICENSE file for license information.
 """LXD Cloud type."""
```
blackboxsw, line 2:
```suggestion
"""KVM Cloud type."""
```

pycloudlib/cloud.py
```diff
@@ -76,7 +81,8 @@ def image_serial(self, image_id):
         """
         raise NotImplementedError
 
-    def get_instance(self, instance_id):
+    @abstractmethod
+    def get_instance(self, instance_id):  # () -> BaseInstance
```
lucasmoura, line 85: I don't know if we should be that flexible, but maybe we could have all the abstract methods to receive a `kwargs` parameter

pycloudlib/ec2/cloud.py
```diff
@@ -215,7 +215,7 @@ def list_keys(self):
             keypair_names.append(keypair["KeyName"])
         return keypair_names
 
-    def snapshot(self, instance, clean=True):
+    def snapshot(self, instance, clean=True, wait=True):
```
blackboxsw, line 218: Good thought in expediting pycloud lib run, this takes quite a while on ec2 sometimes.

pycloudlib/ec2/cloud.py
```diff
@@ -242,9 +242,9 @@ def snapshot(self, instance, clean=True):
         image_ami_edited = response['ImageId']
         image = self.resource.Image(image_ami_edited)
 
-        self._wait_for_snapshot(image)
-
-        instance.start(wait=True)
+        if wait:
+            self._wait_for_snapshot(image)
+            instance.start(wait=True)
```
blackboxsw, line 247: I think we need to pull this back out of if scope, otherwise we don't restart this instance in non-wait cases
```suggestion
        instance.start(wait=wait)
```

pycloudlib/kvm/cloud.py
```diff
@@ -49,48 +50,88 @@ def get_instance(self, instance_name):
...
         """Set up and launch a container.
 
         This will init and start a container with the provided settings.
         If no remote is specified pycloudlib defaults to daily images.
 
         Args:
-            name: string, what to call the instance
-            release: string, [<remote>:]image, what release to launch
-            inst_type: string, type to use
+            image_id: string, [<remote>:]image, what release to launch
```
blackboxsw, line 61: maybe worth mentioning that <remote> is either 'release' or 'daily' for multiplass/kvm

pycloudlib/kvm/cloud.py
```diff
@@ -49,48 +50,88 @@ def get_instance(self, instance_name):
         """
         return KVMInstance(instance_name)
 
-    def launch(self, name, release, inst_type=None, wait=True):
```
blackboxsw, line 52: Given that you are diverging from distribution/release aliases for a given cloud to a specific launch by image_id approach, how do you suggest we handle these cloud-specific lookups of proper image_id aliases for a given Ubuntu release?

pycloudlib/ec2/instance.py
```diff
@@ -16,12 +16,12 @@ class EC2Instance(BaseInstance):
 
     _type = 'ec2'
 
-    def __init__(self, client, key_pair, instance):
+    def __init__(self, key_pair, client, instance):
```
blackboxsw, line 19: Given that you are switch order of positional params, I think we need to fix ec2/cloud,py.launch() to properly init Instance now. Also it would be good during that 'fix' to provide named parameters as kwargs for anything we touch so that future refactors don't get bit by positional param re-ordering.

pycloudlib/kvm/cloud.py
TheRealFalcon: It would be in the implementation of `released_image` or `daily_image`. I.e.,

```python
instance = client.launch(image_id=client.released_image('focal'))```

**Comment by TheRealFalcon** [comment]
> A question about launch how do you propose we will introspect default image IDs on a given cloud for each ubuntu series?

Is this the same as the question you asked inline, or are you asking something else here?

File: pycloudlib/ec2/instance.py (hunk 7 above)
Comment (TheRealFalcon) [reply pycloudlib/ec2/instance.py:19]: If it's part of the base API, I'm actually with not providing default args. The only reason I'm doing this repositioning of args now is because the different APIs weren't consistent with the base. Once everything is consistent there shouldn't be a need for anymore re-orders, and if new parameters are necessary, those can be introduced as kwargs. But I see no need to convert existing args to kwargs.

File: pycloudlib/ec2/cloud.py (hunk 4 above)
Comment (TheRealFalcon) [reply pycloudlib/ec2/cloud.py:247]: We're stopping the image to make a snapshot, so I'm assuming we can't start the instance while the snapshot is still being created. I figured if you didn't want to wait you'd just restart the instance manually, but now that I'm thinking about it again, that's not going to be super obvious. We could spawn a background thread to start it once the snapshot is created, but not sure that's a great idea anyway. There's also the option of status callbacks...

Maybe snapshot shouldn't support a wait argument? Or maybe we keep the argument in, but for now only provide the waiting code and raise a `NotImplementedError` if `wait=False` until we decide to get around to it?

File: pycloudlib/cloud.py (hunk 2 above)
Comment (TheRealFalcon) [reply pycloudlib/cloud.py:85]: I'd default to not providing that now, just because adding them when we need to shouldn't require any additional pain and it keeps the calls a little cleaner and easier to understand if we never need them.

File: pycloudlib/ec2/cloud.py (hunk 4 above)
Comment (blackboxsw) [reply pycloudlib/ec2/cloud.py:247]: @TheRealFalcon I see you implemented that logic in OCI, harm I'd lean toward either raising NoImplementedError on wait=False for all other pycloud subclasses or not supporting wait param on any and just waiting if you are creating an image from a running instance. In most cloud cases, the source instance from which we are creating a snapshot will be brought down anyway, and bringing it up mid-image creation is likely to cause issues anyway I'd guess.

File: pycloudlib/kvm/cloud.py (hunk 6 above)
Comment (blackboxsw) [reply pycloudlib/kvm/cloud.py:52]: ahh right. Thanks, brain fail.+1

File: pycloudlib/ec2/instance.py (hunk 7 above)
Comment (blackboxsw) [reply pycloudlib/ec2/instance.py:19]: That makes sense thanks.

File: examples/base_api.py [hunk 8]
```diff
@@ -0,0 +1,86 @@
...
+        client.delete_image(snapshot_id)
+
+    print('deleting instance...')
+    instance.delete()
+
+
+clouds = {
+    pycloudlib.Azure: {},
+    pycloudlib.EC2: {},
+    # pycloudlib.GCE: {
```
TheRealFalcon, line 69: This will be uncommented in a follow-on PR with google instance support

TheRealFalcon:
When this merges we'll also need to update the server-test-scripts repo with https://github.com/canonical/server-test-scripts/pull/41

examples/base_api.py
```diff
@@ -0,0 +1,86 @@
...
+    assert repr(instance) == repr(client.get_instance(instance.name))
+
+    print('instance name: {}'.format(instance.name))
+    with suppress(NotImplementedError):
+        print('instance ip: {}'.format(instance.ip))
+
+    print('starting instance...')
+    instance.start()
+    print('waiting for cloud-init...')
+    instance.execute('cloud-init status --wait --long')
```
lucasmoura, line 37: Maybe we could call `instance.wait()`  here directly, since most instance types do implement the `cloud-init` check in the wait method

examples/base_api.py
```diff
@@ -0,0 +1,86 @@
...
+        snapshot_id = client.snapshot(instance)
+        print('snapshot image id: {}'.format(snapshot_id))
+    if snapshot_id:
+        assert snapshot_id != image_id
+        print('deleting snapshot...')
+        client.delete_image(snapshot_id)
+
+    print('deleting instance...')
+    instance.delete()
+
```
lucasmoura, line 64: Maybe we could further improve the test by launching a vm from the `snapshot_id`

lucasmoura:
@TheRealFalcon I could not spot anything wrong with the code, just some ideas to add to the `base_api` script, which will be really useful for testing new features

paride (approved):

paride:
I updated the branch but now CI is failing. Reason is we don't have a *versioned* dependency on the linters. Let me fix that, I really want to see this land!

PR #10, opened 2020-07-22
Add Azure support

lucasmoura:
This PR adds azure support to pycloudlib.

Right now, there are still some things to be completed:

* Define how to handle pro images

I will keep working on these details, but I think it is worth to collect feedback on the implementation I have so far.

Furthermore, this is test script I have been using to the test this new feature:

```python
import logging                                                                                                                                                                                                     
                                                                                                                                                                                                                   
from pycloudlib.azure.cloud import Azure                                                                                                                                                                           
                                                                                                                                                                                                                   
                                                                                                                                                                                                                   
logging.getLogger("adal-python").setLevel(logging.WARNING)                                                                                                                                                         

user_data = """ 
#cloud-config
apt:
  - arches:
        - default
      search:
        - http://uk.archive.ubuntu.com/ubuntu/
"""
tenant_id = "TENANT_ID"
client_id = "CLIENT_ID"
client_secret_id = "CLIENT_SECRET"
subscription_id = "SUBSCRIPTION_ID"

test = Azure(
    tag="azure",
    client_id=client_id,
    client_secret_id=client_secret_id,
    tenant_id=tenant_id,
    subscription_id=subscription_id
)

image_id = test.daily_image(release="focal")
pub_key, priv_key = test.create_key_pair(key_name="test_integration")
print(test.list_keys())

pub_path = "pub_test.pem"
priv_path = "priv_test.pem"

with open(pub_path, "w") as f:
    f.write(pub_key)

with open(priv_path, "w") as f:
    f.write(priv_key)
test.use_key(pub_path, priv_path)

vm_instance = test.launch(image_id=image_id, user_data=user_data)
vm_instance.delete()
vm_instance = test.launch(image_id=image_id, user_data=user_data)
print(vm_instance.execute('lsb_release -a'))
print(vm_instance.execute('sudo touch /etc/foobar'))
vm_instance.restart()
print(vm_instance.execute('echo restarted'))
print(vm_instance.instance_name)

image_id_snap = test.snapshot(vm_instance)
vm_instance_snap = test.launch(image_id=image_id_snap, user_data=user_data)
print(vm_instance_snap.execute("lsb_release -a"))
vm_instance_snap.shutdown(wait=True)
test.delete_image(image_id=image_id_snap)
vm_instance_snap.start(wait=True)

print(vm_instance_snap.instance_name)
print(vm_instance_snap.execute('file /etc/foobar'))

retrieved_vm_instance = test.get_instance(
    instance_id=vm_instance_snap.instance_name,
    search_all=True
)
retrieved_vm_instance.execute('echo retrived')
retrieved_vm_instance.delete()

test.delete_key("test_integration")
print(test.list_keys())
test.delete_resource_group()
```

### DIFF
--- a/docs/clouds/azure.md
+++ b/docs/clouds/azure.md
@@ -1,0 +1,118 @@
+# Azure
+
+The following page documents the Azure cloud integration in pycloudlib.
+
+## Credentials
+
+To access Azure via the API requires users to have four different keys:
+
+* client id
+* client secret id
+* tenant id
+* subscription id
+
+To obtain this info, there are two ways supported in pycloudlib
+
+### Azure login
+
+By using the Azure CLI, you can login into your Azure account through it. Once you logged in, the
+CLI will create folder in your home directory which will contain all of the necessary information
+to use the API. To login into you Azure using the CLI, just run the following command:
+
+```shell
+az login
+```
+
+### Passed Directly
+
+All of these four credentials can also be provided directly when initializing the Azure object:
+
+```python
+azure = pycloudlib.Azure(
+    client_id='ID_VALUE',
+    client_secret_id='ID_VALUE',
+    tenant_id='ID_VALUE',
+    subscription_id='ID_VALUE',
+)
+```
+
+This way we can create different Azure instances with different configurations.
+
+## SSH Keys
+
+Azure requires an SSH key to be uploaded before using it. See the SSH Key page for more details.
+
+## Image Lookup
+
+To find latest daily Azure image for a release of Ubuntu:
+
+```python
+azure.daily_image('trusty')
+"Canonical:UbuntuServer:14.04.0-LTS",
+```
+
+The return Azure image can then be used for launching instances.
+
+## Instances
+
+Launching an instance requires at a minimum an Azure image.
+
+```python
+inst_0 = azure.launch('Canonical:UbuntuServer:14.04.0-LTS')
+inst_1 = azure.launch('Canonical:UbuntuServer:18.04-DAILY-LTS')
+```
+
+If further customization of an instance is required, a user can pass additional arguments to the launch command and have them passed on.
+
+```python
+inst = azure.launch(
+    image_id='Canonical:UbuntuServer:14.04.0-LTS',
+    user_data='#cloud-config\nfinal_message: "system up!"',
+)
+```
+
+By default, the launch method will wait for cloud-init to finish initializing before completing. When launching multiple instances a user may not wish to wait for each instance to come up by passing the `wait=False` option.
+
+```python
+instances = []
+for inst in range(num_instances):
+    instances.append(
+        azure.launch('Canonical:UbuntuServer:18.04-DAILY-LTS', wait=False))
+
+for instance in instances:
+    instance.wait()
+```
+
+Similarly, when deleting an instance, the default action will wait for the instance to complete termination. Otherwise, the `wait=False` option can be used to start the termination of a number of instances:
+
+```python
+inst.delete()
+
+for instance in instances:
+    instance.delete(wait=False)
+```
+
+An existing instance can get used by providing an instance-id.
+
+```python
+instance = azure.get_instance('my-azure-vm')
+```
+
+## Snapshots
+
+A snapshot of an instance is used to generate a new backing Azure image. The generated image can in turn get used to launch new instances. This allows for customization of an image and then re-use of that image.
+
+```python
+inst = azure.launch('Canonical:UbuntuServer:14.04.0-LTS')
+inst.execute('touch /etc/foobar')
+image_id_snapshot = azure.snapshot(inst)
+inst_prime = azure.launch(image_id_snapshot)
+```
+
+The snapshot function returns a string of the created AMI ID.
+
+To delete the image when the snapshot is no longer required:
+
+```python
+azure.image_delete(image_id_snapshot)
+```
--- a/examples/az.py
+++ b/examples/az.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+# This file is part of pycloudlib. See LICENSE file for license information.
+"""Basic examples of various lifecycle with an Azure instance."""
+
+import logging
+
+import pycloudlib
+
+
+cloud_config = """#cloud-config
+runcmd:
+  - echo 'hello' > /home/ubuntu/example.txt
+"""
+
+
+def demo():
+    """Show example of using the Azure library.
+
+    Connects to Azure and launches released image. Then runs
+    through a number of examples.
+
+    PS: we assume in this example that you are logged into
+    you Azure account
+    """
+    client = pycloudlib.Azure(tag='azure')
+
+    image_id = client.daily_image(release="focal")
+    pub_key, priv_key = client.create_key_pair(
+        key_name="test_integration")
+
+    pub_path = "pub_test.pem"
+    priv_path = "priv_test.pem"
+
+    with open(pub_path, "w") as f:
+        f.write(pub_key)
+
+    with open(priv_path, "w") as f:
+        f.write(priv_key)
+    client.use_key(pub_path, priv_path)
+
+    instance = client.launch(
+        image_id=image_id,
+        user_data=cloud_config
+    )
+
+    print(instance.ip)
+    instance.wait()
+    print(instance.execute('cat /home/ubuntu/example.txt'))
+
+    snapshotted_image_id = client.snapshot(instance)
+    instance.delete()
+
+    new_instance = client.launch(image_id=snapshotted_image_id)
+    new_instance.delete()
+
+
+if __name__ == '__main__':
+    # Avoid polluting the log with azure info
+    logging.getLogger("adal-python").setLevel(logging.WARNING)
+    logging.basicConfig(level=logging.DEBUG)
+    demo()
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -1,0 +1,820 @@
+# This file is part of pycloudlib. See LICENSE file for license information.
+"""Azure Cloud type."""
+import base64
+
+from azure.mgmt.resource import ResourceManagementClient
+from azure.mgmt.network import NetworkManagementClient
+from azure.mgmt.compute import ComputeManagementClient
+
+import pycloudlib.azure.util as util
+
+from pycloudlib.cloud import BaseCloud
+from pycloudlib.azure.instance import AzureInstance
+from pycloudlib.key import KeyPair
+from pycloudlib.util import get_timestamped_tag
+
+
+class Azure(BaseCloud):
+    """Azure Cloud Class."""
+
+    _type = 'azure'
+
+    UBUNTU_RELEASE = {
+        "trusty": "Canonical:UbuntuServer:14.04.0-LTS",
+        "xenial": "Canonical:UbuntuServer:16.04-DAILY-LTS",
+        "bionic": "Canonical:UbuntuServer:18.04-DAILY-LTS",
+        "focal": "Canonical:0001-com-ubuntu-server-focal-daily:20_04-daily-lts"
+    }
+
+    def __init__(self, tag, client_id=None, client_secret=None,
+                 subscription_id=None, tenant_id=None):
+        """Initialize the connection to Azure.
+
+        Azure will try to read user credentials form the /home/$USER/.azure
+        folder. However, we can overwrite those credentials with the provided
+        id parameters.
+
+        Args:
+            tag: string used to name and tag resources with
+            client_id: user's client id
+            client_secret: user's client secret access key
+            subscription_id: user's subscription id key
+            tenant_id: user's tenant id key
+        """
+        super().__init__(tag)
+        self._log.debug('logging into Azure')
+        self.location = "centralus"
+        self.username = "ubuntu"
+
+        self.registered_instances = {}
+        self.registered_images = {}
+
+        config_dict = {}
+
+        if client_id:
+            config_dict["clientId"] = client_id
+
+        if client_secret:
+            config_dict["clientSecret"] = client_secret
+
+        if subscription_id:
+            config_dict["subscriptionId"] = subscription_id
+
+        if tenant_id:
+            config_dict["tenantId"] = tenant_id
+
+        self.resource_client = util.get_client(
+            ResourceManagementClient, config_dict
+        )
+
+        self.network_client = util.get_client(
+            NetworkManagementClient, config_dict
+        )
+
+        self.compute_client = util.get_client(
+            ComputeManagementClient, config_dict
+        )
+
+        self.resource_group = self._create_resource_group()
+        self.base_tag = tag
+
+    def image_serial(self, image_id):
+        """Find the image serial of the latest daily image for a particular release.
+
+        Args:
+            image_id: string, Ubuntu image id
+
+        Returns:
+            string, serial of latest image
+
+        """
+        raise NotImplementedError
+
+    def _create_network_security_group(self):
+        """Create a network security group.
+
+        This method creates a network security groups that allows the user
+        to ssh into the machine and execute commands.
+
+        Returns:
+            The network security object created by Azure
+
+        """
+        security_group_name = "{}-sgn".format(self.tag)
+        nsg_group = self.network_client.network_security_groups
+
+        self._log.debug('Creating Azure network security group')
+        nsg_call = nsg_group.create_or_update(
+            resource_group_name=self.resource_group.name,
+            network_security_group_name=security_group_name,
+            parameters={
+                "location": self.location,
+                "security_rules": [
+                    {
+                        "name": "SSH",
+                        "properties": {
+                            "priority": 300,
+                            "protocol": "TCP",
+                            "access": "Allow",
+                            "direction": "Inbound",
+                            "sourceAddressPrefix": "*",
+                            "sourcePortRange": "*",
+                            "destinationAddressPrefix": "*",
+                            "destinationPortRange": "22"
+                        }
+                    }
+                ]
+            }
+        )
+
+        return nsg_call.result()
+
+    def _create_resource_group(self):
+        """Create a resource group.
+
+        This method creates an Azure resource group. Every other component that
+        we create will be contained into this resource group. This means that
+        if we delete this resource group, we delete all resources associated
+        with it.
+
+        Returns:
+            The resource group created by Azure
+
+        """
+        resource_name = "{}-rg".format(self.tag)
+        self._log.debug('Creating Azure resource group')
+
+        return self.resource_client.resource_groups.create_or_update(
+            resource_name,
+            {
+                "location": self.location,
+                "tags": {
+                    "name": self.tag
+                }
+            }
+        )
+
+    def _create_virtual_network(self, address_prefixes=None):
+        """Create a virtual network.
+
+        This method creates an Azure virtual network to be used
+        when provisioning a subnet.
+
+        Args:
+            address_prefixes:  list of strings, A list of address prefixes
+                               to be used in this virtual network.
+        Returns:
+            The virtual network created by Azure
+
+        """
+        if address_prefixes is None:
+            address_prefixes = ["10.0.0.0/16"]
+
+        virtual_network_name = "{}-vnet".format(self.tag)
+
+        self._log.debug('Creating Azure virtual network')
+        network_call = self.network_client.virtual_networks.create_or_update(
+            self.resource_group.name,
+            virtual_network_name,
+            {
+                "location": self.location,
+                "address_space": {
+                    "address_prefixes": address_prefixes
+                },
+                "tags": {
+                    "name": self.tag
+                }
+            }
+        )
+
+        return network_call.result()
+
+    def _create_subnet(self, vnet_name, address_prefix="10.0.0.0/24"):
+        """Create a subnet.
+
+        This method creates an Azure subnet to be used when
+        provisioning a network interface.
+
+        Args:
+            address_prefix: string, An address prefix to be used for
+                            this subnet.
+
+        Returns:
+            The subnet created by Azure
+
+        """
+        subnet_name = "{}-subnet".format(self.tag)
+
+        self._log.debug('Creating Azure subnet')
+        subnet_call = self.network_client.subnets.create_or_update(
+            self.resource_group.name,
+            vnet_name,
+            subnet_name,
+            {
+                "address_prefix": address_prefix,
+                "tags": {
+                    "name": self.tag
+                }
+            }
+        )
+
+        return subnet_call.result()
+
+    def _create_ip_address(self):
+        """Create an ip address.
+
+        This method creates an Azure ip address to be used when
+        provisioning a network interface
+
+        Returns:
+            The ip address created by Azure
+
+        """
+        ip_name = "{}-ip".format(self.tag)
+
+        self._log.debug('Creating Azure ip address')
+        ip_call = self.network_client.public_ip_addresses.create_or_update(
+            self.resource_group.name,
+            ip_name,
+            {
+                "location": self.location,
+                "sku": {"name": "Standard"},
+                "public_ip_allocation_method": "Static",
+                "rpublic_ip_address_version": "IPV4",
+                "tags": {
+                    "name": self.tag
+                }
+            }
+        )
+
+        return ip_call.result()
+
+    def _create_network_interface_client(self, ip_address_id, subnet_id,
+                                         nsg_id):
+        """Create a network interface client.
+
+        This method creates an Azure network interface to be used when
+        provisioning a virtual machine
+
+        Args:
+            ip_address_id: string, The ip address id
+            subnet_id: string, the subnet id
+            nsg_id: string, the network security group id
+
+        Returns:
+            The ip address created by Azure
+
+        """
+        nic_name = "{}-nic".format(self.tag)
+        ip_config_name = "{}-ip-config".format(self.tag)
+
+        self._log.debug('Creating Azure network interface')
+        nic_call = self.network_client.network_interfaces.create_or_update(
+            self.resource_group.name,
+            nic_name,
+            {
+                "location": self.location,
+                "ip_configurations": [
+                    {
+                        "name": ip_config_name,
+                        "subnet": {
+                            "id": subnet_id
+                        },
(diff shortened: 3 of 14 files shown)


blackboxsw:
@lucasmoura Let's add the following to pycloudlib/__init__.py
```diff
diff --git a/pycloudlib/__init__.py b/pycloudlib/__init__.py
index 6c8de09..dd48e25 100644
--- a/pycloudlib/__init__.py
+++ b/pycloudlib/__init__.py
@@ -3,6 +3,7 @@
 
 import logging
 
+from pycloudlib.azure.cloud import Azure
 from pycloudlib.ec2.cloud import EC2
 from pycloudlib.gce.cloud import GCE
 from pycloudlib.lxd.cloud import LXD
@@ -10,6 +11,7 @@ from pycloudlib.kvm.cloud import KVM
 from pycloudlib.oci.cloud import OCI
 
 __all__ = [
+    'Azure',
     'EC2',
     'GCE',
     'LXD',
```

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,639 @@
...
+        "xenial": "Canonical:UbuntuServer:16.04-DAILY-LTS",
+        "bionic": "Canonical:UbuntuServer:18.04-DAILY-LTS",
+        "focal": "Canonical:0001-com-ubuntu-server-focal-daily:20_04-daily-lts"
+    }
+
+    def __init__(self, tag, client_id=None, client_secret_id=None,
+                 subscription_id=None, tenant_id=None):
+        """Initialize the connection to Azure.
+
+        Azure will try to read user credentials form the /home/$USER/.azure
```
blackboxsw, line 33: If client_secret_id, subscription_id or tenant_id are None, let's try to fallback explicitly check for presence of this `~/.azure/credentials.json` config file, raising a RuntimeError if absent with the approriate message that "Azure cloud credentials don't seem to be available. Run: az login or provide client_secret_id and subscription_id to pycloudlib.Azure"

or raise a RuntimeError if not os.path.exists(cred_file = os.path.expanduser('~/.azure/credentials.json')

lucasmoura: Weird, even being logged into Azure I do not have that file. But since we are calling the `get_client_from_cli_profile`, should we skip this file check and just verify if any of the id parameters are None to raise the exception ?

docs/clouds/azure.md
```diff
@@ -0,0 +1,118 @@
...
+
+```python
+inst_0 = azure.launch('Canonical:UbuntuServer:14.04.0-LTS')
+inst_1 = azure.launch('Canonical:UbuntuServer:18.04-DAILY-LTS')
+```
+
+If further customization of an instance is required, a user can pass additional arguments to the launch command and have them passed on.
+
+```python
+inst = ec2.launch(
```
blackboxsw, line 68:
```suggestion
inst = azure.launch(
```

lucasmoura: Fixed

pycloudlib/azure/cloud.py
blackboxsw: Good pt @lucasmoura. I think it may make sense then to minimally _log.debug("No AZ client  configuration found') when get_client_from_cli_profile raises CLIError

And we should check the known mandatory config_dict values before trying to fallthrough to get_client_from_json_dict
as the errors we get in that case from azure's libs are annoying to decrypt.


Here's a failure I see on a system with no azure cli config
```
csmith@uptown:~/src/pycloudlib$  cat > 1.py <<EOF
import logging
import sys

from pycloudlib import Azure

logging.basicConfig(filename="azure-test.log", level=logging.DEBUG)
azure = Azure(tag='uaclient-integration')
EOF

csmith@uptown:~/src/pycloudlib$  tox -r -e pylint
csmith@uptown:~/src/pycloudlib$ .tox/pylint/bin/python3 ./1.py
csmith@uptown:~/src/pycloudlib$ .tox/pylint/bin/python3 ./1.py 
  File "./1.py", line 7, in <module>
    azure = Azure(tag='uaclient-integration')
  File "/home/csmith/src/pycloudlib/pycloudlib/azure/cloud.py", line 66, in __init__
    self.resource_client = util.get_client(
  File "/home/csmith/src/pycloudlib/pycloudlib/azure/util.py", line 49, in get_client
    client = get_client_from_json_dict(
  File "/home/csmith/src/pycloudlib/.tox/pylint/lib/python3.8/site-packages/azure/common/client_factory.py", line 182, in get_client_from_json_dict
    authority_url = authority_url + '/' + config_dict['tenantId']
TypeError: can only concatenate str (not "NoneType") to str
```



Here's a patch of some of what I was thinking (though you'll see it conflicts with some individual comments I'll have in a minute or two
```diff
diff --git a/pycloudlib/azure/util.py b/pycloudlib/azure/util.py
index 378ca51..e33baf4 100644
--- a/pycloudlib/azure/util.py
+++ b/pycloudlib/azure/util.py
@@ -3,6 +3,9 @@
 from azure.common.client_factory import (get_client_from_cli_profile,
                                          get_client_from_json_dict)
 from knack.util import CLIError
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 def get_client(resource, config_dict):
@@ -24,34 +27,41 @@ def get_client(resource, config_dict):
 
     """
     try:
-        client = get_client_from_cli_profile(resource)
-    except CLIError:
-        client = None
-
-    if client is None:
-        parameters = {
-            "subscriptionId": config_dict.get("subscription_id"),
-            "tenantId": config_dict.get("tenant_id"),
-            "clientId": config_dict.get("client_id"),
-            "clientSecret": config_dict.get("client_secret"),
-            "activeDirectoryEndpointUrl":
-                "https://login.microsoftonline.com",
-            "resourceManagerEndpointUrl": "https://management.azure.com/",
-            "activeDirectoryGraphResourceId":
-                "https://graph.windows.net/",
-            "sqlManagementEndpointUrl":
-                "https://management.core.windows.net:8443/",
-            "galleryEndpointUrl": "https://gallery.azure.com/",
-            "managementEndpointUrl":
-                "https://management.core.windows.net/"
-        }
-
-        client = get_client_from_json_dict(
-            resource,
-            parameters
+        return get_client_from_cli_profile(resource)
+    except Exception:
+        logger.debug(
+            "No valid azure-cli config found. Trying explicit config params"
         )
-
-    return client
+    required_keys = frozenset({"tenant_id"})
+    import pdb; pdb.set_trace()
+    missing_keys = required_keys.difference(set(config_dict.keys()))
+    if missing_keys:
+        raise RuntimeError(
+            "No AZ cli config found, missing required keys: {}".format(
+                ", ".join(missing_keys)
+            )
+        )
+    parameters = {
+        "subscriptionId": config_dict.get("subscription_id"),
+        "tenantId": config_dict.get("tenant_id"),
+        "clientId": config_dict.get("client_id"),
+        "clientSecret": config_dict.get("client_secret"),
+        "activeDirectoryEndpointUrl":
+            "https://login.microsoftonline.com",
+        "resourceManagerEndpointUrl": "https://management.azure.com/",
+        "activeDirectoryGraphResourceId":
+            "https://graph.windows.net/",
+        "sqlManagementEndpointUrl":
+            "https://management.core.windows.net:8443/",
+        "galleryEndpointUrl": "https://gallery.azure.com/",
+        "managementEndpointUrl":
+            "https://management.core.windows.net/"
+    }
+
+    return get_client_from_json_dict(
+        resource,
+        parameters
+    )
 
 
 def get_offer_from_image_id(image_id):
```

examples/az.py
```diff
@@ -0,0 +1,61 @@
...
+        f.write(priv_key)
+    client.use_key(pub_path, priv_path)
+
+    instance = client.launch(
+        image_id=image_id,
+        user_data=cloud_config
+    )
+
+    print(instance.ip)
+    instance.execute('cloud-init status --wait --long')
```
blackboxsw, line 47: Shouldn't this example just be instance.wait() as it does exacly that cloud-init status wait?

pycloudlib/azure/util.py
```diff
@@ -0,0 +1,88 @@
...
+        config_dict: dict, Id parameters passed by the user to this class.
+
+    Returns:
+        The client for the resource passed as parameter.
+
+    """
+    try:
+        client = get_client_from_cli_profile(resource)
+    except CLIError:
+        client = None
```
blackboxsw, line 29: We should just early exit on success and log a debug message on exception. This way we avoid indenting all the rest of the code below when we fall through on error.

```suggestion
        client = get_client_from_cli_profile(resource)
    except CLIError:
        logger.debug("No azure-cli configuration found. Relying on explicit pycloudlib params")
```

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+        self.location = "centralus"
+        self.username = "ubuntu"
+
+        self.registered_instances = {}
+        self.registered_images = {}
+
+        config_dict = {}
+
+        if client_id:
+            config_dict["client_id"] = client_id
```
blackboxsw, line 55: When setting up config_dict here, let's either:
 1.  actually use the expected camelCaseKeyNames (like clientId, subscriptionId)  so we don't have to remap them in azure.util.get_client
   OR 
 2. pass those client_id, subscription_id params directly into the call signature of get_client since that's the only information we are providing via config_dict. 

Ultimately what this init translates into is either you need az cli config available, or you have to provide all of client_id, subscription_id, client_secret_id, and tenant_id right? so in either get_client , or AzureCloud we should check that config condition and raise an appropriate error.

examples/az.py
lucasmoura: Yes, I forgot about the `wait` method here :facepalm:

pycloudlib/azure/cloud.py
lucasmoura: Yes, that's a great solution. I will add it to the code. Thanks @blackboxsw

pycloudlib/azure/cloud.py
lucasmoura: I like the first idea better, since I think it is cleaner to check if any of the variables is None using the set logic. I will add it to the code

lucasmoura:
@blackboxsw I have updated the code

examples/az.py
```diff
@@ -0,0 +1,61 @@
...
+
+    snapshotted_image_id = client.snapshot(instance)
+    instance.delete()
+
+    new_instance = client.launch(image_id=snapshotted_image_id)
+    new_instance.delete()
+
+
+if __name__ == '__main__':
+    # Avoid poluting the log with azure info
```
blackboxsw, line 58:
```suggestion
    # Avoid polluting the log with azure info
```

examples/az.py
```diff
@@ -0,0 +1,61 @@
...
+def demo():
+    """Show example of using the Azure library.
+
+    Connects to Azure and launches released image. Then runs
+    through a number of examples.
+
+    PS: we assume in this example that you are logged into
+    you Azure account
+    """
+    client = pycloudlib.Azure('azure')
```
blackboxsw, line 25: probably worth calling out the param name for clarity
```suggestion
    client = pycloudlib.Azure(tag='demo-tag')
```

docs/clouds/azure.md
```diff
@@ -0,0 +1,118 @@
...
+inst.delete()
+
+for instance in instances:
+    instance.delete(wait=False)
+```
+
+An existing instance can get used by providing an instance-id.
+
+```python
+instance = azure.get_instance('Canonical:UbuntuServer:18.04-DAILY-LTS')
```
blackboxsw, line 98: This string looks more like an image alias than an instance_id, can we provide an example like either of the following:
 
   azure.get_instance('"my-instance-name")

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+        Returns:
+            The resource group created by Azure
+
+        """
+        resource_name = "{}-rg".format(self.tag)
+        self._log.debug('Creating Azure resource group')
+
+        return self.resource_client.resource_groups.create_or_update(
+            resource_name,
+            {
```
blackboxsw, line 137: Can we provide tags here too so each resource we create is tagged the same (it aids in finding and removing separate resources in the future)

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+            The virtual network created by Azure
+
+        """
+        if address_prefixes is None:
+            address_prefixes = ["10.0.0.0/16"]
+
+        virtual_network_name = "{}-vnet".format(self.tag)
+
+        self._log.debug('Creating Azure virtual network')
+        network_call = self.network_client.virtual_networks.create_or_update(
```
blackboxsw, line 161: Same question for any resource we create or update, we should probably add tags if we can.

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+        Args:
+            release: string, Ubuntu release to look for
+        Returns:
+            string, id of latest image
+
+        """
+        return self.daily_image(release)
+
+    def daily_image(self, release):
+        """Find the iamge info for the latest daily image for a given release.
```
blackboxsw, line 418:
```suggestion
        """Find the image info for the latest daily image for a given release.
```

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+        else:
+            self._log.debug(
+                'Error deleting %s. Request returned %d',
+                image_id, resp_code
+            )
+
+    def released_image(self, release):
+        """Get the released image.
+
+        With the way we are indexing our imaages, it is hard to differentiate
```
blackboxsw, line 404:
```suggestion
        With the way we are indexing our images, it is hard to differentiate
```

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+        between daily and released images, since we would need to have the
+        version of the image to properly provision it. Due to that limitatuin
```
blackboxsw, line 406:
```suggestion
        version of the image to properly provision it. Due to that limitation
```

pycloudlib/azure/util.py
```diff
@@ -0,0 +1,102 @@
...
+    missing_keys = required_keys.difference(set(config_dict.keys()))
+    if missing_keys:
+        raise RuntimeError(
+            "No AZ cli config found, missing required keys: {}".format(
+                ", ".join(missing_keys)
+            )
+        )
+
+    parameters = {
+        **config_dict,
```
blackboxsw, line 50: This merge will work unless someone provides and override of activeDirectoryEndpointUrl in config_dict right?

I think we still need parameters.update(config_dict) as config_dict should override local defaults if desired.

If the following were the case callers of  get_client would not be able to override any of the default params if desired.
>>> config_dict = {'a': 'override_a'}
>>> parameters = {**config_dict, 'a': 'default_a'}
>>> parameters
{'a': 'default_a'}

pycloudlib/util.py
```diff
@@ -287,3 +288,18 @@ def _safe_int(possible_int):
         return int(possible_int)
     except (ValueError, TypeError):
         return None
+
+
+def create_tag(tag):
```
blackboxsw, line 293: good idea, can we also provide a strict=False param or time_suffix=True or something else to allow avoiding appending the datetime suffix

pycloudlib/azure/util.py
```diff
@@ -0,0 +1,102 @@
+# This file is part of pycloudlib. See LICENSE file for license information.
+"""Azure Util Functions."""
+import logging
+
+from azure.common.client_factory import (get_client_from_cli_profile,
+                                         get_client_from_json_dict)
+from knack.util import CLIError
```
blackboxsw, line 7: I think we now need to add a pinned version of knack to our requirements.txt right?

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+            client_secret: user's client secret access key
+            subscription_id: user's subscription id key
+            tenant_id: user's tenant id key
+        """
+        super().__init__(tag)
+        self._log.debug('logging into Azure')
+        self.location = "centralus"
+        self.username = "ubuntu"
+
+        self.registered_instances = {}
```
blackboxsw, line 49: I don't think this approach will work for other pycloudlib.Azure users as this only gets set as you launch instances on this particular run of pycloudlib.Azure. We also want to be able to query existing instances in the cloud (not launched by this particular instance of pycloudlib). I think we should use the compute api for this compute_client.virtual_machines.list(resource_group_name='blah')

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+                ]
+            }
+        )
+
+        return nsg_call.result()
+
+    def _create_resource_group(self):
+        """Create a resource group.
+
+        This method creates an Azure resource group. Every other componet that
```
blackboxsw, line 123:
```suggestion
        This method creates an Azure resource group. Every other component that
```

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,644 @@
...
+        """Get an instance by id.
+
+        Args:
+            instance_id: the instance name to search by
+
+        Returns:
+            An instance object to use to manipulate the instance further.
+
+        """
+        if instance_id in self.registered_instances:
```
blackboxsw, line 590: This really needs to be calling azure.compute_client.virtual_machines.list or .get(vm_name='blah') api.
What this may mean is that we have to either provide an optional resource_group parameter to search for the named instance in your desired groupname

examples/az.py
lucasmoura: Fixed

examples/az.py
lucasmoura: Updated

docs/clouds/azure.md
lucasmoura: Yes, that was a mistake. Solved now

pycloudlib/azure/cloud.py
lucasmoura: Added tags to all resources

pycloudlib/azure/cloud.py
lucasmoura: Fixed

pycloudlib/azure/cloud.py
lucasmoura: Fixed

pycloudlib/azure/cloud.py
lucasmoura: Fixed

pycloudlib/azure/util.py
lucasmoura: Although we are not currently supporting overriding these fields, you are right that this will pose a problem if we want to support that. I will update the code

pycloudlib/azure/cloud.py
lucasmoura: Yes, you are right. I was thinking about the resources only allocated by the instance, but we should think globally here. I will update the code

lucasmoura:
@blackboxsw I have updated the PR. I have created a mixed implementation for the `get_instance` method. I am still using the `referenced_images`, but now the user is allowed to perform a global search for instances.

lucasmoura:
@blackboxsw I have just updated the PR with a way to reuse existing network interfaces. This is an effort to avoid us to create those resources every time we need to provision a virtual machine.

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,851 @@
...
+
+        raise RuntimeError(
+            """
+            Error locationg the ip address: {}.
+            This ip address was not found in this subscription.
+            """
+        )
+
+    def _retrive_instance_ip(self, instance):
+        """Retrive public ip address of instance.
```
blackboxsw, line 712:
```suggestion
    def _retrieve_instance_ip(self, instance):
        """Retrieve public ip address of instance.
```

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,851 @@
...
+            name = self.tag
+
+        self._log.debug('using SSH key %s', name)
+        self.key_pair = KeyPair(public_key_path, private_key_path, name)
+
+    def _list_instances(self):
+        """List instances in associated with a subscription id.
+
+        Returns:
+            An list of azure virtual machine associated with the subscription
```
blackboxsw, line 681: This is more lie _get_instances since it is not returning a list of instance names but actually the virtal machine object 
If we were making this a public method at some point, I'd like us to wrap each Virtual Machine as an AzureInstance.
```suggestion
    def _list_instances(self):
        """Return an iterable of Azure VirtualMachine instances associated with a subscription id.

        Returns:
            A list of azure virtual machine objects associated with the subscription
```

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,851 @@
...
+            vm_parameters["storage_profile"]["image_reference"] = {
+                "publisher": 'Canonical',
+                "offer": util.get_offer_from_image_id(image_id),
+                "sku": util.get_sku_from_image_id(image_id),
+                "version": "latest"
+            }
+        else:
+            vm_parameters["storage_profile"]["image_reference"] = {
+                "id": image_id
+            }
```
blackboxsw, line 364: Maybe it's worth creating a private method for this here 
_get_image_reference_params(image_id)

This could return your dict
either 
{publisher:.... offer:... sku:... version:...} 
OR
{'id': image_id}

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,851 @@
...
+        return vm_call.result()
+
+    def delete_image(self, image_id):
+        """Delete an image from Azure.
+
+        Args:
+            image_id: string, The id of the image to be deleted
+        """
+        self._log.debug('Deleting Azure image: %s', image_id)
+        all_images = self.compute_client.images.list()
```
blackboxsw, line 403: It seems sort of costly to hit the `list()` method and walk through all images on these types of methods where we require an image_id param? I'm wondering why we wouldn't just call self.compute.client.images.get(image_name=image_id, resource_group_name=self.tag)

Maybe you are trying to be more flexible and searching all images in all resource groups so that they can be removed instead of a specific/known resource group?

I ask because this behavior differs from list_keys method for instance which is actually just looking within the self.resource_group.name

pycloudlib/azure/cloud.py
```diff
@@ -0,0 +1,851 @@
...
+        self.tag = create_tag(self.base_tag)
+
+        if self.resource_group is None:
+            self.resource_group = self._create_resource_group()
+
+        # Check if we already have an existing network interface that is not
+        # attached to a virtual machine. If we have, we will just use it
+        nic = self._check_for_network_interfaces()
+
+        if nic is None:
```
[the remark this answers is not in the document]
lucasmoura: It seems that it does require that parameter. The `network_profile` is a required field when provisioning a virtual machine. And if you try to create one with an empty `network_profile`, it raises the error:

```
Message: Virtual machine <name-of-the-vm> must have at least one network interface.
```

pycloudlib/azure/cloud.py
lucasmoura: Fair enough, I will update the name of the method. But to make each of the returned virtual machines an AzureInstance, we would need to retrieve the IP address of each virtual machine as well. Which can kind of expensive, considering we are now only using this method to retrieve the right instance.

But I agree that if we want to make this method public, it is a path we must follow to do that.

pycloudlib/azure/cloud.py
lucasmoura: My idea here was to extract the image_name from the Image object returned by azure, since in the API we need to the name of the image, instead of the id. But we can extract the name from the id, so I think it is better to use that approach.

lucasmoura:
@blackboxsw I have updated the code

lucasmoura:
@blackboxsw Done, I have updated the code

pycloudlib/azure/instance.py
```diff
@@ -0,0 +1,111 @@
...
+        """Wait for instance stop."""
+        raise NotImplementedError
+
+    @property
+    def ip(self):
+        """Return IP address of instance."""
+        return self._instance["ip_address"]
+
+    @property
+    def instance_name(self):
```
blackboxsw, line 41: Since we are already on Instance class, no need to reuse the instance_ prefix.

```suggestion
    def name(self):
```

pycloudlib/azure/instance.py
```diff
@@ -0,0 +1,111 @@
...
+        self.boot_timeout = 300
+        self.status = "active"
+
+    def wait_for_delete(self):
+        """Wait for instance to be deleted."""
+        raise NotImplementedError
+
+    def wait_for_stop(self):
+        """Wait for instance stop."""
+        raise NotImplementedError
```
blackboxsw, line 33: Also let's drop these.

pycloudlib/azure/instance.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+
+        if wait:
+            delete.wait()
+
+        self.status = "deleted"
+
+    def wait(self):
+        """Wait for instance to be up and cloud-init to be complete."""
+        self.execute("cloud-init status --wait")
```
blackboxsw, line 111: this should be self._wait_for_system()

pycloudlib/azure/instance.py
lucasmoura: If we remove those methods we will get some pylint warnings, since they are abstract methods in the base class.

pycloudlib/azure/instance.py
lucasmoura: The reason I have used that name is because of the conflict with this definition:
https://github.com/canonical/pycloudlib/blob/master/pycloudlib/instance.py#L31

pycloudlib/instance.py
```diff
@@ -28,12 +28,17 @@ def __init__(self, key_pair):
...
         self.boot_timeout = 120
         self.key_pair = key_pair
         self.port = '22'
         self.username = 'ubuntu'
 
+    @property
+    def name(self):
+        """Return instance name."""
+        self.name = ''
+        return self.name
```
blackboxsw, line 40: We can't set self.name within the @property method name as there is no setter. You'd end up getting the following error on BaseInstance:

```Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 6, in name
AttributeError: can't set attribute
```

If we just return an empty string we should be good here. Let subclasses implement this at will

```suggestion
        return ''
```

PR #13, opened 2020-07-31
Add support for Azure pro images

lucasmoura:
Allow users to launch Azure pro images from both marketplace and snapshots of Azure pro virtual machines.

To support that, we had to add the `plan`field in the vm parameters dict.

To test these changes, I have used the following image ids:
```
Canonical:0001-com-ubuntu-pro-bionic:pro-18_04-lts
Canonical:0001-com-ubuntu-pro-focal:pro-20_04-lts
 ```

### DIFF
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -27,7 +27,7 @@ class Azure(BaseCloud):
     }
 
     def __init__(self, tag, client_id=None, client_secret=None,
-                 subscription_id=None, tenant_id=None):
+                 subscription_id=None, tenant_id=None, region="centralus"):
         """Initialize the connection to Azure.
 
         Azure will try to read user credentials form the /home/$USER/.azure
@@ -40,10 +40,11 @@ class Azure(BaseCloud):
             client_secret: user's client secret access key
             subscription_id: user's subscription id key
             tenant_id: user's tenant id key
+            region: The region where the instance will be created
         """
         super().__init__(tag)
         self._log.debug('logging into Azure')
-        self.location = "centralus"
+        self.location = region
         self.username = "ubuntu"
 
         self.registered_instances = {}
@@ -296,7 +297,7 @@ class Azure(BaseCloud):
 
         return nic_call.result()
 
-    def _create_vm_parameters(self, vm_name, image_id, nic_id, user_data):
+    def _create_vm_parameters(self, name, image_id, nic_id, user_data):
         """Create the virtual machine parameters to be used for provision.
 
         Composes the dict that will be used to provision an Azure virtual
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pycloudlib/azure/cloud.py [hunk 1]
```diff
@@ -357,6 +357,16 @@ def _create_vm_parameters(self, vm_name, image_id, nic_id, user_data):
         vm_parameters["storage_profile"][
             "image_reference"] = util.get_image_reference_params(image_id)
 
+        # We can have pro images from two differente sources, marktplaces
+        # and snapshots. A snapshot image do not have encoded in their
+        # image id the necessary info to create the 'plan' dict. To solve
+        # that, we are getting the necessary info from the registered_images
+        # dict, were we store the required information
+        registered_image = self.registered_images.get(image_id)
```
blackboxsw, line 365: The way we are setting up and using registered_images is really just to keep track of the original image_id on snapshots right? What do you think about instead adding a tag "img_src:<image_id>" for any snapshot created by pycloudlib? Would we then be able determine that information from snapshot metadata in the future via az SDK calls checking tags on an image (if the image_id isn't of  the form "publisher:offer:sku"?

lucasmoura:
@blackboxsw Code updated. I had to made some changes to it to still allows us to launch images, but if you do not agree with the changes, let me know

pycloudlib/azure/cloud.py
```diff
@@ -793,7 +804,10 @@ def snapshot(self, instance):
                     "id": instance.id
                 },
                 "tags": {
-                    "name": self.tag
+                    "name": self.tag,
+                    "id": "pycloudlib-{}-{}".format(
+                        instance.offer, instance.sku
+                    )
```
blackboxsw, line 810: Can we add an instance_id property to Azure instance and name this tag key  "src-img-id" and make it the same format as the original launched image_id?


```suggestion
                    "src-image-id": instance.instance_id
```


This would involve addiing the following to azure/instance.py
```diff
diff --git a/pycloudlib/azure/instance.py b/pycloudlib/azure/instance.py
index 56bd218..9f75b3e 100644
--- a/pycloudlib/azure/instance.py
+++ b/pycloudlib/azure/instance.py
@@ -32,6 +32,16 @@ class AzureInstance(BaseInstance):
         """Wait for instance stop."""
         raise NotImplementedError
 
+    @property
+    def image_id(self):
+        """Return the image_id from which this instance was created."""
+        storage_profile = self._instance['vm'].as_dict()['storage_profile']
+        image_ref = storage_profile['image_reference']
+
+        return ":".join(
+            [image_ref['publisher'].lower(), image_ref['offer'],
+             image_ref['sku'], image_ref['version']])
+
     @property
     def ip(self):
         """Return IP address of instance."""
```

lucasmoura:
@blackboxsw Done. I have updated the code

lucasmoura:
@blackboxsw Done

PR #18, opened 2020-09-01
tox: pin the flake8 and pylint versions and introduce the tip envs

paride:


paride:
Once this is merged and the api refactor PR updated with master CI should pass for it. Once that's merged we can make pylint 2.6.0 happy and bump its version in `tox.ini`.

PR #19, opened 2020-09-01
Add timeout to paramiko connect

lucasmoura:
In `uaclient` we are having a problem with connectivity issues regarding paramiko connect. One hypothesis that we have is that the connect get `stuck` and since we have no timeout on that operation, it hangs indefinitely. This PR is adding a timeout parameter to this operation.

Although this should be parametrized by all instance types, I think it is better to wait for #9 to land, to make that consistent. Also, we could also check if the timeout issue will actually fix the uaclient problem.

### DIFF
--- a/pycloudlib/instance.py
+++ b/pycloudlib/instance.py
@@ -34,8 +34,10 @@ class BaseInstance(ABC):
         self.key_pair = key_pair
         self.port = '22'
         self.username = 'ubuntu'
+        self.connect_timeout = 60
 
-    @abstractproperty
+    @abstractmethod
+    @property
     def name(self):
         """Return instance name."""
         raise NotImplementedError
@@ -311,6 +313,7 @@ class BaseInstance(ABC):
                     username=self.username,
                     hostname=self.ip,
                     port=int(self.port),
+                    timeout=self.connect_timeout,
                     key_filename=self.key_pair.private_key_path,
                 )
                 self._ssh_client = client
@@ -318,6 +321,10 @@ class BaseInstance(ABC):
             except (ConnectionRefusedError, AuthenticationException,
                     BadHostKeyException, ConnectionResetError, SSHException,
                     OSError) as e:
+                self._log.info(
+                    "%s\nRetrying ssh connection %d more time(s) to  %s@%s:%s",
+                    e, retries, self.username, self.ip, self.port
+                )
                 last_exception = e
                 retries -= 1
                 time.sleep(10)


pycloudlib/instance.py
```diff
@@ -33,6 +33,8 @@ def __init__(self, key_pair):
         self.port = '22'
         self.username = 'ubuntu'
 
+        self.connect_timeout = connect_timeout if connect_timeout else 60
```
OddBloke, line 36: This prohibits setting a `connect_timeout` of 0; 0 is False-ish, so it will be set to 60 here.

```suggestion
        self.connect_timeout = connect_timeout if connect_timeout is not None else 60
```

blackboxsw (dismissed):
+1, if we need to, we might end up having to expose banner_timeout and auth_timeout params to paramiko connect()

lucasmoura:
@OddBloke I have address the changes for now. Also, I have decided to not add this parameter to the all of the clouds we are supporting in the moment, since the problem may be a little deeper than that. But if it is, I see no problem on allowing the instance to control that timeout behavior

pycloudlib/instance.py
```diff
@@ -311,13 +317,15 @@ def _ssh_connect(self):
...
                     port=int(self.port),
+                    timeout=self.connect_timeout,
                     key_filename=self.key_pair.private_key_path,
                 )
                 self._ssh_client = client
                 return client
             except (ConnectionRefusedError, AuthenticationException,
                     BadHostKeyException, ConnectionResetError, SSHException,
                     OSError) as e:
+                self._log.warning("Connection error: %s", e)
```
blackboxsw, line 328: can we change this text so that it doesn't say error in a warning message (because we are retrying).

I had a previous suggestion that would work, though I think log.warning is reasonable instead of log.info.

lucasmoura:
@blackboxsw updated

pycloudlib/instance.py
```diff
@@ -23,7 +23,7 @@ class BaseInstance(ABC):
 
     _type = 'base'
 
-    def __init__(self, key_pair):
+    def __init__(self, key_pair, connect_timeout=None):
```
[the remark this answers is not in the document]
lucasmoura: My initial idea was to update all `launch` methods to allow this parameter to be passed. But I did not thing for the case of `get_instance`.

But I guess we can have the default value as `60`  as you said and do not add this parameter into the `__init__` method. What do you think @blackboxsw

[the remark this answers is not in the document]
blackboxsw: @lucasmoura +1 let's leave the 60 second default and not worry about parameterizing it unless we find a use-case to for certain call sites

lucasmoura:
@blackboxsw @OddBloke Code updated

pycloudlib/instance.py
```diff
@@ -34,8 +34,10 @@ def __init__(self, key_pair):
         self.key_pair = key_pair
         self.port = '22'
         self.username = 'ubuntu'
+        self.connect_timeout = 60
 
     @abstractproperty
+    @property
```
OddBloke, line 40: Thanks for adding the `@property` decorator! Per https://docs.python.org/3/library/abc.html#abc.abstractproperty, this should be:

```suggestion
    @property
    @abstractmethod
```

because `abstractproperty` has been deprecated since 3.3 (which predates even trusty, which is 3.4).

OddBloke (approved):
One minor inline comment which can come as a follow-up if Chad approves this before it's addressed.

lucasmoura:
@OddBloke Updated

PR #22, opened 2020-09-02
Fix stdout+stderr redirection to /dev/null

paride:
This started by fixing a bug in the deleted code:

  cloud-init status --wait 2&>1 > /dev/null

should have been:

  cloud-init status --wait > /dev/null 2>&1

however now that Trusty reached EOL we can just assume that
`cloud-init status --wait` is available and drop the wait
logic based on polling runlevel(8).


paride:
Thanks @blackboxsw, now it just fixes the redirection.

PR #24, opened 2020-09-03
Fix order of decorators in instance module

lucasmoura:
With the current decorator order for the `name` method in the `instance` module we raise the following error when trying to run the code:

```python
File "/usr/lib/python3.6/abc.py", line 25, in abstractmethod
    funcobj.__isabstractmethod__ = True
AttributeError: attribute '__isabstractmethod__' of 'property' objects is not writable
```

This is raised because the decorator order is incorrect.


paride (approved):
Thanks!

PR #27, opened 2020-09-16
Gce instance

TheRealFalcon:
Expand the functionality of GCE so it roughly aligns with the functionality of the other clouds

### DIFF
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -7,6 +7,7 @@ instance. It however, does not allow any further actions from occuring.
 """
 
 import logging
+import os
 import time
 from itertools import count
 
@@ -108,9 +109,13 @@ class GCE(BaseCloud):
         Args:
             image_id: string, id of the image to delete
         """
+        api_image_id = self.compute.images().get(
+            project=self.project,
+            image=os.path.basename(image_id)
+        ).execute()['id']
         response = self.compute.images().delete(
             project=self.project,
-            image=image_id,
+            image=api_image_id,
         ).execute()
 
         raise_on_error(response)
@@ -182,7 +187,7 @@ class GCE(BaseCloud):
         ).execute()
         raise_on_error(operation)
 
-        self._wait_for_operation(operation)
+        self._wait_for_operation(operation, operation_type='zone')
 
         result = self.compute.instances().get(
             project=self.project,
@@ -210,7 +215,7 @@ class GCE(BaseCloud):
             clean: run instance clean method before taking snapshot
 
         Returns:
-            An image object
+            An image id
         """
         response = self.compute.disks().list(
             project=self.project, zone=self.zone
@@ -227,16 +232,18 @@ class GCE(BaseCloud):
         instance.shutdown()
 
         snapshot_name = '{}-image'.format(instance.name)
-        response = self.compute.images().insert(
+        operation = self.compute.images().insert(
             project=self.project,
             body={
                 'name': snapshot_name,
                 'sourceDisk': instance_disks[0]['selfLink'],
             }
         ).execute()
-        raise_on_error(response)
+        raise_on_error(operation)
+        self._wait_for_operation(operation)
 
-        return snapshot_name
+        return 'projects/{}/global/images/{}'.format(
+            self.project, snapshot_name)
 
     def use_key(self, public_key_path, private_key_path=None, name=None):
         """Use an existing already uploaded key.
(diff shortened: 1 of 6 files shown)


pycloudlib/gce/cloud.py
```diff
@@ -98,19 +109,30 @@ def delete_image(self, image_id):
...
+            instance_id: The instance ID returned upon creation
+            project: Project used when creating this instance
+            zone: Zone used when creating this instance
 
         Returns:
             An instance object to use to manipulate the instance further.
 
         """
-        raise NotImplementedError
+        return GceInstance(self.key_pair, instance_id, project, zone, name)
```
blackboxsw, line 135: We need to update this callsite to source self.project/self.region/self.zone if not provided.

blackboxsw (changes requested):
snapshot on a GCEInstance seems to give  me a traceback:
Exception ignored in: <function BaseInstance.__del__ at 0x7f97919ef940>
  File "/home/ubuntu/pycloudlib/pycloudlib/instance.py", line 104, in __del__
    if self._sftp_client:
AttributeError: 'GceInstance' object has no attribute '_sftp_client'
Exception ignored in: <function BaseInstance.__del__ at 0x7f97919ef940>
  File "/home/ubuntu/pycloudlib/pycloudlib/instance.py", line 104, in __del__
    if self._sftp_client:
AttributeError: 'GceInstance' object has no attribute '_sftp_client'


Can we get a minor py3 unittest covering this method?

pycloudlib/gce/cloud.py
```diff
@@ -153,36 +173,77 @@ def launch(self, image_id, instance_type='n1-standard-1', user_data=None,
...
 
-    def snapshot(self, instance, clean=True):
+        return self.get_instance(
+            result['id'],
+            self.project,
+            self.zone,
+            name=result['name']
+        )
+
+    def snapshot(self, instance: GceInstance, clean=True, **kwargs):
```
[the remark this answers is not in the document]
TheRealFalcon: Actually, now that I see this again, we don't need the `**kwargs`...that's just for the base class in case any functions have more kwargs they can take (like oci). So no need to change ec2 either.

pycloudlib/gce/cloud.py
TheRealFalcon: Source from where? I figured it'd be invalid if they're not provided.

There's `None` checking for it when the instance is created, so I didn't think we'd need to also have the checking here.
https://github.com/canonical/pycloudlib/pull/27/files#diff-d800499e0c4918d48411248888644f8aR27

TheRealFalcon:
@blackboxsw How long ago did you pull my branch? I added a 2nd commit the other day that I thought fixed the snapshot issue.

pycloudlib/gce/cloud.py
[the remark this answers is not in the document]
blackboxsw: +1 works for me

pycloudlib/gce/cloud.py
blackboxsw: @TheRealFalcon This question really is, are you intending to support project/zone overrides in GCE.get_instance?
We already persist the project and zone on the GCE.BaseCloud instance, maybe we should make those params mandatory in __init__ and we don't have to surface them here too. Or maybe you intended to make get_instance allow for default overrides of project and zone in case a user wanted to switch zones using an already initialized GCE cloud?.

TheRealFalcon: Ah, I see what you're saying now. I wasn't intending they'd be overridden in the `get_instance` call, so it makes sense to use the instance variables rather than passing them there. I think I was thinking of it as some kind of static-like method, even though I don't think we want that. I'll fix it

pycloudlib/gce/util.py
```diff
@@ -0,0 +1,19 @@
+"""Common GCE utils."""
+
+
+def raise_on_error(response):
+    """Look for errors in response and raise if found."""
+    if 'httpErrorStatusCode' in response:
+        raise Exception(
```
blackboxsw, line 7: Is it worth fielding urllib.error.HTTPError? maybe raise RuntimeError to make the error a bit more specialized than Exception?

TheRealFalcon: Sure, I can make a custom exception for the other one too

pycloudlib/gce/instance.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def __repr__(self):
+        """Create string representation of class."""
+        return '{}(instance_id={})'.format(
+            self.__class__.__name__,
+            self.instance_id,
+        )
+
+    @property
+    def name(self):
```
blackboxsw, line 48: Can we add a def id(self): that is an alias to name property as current callers on ec2 typically use this alias since our cloud get_instance calls for an instance_id param (instead of name)

blackboxsw: actual instance_id is better thanks .

blackboxsw (approved):
+1! LGTM bring it home

Issue #30, opened 2020-09-29
`pycloudlib.instance.BaseInstance.execute` will always raise an exception if commands exit non-zero on KVM or lxd


OddBloke:
`pycloudlib.util.subp` specifies a default `rcs` of `(0,)`: https://github.com/canonical/pycloudlib/blob/fcf6268df72540b99872c1312b8716a7e88b7564/pycloudlib/util.py#L146 which means that, unless a caller passes a different value for `rcs`, it will raise an exception if the command it executes exits non-zero: https://github.com/canonical/pycloudlib/blob/fcf6268df72540b99872c1312b8716a7e88b7564/pycloudlib/util.py#L192-L200

`BaseInstance.execute` does not pass a value for `rcs` in for lxd: https://github.com/canonical/pycloudlib/blob/fcf6268df72540b99872c1312b8716a7e88b7564/pycloudlib/instance.py#L152-L153 or for KVM: https://github.com/canonical/pycloudlib/blob/fcf6268df72540b99872c1312b8716a7e88b7564/pycloudlib/instance.py#L159-L160 hence the exception being raised.

PR #32, opened 2020-09-29
instance: allow KVM/LXD commands to exit non-zero

OddBloke:
`pycloudlib.util.subp`'s default behaviour is to raise an exception if
the command being executed exits non-zero.  Before this commit, this
meant that executing a command against a LXD or KVM instance that exited
non-zero would raise an exception (instead of returning a `Result`
object as other instances do).

Fixes: #30

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -8,8 +8,9 @@ jobs=4
 # R0903: Too few public methods (%s/%s)
 # R0913: Too many arguments
 # C0103: Invalid name
+# C0122: misplaced-comparison-constant
 # R0201: Method could be a function (no-self-use)
-disable=R0902, R0903, R0913, W0221, C0103, R0201
+disable=R0902, R0903, R0913, W0221, C0103, C0122, R0201
 
 [TYPECHECK]
 # Ignore the googleapiclient module to avoid no-member checks
--- a/pycloudlib/instance.py
+++ b/pycloudlib/instance.py
@@ -150,14 +150,16 @@ class BaseInstance(ABC):
 
         if self._type == 'lxd':
             base_cmd = ['lxc', 'exec', self.name, '--']
-            return subp(base_cmd + list(command))
+            return subp(base_cmd + list(command), rcs=None)
 
         # multipass handling of redirects is buggy, so we don't bind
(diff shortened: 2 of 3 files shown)


paride (approved):
I'm +1 for this (approving), but a bit surprised that we like "constant first asserts" to the point of disabling pylint's C0122 entirely. It doesn't seem we have the same disable in the other projects, and the first `assert` example in pytest's documentation [1] has the constant at the right hand side. Are "constant first asserts" preferred for some reason I'm missing?

/me asking for his TIL moment :-)

[1] https://docs.pytest.org/en/stable/assert.html#asserting-with-the-assert-statement

OddBloke:
On Wed, Sep 30, 2020 at 05:10:08AM -0700, Paride Legovini wrote:
> It doesn't seem we have the same disable in the other projects,

cloud-init disables all of pylint's C* warnings, so it's included in
that.

> and the first `assert` example in pytest's documentation [1]
> has the constant at the right hand side. Are "constant first asserts"
> preferred for some reason I'm missing?

This isn't exactly an answer to your question, but the preference is
expressed in our HACKING doc for cloud-init, specifically the fourth
bullet under
https://cloudinit.readthedocs.io/en/latest/topics/hacking.html#unit-testing.
Disabling the warning here isn't necessarily intended to indicate that
we should follow that preference in this codebase, but it does mean that
we aren't _forced_ to write assertions in opposite orders in the two
projects.

To actually answer your question: we don't only apply our rule to
constants, it gives us consistency between assertions, and there are
cases where it's clearer IMO:

```py
assert "constant" == (
    some_very_long_invocation_that_requires_a_line_break(etc, etc, etc))
```

is clearer than:

```py
assert some_very_long_invocation_that_requires_a_line_break(
    etc, etc, etc) == "constant"
```

The constant can be very easy to lose in the function invocation "noise"
in the latter case, in my experience.

The pylint warning exists because some languages recommend spelling
conditions "<constant> == <variable>" so that if you typo it as
"<constant> = <variable>" then you don't accidentally (and silently)
perform assignment.  This is not necessary in Python (hence the
warning), but that isn't the reason we're doing it (hence disabling it).

paride:
Thanks Dan!

PR #36, opened 2020-09-30
instance: refactor BaseInstance.wait

### LINKED ISSUE #35: Not all `pycloudlib.instance.BaseInstance` subclasses call `self._wait_for_system`
### Problem Statement

`BaseInstance` has an abstract `wait` method; this is called when consumers want to wait for the instance to launch and for cloud-init to be complete:

https://github.com/canonical/pycloudlib/blob/1cd00cdc7deed8ae4df927f1a1ff1a25c2349fa0/pycloudlib/instance.py#L87-L90

It also has a concrete `_wait_for_system` method, which performs the waiting for `cloud-init` to complete (this is just the first few lines):

https://github.com/canonical/pycloudlib/blob/1cd00cdc7deed8ae4df927f1a1ff1a25c2349fa0/pycloudlib/instance.py#L361-L363

It follows that whenever `wait` is called, `_wait_for_system` should also be called. However, with the code as-is, we rely on `BaseInstance` subclasses doing so in their implementation of `wait`; they currently don't always.

### Proposed Solution

Instead of fixing up those specific cases for now, I instead propose that we refactor the way that waiting works to ensure that the "call `_wait_for_system` when `wait` is called" invariant will always hold. My proposed steps:

* add a no-op concrete method to `BaseInstance` named `_wait_for_start`
* convert `wait` to a concrete method which calls `self._wait_for_start()` then `self._wait_for_system()`
* remove the implementations of `wait` which currently only call `self._wait_for_system()` (i.e. Azure, LXD, KVM)
* refactor the remaining implementations of `wait` to drop the call to `self._wait_for_system()` (if present), and rename them to `_wait_for_start`

This means that calling `instance.wait()` will always call `self._wait_for_system`, while still allowing cloud-specific waiting to be captured in `self._wait_for_start`.

OddBloke:
This modifies BaseInstance.wait from an abstract to a concrete method
which calls `self._wait_for_start` then `self._wait_for_system`.  This
ensures that we will always call `self._wait_for_system` when `wait` is
called, while still allowing subclasses to define their own additional
waiting logic.

Any existing subclass `wait` implementations which only called
`self._wait_for_system` have been removed (as the superclass now
performs that itself).  The remaining implementations have been renamed
to `_wait_for_start` and have had any calls to `self._wait_for_system`
removed (as the superclass will now do that).

Fixes #35

### DIFF
--- a/pycloudlib/azure/instance.py
+++ b/pycloudlib/azure/instance.py
@@ -140,10 +140,6 @@ class AzureInstance(BaseInstance):
 
         self.status = "deleted"
 
-    def wait(self):
-        """Wait for instance to be up and cloud-init to be complete."""
-        self._wait_for_system()
-
     def console_log(self):
         """Return the instance console log."""
         raise NotImplementedError
--- a/pycloudlib/ec2/instance.py
+++ b/pycloudlib/ec2/instance.py
@@ -210,13 +210,12 @@ class EC2Instance(BaseInstance):
         if wait:
             self.wait()
 
-    def wait(self):
-        """Wait for instance to be up and cloud-init to be complete."""
+    def _wait_for_instance_start(self):
+        """Wait for instance to be up."""
         self._log.debug('wait for instance running %s', self._instance.id)
         self._instance.wait_until_running()
         self._log.debug('reloading instance state %s', self._instance.id)
         self._instance.reload()
-        self._wait_for_system()
 
     def wait_for_delete(self):
         """Wait for instance to be deleted."""
--- a/pycloudlib/gce/instance.py
+++ b/pycloudlib/gce/instance.py
@@ -135,11 +135,10 @@ class GceInstance(BaseInstance):
         if wait:
             self.wait()
 
-    def wait(self):
+    def _wait_for_instance_start(self):
         """Wait for instance to be up."""
         self._wait_for_status('RUNNING')
         self._ip = self._get_ip()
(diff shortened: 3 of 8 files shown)


TheRealFalcon (approved):
Looks fine to me as is, but some minor comments:
I know I already agreed with your initial proposal, but now I'm wondering if something like `_wait_for_instance_start` and `_wait_for_cloud_init` would be more descriptive than `_wait_for_start` and `_wait_for_system`. I'm not actually sure what either method does by looking at them.

I question if the unit test adds anything of value. There's no conditionals or varying input or differences in returns it needs to verify. If the implementations of any of these methods change, it won't surface any bugs. I just see it as a larger maintenance overhead.

OddBloke:
> I know I already agreed with your initial proposal, but now I'm wondering if something like `_wait_for_instance_start` and `_wait_for_cloud_init` would be more descriptive than `_wait_for_start` and `_wait_for_system`. I'm not actually sure what either method does by looking at them.

Yep, those are better names; changed.

> I question if the unit test adds anything of value. There's no conditionals or varying input or differences in returns it needs to verify. If the implementations of any of these methods change, it won't surface any bugs. I just see it as a larger maintenance overhead.

Yeah, I see what you mean: this falls into the category of things that I probably wouldn't request a test for in review, but I think I lean towards including it (though perhaps this is sunk cost fallacy :p).

PR #40, opened 2020-10-29
Add vm support to lxd

lucasmoura:
This PR adds LXD support for launching vms. Although this is working version, there are still some aspects of the code that may be updated:

* `wait_for_cloudinit`: Right now, I am looping on any `OSError` exception raised, meaning that unrelated `lxd-agent` errors will trigger unnecessary sleep calls. I can parse the `OSError` here and search for `lxd-agent` before looping again.
* `image_id parameter`: I am assuming that the user will always pass as image_id the raw release name [`xenial`, `bionic`, `focal`] which may not be always true
* `trusty`: There is no way to launch trusty vm through lxd, but I am not informing the user about that.

Although I have those improvements to implement, I think the code is already in a good state for an initial review


blackboxsw (changes requested):
Thanks @lucasmoura for the first pass here. A number of comments and cleanup to look at

pycloudlib/lxd/cloud.py
```diff
@@ -57,19 +92,61 @@ def get_instance(self, instance_id):
         """
         return LXDInstance(instance_id)
 
-    # pylint: disable=R0914
+    def _set_release_image(self, release, is_vm):
```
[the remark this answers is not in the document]
lucasmoura: Yes, this one was just to avoid triggering pylint warnings for `R0912`. But since I am disabling it on the `init` method now, there is no need to keep it here

pycloudlib/lxd/cloud.py
```diff
@@ -78,20 +155,60 @@ def init(
...
-            cmd.append(name)
+                self.create_profile(
+                    profile_name=profile_name,
+                    profile_config=base_vm_profiles[base_release]
+                )
+
+                profile_list = [profile_name]
+
+        if use_ssh:
+            error_msg = (
```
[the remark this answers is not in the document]
lucasmoura: I have updated that part of the code, since I realized my checks here were not making much sense in the context of an user providing its own ssh keys to issue commands.

pycloudlib/lxd/cloud.py
```diff
@@ -78,20 +155,60 @@ def init(
...
+                "You need to provide a valid ssh key through"
+                "ssh_authorized_keys or ssh_import_id in your"
+                "user_data to run commands through ssh"
+            )
+            search_pattern = re.compile("(ssh_authorized_keys|ssh_import_id)")
+            user_data_str = user_data or config_dict.get("user.user-data")
+
+            if user_data_str and search_pattern.match(user_data_str) is None:
+                raise Exception(error_msg)
```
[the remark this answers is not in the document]
lucasmoura: Same thing here

pycloudlib/lxd/cloud.py
```diff
@@ -78,20 +155,60 @@ def init(
...
+
+            self._manage_ssh_key_pair(name)
+            user_data = textwrap.dedent(
```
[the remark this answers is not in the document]
lucasmoura: Yes, I have addressed that in my last code update. What I am doing is appending the `ssh_authorized_keys` bit into any existing `user_data`. But maybe we could handle that differently.

pycloudlib/lxd/cloud.py
```diff
@@ -128,16 +243,18 @@ def init(
...
         self._log.debug('Created %s', name)
 
-        return LXDInstance(name)
+        return LXDInstance(name, is_vm, self.key_pair)
 
     def launch(self, image_id, instance_type=None, user_data=None, wait=True,
                name=None, ephemeral=False, network=None, storage=None,
-               profile_list=None, config_dict=None, **kwargs):
+               profile_list=None, config_dict=None, is_vm=False,
+               use_ssh=False, **kwargs):
```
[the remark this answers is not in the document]
lucasmoura: Right now, we are using the first approach. If we have a key pair, lxc will always communicate with the machine through ssh.
But I think refactoring it to support the `prefer_ssh`  approach is an excellent idea. I will refactor the code for that

blackboxsw (commented):
bump regex request.

lucasmoura:
@blackboxsw Code updated.

1) It seems that the `ssh` issue was being caused by using the `--force` flag on `lxc stop`
2)  I have kept the the `use_ssh`  parameter for launch. The reason for that is that `execute` method is being called internally by pycloudlib. Meaning that parameters like `force_ssh` would only work for directly user call of the `execute` method. To be consistent, I am always executing commands through ssh when we have ssh keys in the instance. But we can better discuss this.

lucasmoura:
Also @blackboxsw this is the script I am using to test this new feature regarding FIPS on xenial machine:

```python
import textwrap
import logging
import time
import pycloudlib


RELEASE = 'xenial'
STAGING_TOKEN = "YOUR-TOKEN"


def main():

    lxd = pycloudlib.LXD('example-vm')
    name = 'pycloudlib-vm'

    ssh_user_data = textwrap.dedent(
        """\
        #cloud-config
        ssh_authorized_keys:
            - ssh-rsa blah
        """
    )
    
    image_id = lxd.released_image("xenial", is_vm=True)
    inst = lxd.launch(name=name, image_id=image_id, is_vm=True, use_ssh=True)
    result = inst.execute("lsb_release -a")
    print(result)

    print(inst.execute("sudo add-apt-repository ppa:canonical-server/ua-client-daily -y"))
    print(inst.execute("sudo apt-get update"))
    result = inst.execute("sudo apt-get install ubuntu-advantage-tools -y")

    print(inst.execute("ua version"))
    inst.execute("ssh-import-id lamoura")
    inst.execute("sudo sed -i 's/contracts.can/contracts.staging.can/' /etc/ubuntu-advantage/uaclient.conf")
    inst.execute("sudo ua attach {}".format(STAGING_TOKEN))
    inst.execute("sudo ua disable livepatch")
    inst.execute("sudo ua enable fips --beta --assume-yes")

    inst.restart(wait=True, force=False)
    print(inst.execute("ua status --all"))
    print(inst.execute("lsb_release -a"))
    print(inst.execute("uname -r"))

    inst.shutdown(force=False)
    inst.start()
    inst.restart(force=False)
    inst.delete()


if __name__ == "__main__":
    main()
```

blackboxsw (approved):
just a few nits remaining on this round.

lucasmoura:
@blackboxsw Code updated

lucasmoura:
@blackboxsw Just modified how we create the ssh keys. You were right here, it is better to create the ssh keys outside of the launch method

pycloudlib/tests/lxd/__init__.py
```diff
@@ -0,0 +1 @@
+"""Tests related to pycloudlib.lxd module."""
```
blackboxsw, line 1: I would suggest we align with other projects and have a tests subdirectory below the module we are testing (not a top-level pycloudlib/tests  directory that has a full mapping of all submodules.

I think the lxd tests should live under `pycloublib/lxd/tests` instead of `pycloudlib/tests/lxd` so devs don't have to cd  up and across from a top-level directory.

```

File: pycloudlib/lxd/instance.py [hunk 7]
```diff
@@ -190,17 +216,23 @@ def restore(self, snapshot_name):
                         self.name, snapshot_name)
         subp(['lxc', 'restore', self.name, snapshot_name])
 
-    def shutdown(self, wait=True):
+    def shutdown(self, wait=True, force=True):
```
blackboxsw, line 219: We should change pycloudlib/instance.py and pycloudlib/*/instance.py:shutdown and restart methods to accept **kwargs
```suggestion
    def shutdown(self, wait=True, force=True, **kwargs):
```

pycloudlib/tests/lxd/test_cloud.py
```diff
@@ -0,0 +1,143 @@
...
+        cloud = LXD(tag="test")
+
+        fake_stdout = io.StringIO()
+        with contextlib.redirect_stdout(fake_stdout):
+            cloud.create_profile(
+                profile_name="test_profile",
+                profile_config="profile_config"
+            )
+
+        expected_msg = "The profile named test_profile already exist"
```
blackboxsw, line 26:
```suggestion
        expected_msg = "The profile named test_profile already exists"
```

blackboxsw (changes requested):

lucasmoura:
@blackboxsw Updated

pycloudlib/lxd/instance.py
```diff
@@ -272,3 +304,33 @@ def wait_for_stop(self):
...
+        If the instance is a virtual machine, we need to wait for the
+        lxd-agent to be ready before getting the cloud-init status.
+        To do that, we retry on vm instances if raise_on_failure
+        is specified.
+
+        :param raise_on_failure:
+            When `True`, if the process waiting for cloud-init exits non-zero
+            then this method will raise an `OSError`.
+        """
+        if self._is_vm and raise_on_failure:
```
blackboxsw, line 320: when lxd.get_instance is called we don't detect Type: virtual-machine on those returned LXDInstances, as a result self._is_vm is False and we fall through to the setup()._wait_for_cloud_init which fails with 
```
  File "/home/csmith/src/pycloudlib/pycloudlib/instance.py", line 392, in _wait_for_cloudinit
    raise OSError(
OSError: cloud-init failed to start: out:  error: Error: Failed to connect to lxd-agent
```

pycloudlib/lxd/instance.py
```diff
@@ -11,20 +12,31 @@ class LXDInstance(BaseInstance):
...
     def __repr__(self):
         """Create string representation for class."""
         return 'LXDInstance(name={})'.format(self.name)
 
+    def _run_command(self, command, stdin):
+        """Run command in the instance."""
+        if self.key_pair:
+            return super()._run_command(command, stdin)
+
+        base_cmd = ['lxc', 'exec', self.name, '--']
```
blackboxsw, line 37: I think we should default to non-root user for lxc exec commands (as we do in ua-tools integration tests)
```suggestion
        base_cmd = ['lxc', 'exec', "--user", "1000", self.name, '--']
```

lucasmoura:
@blackboxsw Thanks for the review. I have updated the code

pycloudlib/lxd/instance.py
```diff
@@ -11,20 +12,51 @@ class LXDInstance(BaseInstance):
...
 
         Args:
             name: name of instance
+            key_pair: SSH key object
         """
-        super().__init__(key_pair=None)
+        super().__init__(key_pair=key_pair)
 
         self._name = name
+        self._is_vm = None
```
blackboxsw, line 25: This should be a class attr, no need to set in __init__ we can reset that via is_vm() later as you already do
```diff
diff --git a/pycloudlib/lxd/instance.py b/pycloudlib/lxd/instance.py
index 20ddf7d..9594f47 100644
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -11,6 +11,7 @@ class LXDInstance(BaseInstance):
     """LXD backed instance."""
 
     _type = 'lxd'
+    _is_vm = None
 
     def __init__(self, name, key_pair=None):
         """Set up instance.
@@ -22,7 +23,6 @@ class LXDInstance(BaseInstance):
         super().__init__(key_pair=key_pair)
 
         self._name = name
-        self._is_vm = None
 
     def __repr__(self):
         """Create string representation for class."""
```

blackboxsw (commented):
+1 thanks for this Lucas!
minor nit https://github.com/canonical/pycloudlib/pull/40/files#r521666099 but I can land it with that change

blackboxsw (approved):

PR #42, opened 2020-11-03
Stop depending on specific versions of dependencies

### LINKED ISSUE #41: pycloudlib should not require specific versions of its dependencies
As a library, pycloudlib should specify its abstract dependencies (via `install_requires` with only lower/upper bounds expressed) so that consumers of the library can express their concrete requirements (via `requirements.txt`) without issues.

This is particularly important now that we're talking about pycloudlib being used across teams: we should not all have to be in exact lockstep on the versions of cloud libraries that we use.

(See https://packaging.python.org/discussions/install-requires-vs-requirements/ for some background.)

OddBloke:
As a library, pycloudlib should allow consumers to select precisely the versions of dependencies that they want to use. This commit achieves that by specifying lower (and where necessary, upper) bounds on the versions required, instead of specific versions.

It moves these specifications from `requirements.txt` to `setup.py`: per https://packaging.python.org/discussions/install-requires-vs-requirements/, this is the appropriate place for libraries to express their dependencies. `requirements.txt` will now install the local version of the library.

Fixes #41

### cloud-init

Test that the integration tests work (to the same level as `master`) on:

- [x] Azure
- [x] EC2
- [x] GCE
- [x] LXD container
- [x] Oracle

The [`deps` branch in my repo](https://github.com/OddBloke/cloud-init/tree/deps) is the corresponding refactor of cloud-init's dependencies, and currently has this proposed branch hard-coded as its `pycloudlib` dependency; this allows for easier testing of cloud-init.

### ubuntu-advantage-client

Test that the behave tests work (to the same level as `master`) on:

- [x] Azure
- [x] EC2
- [x] LXD container # not applicable (pycloudlib not used yet)
- [x] LXD VM           # not applicable (pycloudlib not used yet)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -9,8 +9,9 @@ jobs=4
 # R0913: Too many arguments
 # C0103: Invalid name
 # C0122: misplaced-comparison-constant
+# C0301: line-too-long  (flake8 already catches this for us)
 # R0201: Method could be a function (no-self-use)
-disable=R0902, R0903, R0913, W0221, C0103, C0122, R0201
+disable=R0902, R0903, R0913, W0221, C0103, C0122, C0301, R0201
 
 [TYPECHECK]
 # Ignore the googleapiclient module to avoid no-member checks
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,15 +1,1 @@
-boto3==1.14.20
-botocore==1.17.20
-google-api-python-client==1.7.7
-paramiko==2.7.1
-pyyaml==5.1
-requests==2.22
-oci==2.17.0
-azure-mgmt-resource==10.0.0
-azure-mgmt-network==11.0.0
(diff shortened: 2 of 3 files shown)


OddBloke:
For cloud-init, I've tested Azure, EC2 and LXD containers locally. I don't have a working setup for GCE or Oracle, so I can't readily test them; @TheRealFalcon, would you be able to test those two?

@lucasmoura @blackboxsw Could you look into performing a UA client test run with this change? (You can probably do it by modifying your requirements and pushing to a branch that Travis will build?)

lucasmoura:
@OddBloke Just got a PR open to test the new dependencies config:

https://github.com/canonical/ubuntu-advantage-client/pull/1233

TheRealFalcon:
GCE and Oracle worked for me

TheRealFalcon (approved):

PR #43, opened 2020-11-05
Remove unnecessary oci RUNNING launch check

TheRealFalcon:
When launching an oci instance, there is a check to ensure the state is
RUNNING before it is returned. I removed the check because If wait=False
this won't work

### DIFF
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -135,13 +135,7 @@ class OCI(BaseCloud):
             raise Exception(
                 "Unable to retrieve instance with id: {} . "
                 "Is it a valid instance id?".format(instance_id)) from e
-        state = instance_data.lifecycle_state
-        if state != 'RUNNING':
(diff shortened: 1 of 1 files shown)


pycloudlib/oci/cloud.py
```diff
@@ -135,13 +135,7 @@ def get_instance(self, instance_id):
             raise Exception(
                 "Unable to retrieve instance with id: {} . "
                 "Is it a valid instance id?".format(instance_id)) from e
-        state = instance_data.lifecycle_state
```
OddBloke, line 138: CI is failing because `instance_data` is now unused; this LGTM otherwise.

TheRealFalcon: I removed the variable but want to keep the call to make sure if we're calling `get_instance` manually that we're passing in a valid `instance_id`

OddBloke (approved):
Thanks!

PR #45, opened 2020-11-12
Set correct xenial image for LXD

lucasmoura:
We are only using the `images:ubuntu/16.04/cloud` when we are creating virtual machines. However, this is wrong, since we always need to use that machine, for containers and virtual machines. The reason for that is that other xenial images will not have the necessary requirements to allow the use of `lxc exec` commands

### DIFF
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -36,7 +36,7 @@ class LXD(BaseCloud):
     _daily_remote = 'ubuntu-daily'
     _releases_remote = 'ubuntu'
 
-    XENIAL_VM_IMAGE = "images:ubuntu/16.04/cloud"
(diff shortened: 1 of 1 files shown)


pycloudlib/lxd/cloud.py
```diff
@@ -388,7 +388,7 @@ def _search_for_image(
             string, LXD fingerprint of latest image
 
         """
-        if is_vm and release == "xenial":
+        if release == "xenial":
             # xenial needs to launch images:ubuntu/16.04/cloud
             # because it contains the HWE kernel which has vhost-vsock support
             return self.XENIAL_VM_IMAGE
```
blackboxsw, line 394: let's rename this to XENIAL_IMAGE_VSOCK_SUPPORT

lucasmoura:
@blackboxsw Updated

blackboxsw (approved):
LGTM

PR #49, opened 2020-11-12
Retry cloudinit wait on all LXD instance types

lucasmoura:
Currently, the `images:ubuntu/16.04/cloud` errors out when we try to run the `_wait_for_cloudinit`  method. If we retry on that first error, the instance is launched without an issue. To address that, we are relaxing the retry check we have on `_wait_for_cloudinit`

### DIFF
--- a/pycloudlib/instance.py
+++ b/pycloudlib/instance.py
@@ -99,6 +99,7 @@ class BaseInstance(ABC):
             complete successfully.
         """
         self._wait_for_instance_start()
+        self._wait_for_execute()
         self._wait_for_cloudinit(raise_on_failure=raise_on_cloudinit_failure)
 
     @abstractmethod
@@ -361,6 +362,32 @@ class BaseInstance(ABC):
         self._tmp_count += 1
         return path
 
+    def _wait_for_execute(self):
+        """Wait until we can execute a command in the instance."""
+        self._log.debug('_wait_for_execute to complete')
+
+        test_instance_command = "whoami"
+        result = self.execute(test_instance_command)
+        if result.failed:
+            retries = 10
+            while retries:
+                result = self.execute(test_instance_command)
+
+                if result.ok:
+                    break
+
+                retries -= 1
+                time.sleep(10)
+
+        if result.failed:
+            raise OSError(
+                "{}\n{}".format(
+                    "Instance can't be reached",
+                    "Failed to execute {} command".format(
+                        test_instance_command)
+                )
+            )
+
     def _wait_for_cloudinit(self, *, raise_on_failure: bool):
         """Wait until cloud-init has finished.
 
@@ -369,7 +396,9 @@ class BaseInstance(ABC):
             then this method will raise an `OSError`.
         """
         self._log.debug('_wait_for_cloudinit to complete')
-        has_wait = "--wait" in self.execute("cloud-init status --help").stdout
+
+        result = self.execute("cloud-init status --help")
+        has_wait = "--wait" in result.stdout
 
         if has_wait:
             cmd = ["cloud-init", "status", "--wait", "--long"]
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -354,7 +354,6 @@ class LXDInstance(BaseInstance):
                     # lxd vms. However, if we wait a moment before trying
                     # again, the error fades away. That's why we are treating
                     # it here.
-                    print("cloud-init failed to start: out: ......." in str(e))
                     if "cloud-init failed to start: out: ......." in str(e):
                         time.sleep(sleep_time)
                         continue
--- a/pycloudlib/tests/test_instance.py
+++ b/pycloudlib/tests/test_instance.py
@@ -67,6 +67,7 @@ class TestWait:
         with mock.patch.multiple(
             instance,
             _wait_for_instance_start=mock.DEFAULT,
+            _wait_for_execute=mock.DEFAULT,
             _wait_for_cloudinit=mock.DEFAULT,
         ) as mocks:
             if raise_on_cloudinit_failure is not None:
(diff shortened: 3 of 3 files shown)


pycloudlib/lxd/instance.py
```diff
@@ -351,11 +351,10 @@ def _wait_for_cloudinit(self, *, raise_on_failure: bool):
...
 
                     # We are getting those types of messages when booting
-                    # lxd vms. However, if we wait a moment before trying
+                    # lxd images. However, if we wait a moment before trying
                     # again, the error fades away. That's why we are treating
                     # it here.
-                    print("cloud-init failed to start: out: ......." in str(e))
-                    if "cloud-init failed to start: out: ......." in str(e):
+                    if "cloud-init failed to start: out:" in str(e):
                         time.sleep(sleep_time)
```
blackboxsw, line 358: We probably need to log a debug level message here. ("Retrying on expected failure: %s", str(e)) so we can grok what is going on as this has a potential to keep retrying for other potentially unexpected failure cases where we should exit right?

blackboxsw (changes requested):

OddBloke:
Can you explain the issue this is intended to solve in more detail? The code being refactored makes reference to `lxd-agent`, but that's only present in LXD VMs, AIUI, so it's not clear what's being addressed for LXD containers.

lucasmoura:
Hi @OddBloke It turns out that the issue was related to that line of code here:
https://github.com/canonical/pycloudlib/blob/master/pycloudlib/instance.py#L372

It turns that on an early stage of boot that execute command was returning empty. In that situation, we ended up on the loop check which was producing an empty error msg:
```
OSError: cloud-init failed to start: out:  error:
```

By looping on `cloud-init --help`  until `result.ok` is True, solved that issue.

However, we are now discussing a secong issue that is only happening on bionic vms, where it seems that `cloud-init status --wait --long` return -1, but the error message is:
```
cloud-init failed to start: out: ...................................... error:
```

We originally thought it was a timeout issue, but that doesn't seen the case

pycloudlib/instance.py
```diff
@@ -369,7 +369,12 @@ def _wait_for_cloudinit(self, *, raise_on_failure: bool):
             then this method will raise an `OSError`.
         """
         self._log.debug('_wait_for_cloudinit to complete')
-        has_wait = "--wait" in self.execute("cloud-init status --help").stdout
+
+        result = self.execute("cloud-init status --help")
+        while result.failed:
+            result = self.execute("cloud-init status --help")
```
OddBloke, line 375: This could probably use an overall timeout, so that we don't end up with, e.g., Travis jobs running indefinitely if `status --help` never succeeds.

lucasmoura:
@OddBloke code updated

pycloudlib/instance.py
```diff
@@ -369,7 +369,26 @@ def _wait_for_cloudinit(self, *, raise_on_failure: bool):
             then this method will raise an `OSError`.
         """
         self._log.debug('_wait_for_cloudinit to complete')
-        has_wait = "--wait" in self.execute("cloud-init status --help").stdout
+
+        result = self.execute("cloud-init status --help")
+
+        if result.failed:
```
OddBloke, line 375: This has a problem: `cloud-init status --help` on trusty returns non-zero:

```
root@prepared-gelding:~# cloud-init status --help
usage: cloud-init [-h] [--version] [--file FILES] [--debug] [--force]
                  {init,modules,query,single} ...
cloud-init: error: invalid choice: 'status' (choose from 'init', 'modules', 'query', 'single')
root@prepared-gelding:~# echo $?
2
```

So for trusty systems (or any others with a pre-`status` cloud-init), this is always going to wait ~100s and then raise an `OSError`. This is why we were only checking for the text previously: the actual status code is orthogonal to what we're determining.

I think this may indicate a broader problem with this proposal: it's trying to combine a "can we run commands on this system yet" check with a "does cloud-init have status" check, and the two aren't compatible: a non-zero return code is not exceptional for the "status" check, but it _is_ for "can we run commands".

I think you might have more success if you try implementing this check with a command that has a stable exit code across all supported platforms/distros/releases: if `self.execute("whoami")` exits non-zero then it is near certain that that's an issue with the execution mechanism.

(And then, once we've established that the execution mechanism is working, then we can safely call `cloud-init status --help` as we were previously.)

lucasmoura: Yes, you are completely right here. I missed the trusty issue that you mentioned. I will update the code with your suggestion

lucasmoura:
@OddBloke Thanks for that catch. I have updated the code

pycloudlib/instance.py
```diff
@@ -369,7 +369,32 @@ def _wait_for_cloudinit(self, *, raise_on_failure: bool):
...
+                time.sleep(10)
+
+            if result.failed:
+                raise OSError(
+                    "{}\n{}".format(
+                        "Instance can't be reached",
+                        "Failed to execute {} command".format(
+                            test_instance_command)
+                    )
+                )
```
OddBloke, line 394: Thanks for this update, this looks correct to me now. This feels like it could quite naturally be refactored to a separate method now: `_wait_for_execute` or something along those lines. And it might make sense to call it from `BaseInstance.wait` instead of from `BaseInstance._wait_for_cloudinit` as it's not particularly related to cloud-init any longer?

pycloudlib/lxd/instance.py
```diff
@@ -354,7 +354,6 @@ def _wait_for_cloudinit(self, *, raise_on_failure: bool):
                     # lxd vms. However, if we wait a moment before trying
```
OddBloke, line 354: Does the retrying we're implementing elsewhere mean that this retrying is no longer necessary?

lucasmoura: Unfortunately no. This is related to that second problem I commented regarding `cloud-init status --wait --long` returning `-1` in the middle of the operation. We still don't know what is causing that issue

lucasmoura:
@OddBloke Code updated

OddBloke (approved):
This LGTM now (pending Chad's comment being resolved).

blackboxsw (approved):
Good review @OddBloke my comment was not applicable to this better approach. Thanks @lucasmoura

Issue #48, opened 2020-11-12
Launching xenial LXD VM images should not default to the images: remote


OddBloke:
The expectation of the cloud-init integration tests is that pycloudlib will operate on official Ubuntu cloud images by default. This is also [the documented expectation in the design doc](https://pycloudlib.readthedocs.io/en/latest/design.html#images):

> Instances are expected to use the latest daily image, unless another image is specifically requested.

For LXD, that means fetching images from the `ubuntu-daily:` remote (and not from `images:`).

(As UA client has specific requirements that diverge from this sensible default behaviour, I believe its testing code should be specifically requesting an image that meets those requirements, rather than modifying the default behaviour for all consumers.)

OddBloke:
(@TheRealFalcon FYI, this is a blocker for moving the integration tests to be voting for cloud-init, IMO: they aren't testing what they should be testing (and what the existing tests are testing).)

TheRealFalcon:
The 'images:' remote is only being used for xenial VMs running on LXD, correct? IIRC there's an issue that prevents xenial from running on an LXD VM at all, so unless there's a reasonable workaround, it's this or no Xenial at all.

OddBloke:
> The 'images:' remote is only being used for xenial VMs running on LXD, correct? IIRC there's an issue that prevents xenial from running on an LXD VM at all, so unless there's a reasonable workaround, it's this or no Xenial at all.

Not as of https://github.com/canonical/pycloudlib/commit/e9791a95d05116b9693249aaf2f8abee1451bd4a, no.

OddBloke:
For reference:

```sh
$ CLOUD_INIT_OS_IMAGE=xenial tox -r -e integration-tests -- tests/integration_tests/bugs/test_lp1886531.py
...
INFO     integration_testing:conftest.py:76 Setting up environment for lxd_container
INFO     integration_testing:conftest.py:102 Done with environment setup
INFO     integration_testing:clouds.py:64 Launching instance with launch_kwargs:
image_id=images:ubuntu/16.04/cloud
user_data=#cloud-config
bootcmd:
- rm -f /etc/fstab

wait=False
```

OddBloke:
This is now causing cloud-init integration test failures: our apt tests pass configuration which will render apt unable to update or install (specifically because we configure `squid.internal` as a proxy to test that configuration is written correctly). This means that when we attempt to install openssh-server, it fails (and we never gain SSH access to the system).

blackboxsw:
[...]

Issue #47, opened 2020-11-12
lxd.image_serial(images) doesn't work with VM images


paride:
```
In [3]: img = lxd.daily_image(release="focal", is_vm=True)

In [4]: lxd.image_serial(img)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-4-9f0c395b0f4b> in <module>
----> 1 lxd.image_serial(img)

~/canonical/venv/pycloudlib/lib/python3.8/site-packages/pycloudlib/lxd/cloud.py in image_serial(self, image_id)
    464         image_info = self._image_info(image_id)
    465 
--> 466         return image_info[0]['version_name']
    467 
    468     def delete_image(self, image_id):

IndexError: list index out of range
```

TheRealFalcon:
This appears to have been fixed by #52 .

PR #52, opened 2020-11-13
Fix lxd related issues

### LINKED ISSUE #46: lxd launch() fails if name parameter is not specified
When launching an LXD instance without passing the `name` parameter to `launch()` it fails with error:
```
AttributeError: 'NoneType' object has no attribute 'encode'
```
Pycloudlib should handle this case better, falling back to the container/vm name automatically generated by lxd.

lucasmoura:
This PR fixes problems that we are experiencing after merging #40 and #45.

We are fixing:

* LXD not being able to launch images without name parameter
* LXD always using images remote for xenial instances
* Paramiko not adding the ssh key type to the public key content
* LXD image_serial not being able to verify if a image id is related to a vm or not.

PS: I still need to update our integration tests to avoid mistakes like this in the future. However, since I don't know how fast I will be able to do that, I think there is no need to wait for that to land this PR.

Fixes #46 #47 #48

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,35 @@
+language: python
+dist: bionic
+
+
+install:
+  # Required so `git describe` will definitely find a tag; see
+  # https://github.com/travis-ci/travis-ci/issues/7422
+  - git fetch --unshallow
+
+matrix:
+  fast_finish: true
+  include:
+    - name: run-lxd-example
+      install:
+          - pip install -r requirements.txt
+      script:
+          - sudo apt-get remove --yes --purge lxd lxd-client
+          - sudo rm -Rf /var/lib/lxd
+          - sudo snap install lxd
+          - sudo lxd init --auto
+          - sudo usermod -a -G lxd $USER
+          - sg lxd -c "python examples/lxd.py"
+    - name: run-cloudinit-integration-tests
+      install:
+          - pip install -r requirements.txt
+          - pip install pytest
+      script:
+          - sudo apt-get remove --yes --purge lxd lxd-client
+          - sudo rm -Rf /var/lib/lxd
+          - sudo snap install lxd
+          - sudo lxd init --auto
+          - sudo usermod -a -G lxd $USER
+          - git clone https://github.com/canonical/cloud-init.git
+          - cd cloud-init
+          - sg lxd -c "pytest tests/integration_tests/"
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -3,6 +3,7 @@
 import io
 import re
 import textwrap
+from abc import abstractmethod
 import paramiko
 
 from pycloudlib.cloud import BaseCloud
@@ -29,17 +30,12 @@ class UnsupportedReleaseException(Exception):
         )
 
 
-class LXD(BaseCloud):
-    """LXD Cloud Class."""
+class BaseLXD(BaseCloud):
+    """LXD Base Cloud Class."""
 
     _type = 'lxd'
     _daily_remote = 'ubuntu-daily'
     _releases_remote = 'ubuntu'
-
-    XENIAL_IMAGE_VSOCK_SUPPORT = "images:ubuntu/16.04/cloud"
-    VM_HASH_KEY = "combined_disk1-img_sha256"
-    TRUSTY_CONTAINER_HASH_KEY = "combined_rootxz_sha256"
-    CONTAINER_HASH_KEY = "combined_squashfs_sha256"
 
     def __init__(self, tag, timestamp_suffix=True):
         """Initialize LXD cloud class.
@@ -156,45 +152,14 @@ class LXD(BaseCloud):
         pub_key = key.get_base64()
         key.write_private_key(priv_str, password=None)
 
-        return pub_key, priv_str.getvalue()
-
-    def _extract_release_from_image_id(self, image_id, is_vm=False):
-        """Extract the base release from the image_id.
-
-        Args:
-            image_id: string, [<remote>:]<release>, what release to launch
-                     (default remote: )
-
-        Returns:
-            A string contaning the base release from the image_id that is used
-            to launch the image.
-        """
-        release_regex = (
-            "(.*ubuntu.*(?P<release>(" +
-            "|".join(UBUNTU_RELEASE_VERSION_MAP) + "|" +
-            "|".join(UBUNTU_RELEASE_VERSION_MAP.values()) +
-            ")).*)"
-        )
-        ubuntu_match = re.match(release_regex, image_id)
-        if ubuntu_match:
-            release = ubuntu_match.groupdict()["release"]
-            for codename, version in UBUNTU_RELEASE_VERSION_MAP.items():
-                if release in (codename, version):
-                    return codename
-
-        # If we have a hash in the image_id we need to query simplestreams to
-        # identify the release.
-        return self._image_info(image_id, is_vm)[0]["release"]
+        return "ssh-rsa {}".format(pub_key), priv_str.getvalue()
 
     # pylint: disable=R0914,R0912,R0915
-    def init(
+    def _prepare_command(
             self, name, release, ephemeral=False, network=None, storage=None,
             inst_type=None, profile_list=None, user_data=None,
-            config_dict=None, is_vm=False):
-        """Init a container.
-
-        This will initialize a container, but not launch or start it.
-        If no remote is specified pycloudlib default to daily images.
+            config_dict=None):
+        """Build a the command to be used to launch the LXD instance.
 
         Args:
             name: string, what to call the instance
@@ -207,12 +172,10 @@ class LXD(BaseCloud):
             profile_list: list, optional, profile(s) to use
             user_data: used by cloud-init to run custom scripts/configuration
             config_dict: dict, optional, configuration values to pass
-            is_vm: boolean, optional, defines if a virtual machine will
-                   be created
 
         Returns:
-            The created LXD instance object
-
+            A list of string representing the command to be run to
+            launch the LXD instance.
         """
         profile_list = profile_list if profile_list else []
         config_dict = config_dict if config_dict else {}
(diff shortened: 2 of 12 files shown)


pycloudlib/lxd/cloud.py
```diff
@@ -448,11 +459,13 @@ def _image_info(self, image_id, is_vm=False):
 
         return image_info
 
-    def image_serial(self, image_id):
+    def image_serial(self, image_id, is_vm=False, **kwargs):
```
OddBloke, line 462: I'm a little concerned about adding cloud-specific keyword arguments to these APIs. In particular, doing so with `is_vm` means that code that's written against the pure `BaseCloud` interface needs special handling code to deal with LXD VM instances.

Consider this simple launch example (inspired by cloud-init's integration testing):

```py
def do_launch(cloud: BaseCloud, image_id: str) -> BaseInstance:
    image_id = cloud.released_image(image_id)
    instance = cloud.launch(image_id, wait=False)
    instance.wait(raise_on_cloudinit_failure=False)
    return instance
```

This can be used to launch EC2 or LXD containers:

```py
do_launch(pycloudlib.ec2.cloud.EC2("our_tag"))
do_launch(pycloudlib.lxd.cloud.LXD("our_tag"))
```

but there's no way to launch a LXD VM using it. It would need modification to take `kwargs` (because it doesn't make sense to have `is_vm` on a function which handles non-LXD clouds too), and of course its callers would also need special handling to pass in `kwargs={"is_vm": True}` iff they were launching a LXD VM. This would propagate all the way out through APIs which otherwise could just be passing around a `BaseCloud` instance.

It feels to me like we could modify this abstraction to improve matters. We could have `is_vm` be an instance attribute, and so passed at instantiation time; this would mean that instances of `BaseCloud` could still be passed around. However, that still doesn't allow us to pass around `BaseCloud` _subclasses_ and instantiate them at will: you have to carry `(BaseCloudSubclass, {"kwargs": "to instantiate it with"})` around to be able to do that.

So I think that perhaps the better abstraction is to have separate classes for `LXDContainer` and `LXDVirtualMachine`. This would allow them both to be treated as equal, first-class "clouds", and would enable use cases like those described above. These would share a lot of their implementation, of course. This also follows the abstraction that James settled on in the cloud-init integration testing: we have a `LxdContainerCloud` (which wraps `pycloudlib.lxd.cloud.LXD`), not a `LxdCloud`.

What do folks think?

TheRealFalcon: I like this proposal. Even though much of the implementation is the same, I think it makes sense to think of them as two separate "clouds". It also allows for further divergence in implementation.

OddBloke: As further background, LXD has `instance_type` and `type` separate in its API: https://linuxcontainers.org/lxd/docs/master/rest-api#post-. This indicates, to me, that container vs VM is a more fundamental difference in their data model than "instance type" is in most other "clouds", which I think supports the use of separate classes in our own data model.

lucasmoura: I think it is a great idea, since you totally right that the `is_vm` modification is breaking the main abstraction of pycloudlib (Which I missed when implementing that). I can update the code without an issue

pycloudlib/lxd/cloud.py
```diff
@@ -238,11 +241,19 @@ def init(
                 profile_list = [profile_name]
 
         if self.key_pair:
+            pub_key = self.key_pair.public_key_content
+
+            # When we create keys through paramiko, we end up not
+            # having the key type on the public key content. Because
+            # of that, we are manually adding the ssh-rsa type into it
+            if "ssh-" not in pub_key:
+                pub_key = "ssh-rsa {}".format(pub_key)
```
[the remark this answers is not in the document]
lucasmoura: Totally right, It is definitely better to update the method that creates the key instead

pycloudlib/lxd/cloud.py
blackboxsw: @OddBloke I was originally likening LXD deployment of vms vs containers as sort of like ec2 managing different instance_types. Generic LXD cloud provides an wrapper around the lxc command set which is used by both lxc --vm and container based instance launches.

TLDR: +1 on  LXDContainer/LXDVirtualmachine subclasses.

I spent a good hour trying to justify a common single LXD cloud as it is abstracting the common CLI for vm and container deployments,  but a couple of things swayed me to subclasses (other than your well thought out explanation)
   - vms vs containers requiring special image filtering, search and launch knowledge,
   - vms require specific lxc profile love during setup and deployment
   - method signature bloat adding in LXCCloud is_vm param into multiple LXDCloud methods such as init, __init__, launch and image* methods indicating enough of a smell that this is better separated from a common class.

I also thought (incorrectly) that `--vm` was nearly a simple alias for `--type virtual-machine`. 

But, when looking at `lxc --type` vs `lxc --vm`, as Dan pointed out in MM, one is not --vm is not actually an alias --type virtual-machine. The two params provide [separate fields](https://github.com/lxc/lxd/blob/master/lxc/init.go#L208-L216) to [lxd's API](https://linuxcontainers.org/lxd/docs/master/rest-api#post-) that direct the underlying cloud platform chosen for the instance deployment.

Through @lucasmoura's good work, we are seeing enough discrepancies in these underlying platforms and how to query images as well as which profiles to use that probably warrant the subclasses suggestion.
Ultimately, under the hood below lxc's CLI, they are unique cloud platforms and probably should be treated as such.

lucasmoura:
@OddBloke @blackboxsw @TheRealFalcon I have refactored the code to separate the behavior of LXD containers from virtual machines.

I believe now we can launch LXD vms as we launch any other cloud in the codebase.

However, there are some abstractions that maybe could be improved, such as the use of `_get_image_hash_key`. But if anyone has any idea for improvements, just let me know and I will code them.

Also @paride we have kind of an exception for `image_serial` when we are using Xenial vms. If you want to change how to handle that situation, just let me know as well.

paride:
@lucasmoura I think the exception for Xenial is fine. I'm also +1 for the container/vm split, I think it makes sense and helps maintaining the api consistent. I'm going to run a couple of bootspeed jobs against this branch to check that everything works and report back.

paride:
Works nicely!

pycloudlib/lxd/cloud.py
```diff
@@ -511,3 +494,220 @@ def _find_image(self, release, arch=LOCAL_UBUNTU_ARCH, daily=True):
...
+            arch: string, architecture to use
+
+        Returns:
+            string, LXD fingerprint of latest image
+
+        """
+        if release == "xenial":
+            # xenial needs to launch images:ubuntu/16.04/cloud
+            # because it contains the HWE kernel which has vhost-vsock support
+            return self.XENIAL_IMAGE_VSOCK_SUPPORT
```
OddBloke, line 684: This is a different image to the one that consumers will expect (it's from `images:`, instead of from `ubuntu:`); I'd like to see us log something to give people a better chance of grokking what's going on.

pycloudlib/lxd/cloud.py
```diff
@@ -511,3 +494,220 @@ def _find_image(self, release, arch=LOCAL_UBUNTU_ARCH, daily=True):
...
+
+        Args:
+            image_id: string, LXD image fingerprint
+
+        Returns:
+            string, serial of latest image
+
+        """
+        if image_id == self.XENIAL_IMAGE_VSOCK_SUPPORT:
+            return ""
```
OddBloke, line 711: Why `""` instead of `None`?

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+          - pip install pytest
+      script:
+          - sudo apt-get remove --yes --purge lxd lxd-client
+          - sudo rm -Rf /var/lib/lxd
+          - sudo snap install lxd
+          - sudo lxd init --auto
+          - sudo usermod -a -G lxd $USER
+          - git clone https://github.com/canonical/cloud-init.git
+          - cd cloud-init
+          - sg lxd -c "pytest tests/integration_tests/"
```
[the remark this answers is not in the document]
OddBloke: (Except tox might have a hard time using the current pycloudlib rather than cloud-init's default one, so I'll give this further thought.)

OddBloke (commented):
Thanks Lucas, this looks much cleaner and clearer now. I still have a few inline comments/questions, but none of them should require a refactor!

.travis.yml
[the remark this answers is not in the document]
lucasmoura: Yes, that was exactly why I have not used tox :(

pycloudlib/lxd/tests/test_cloud.py
```diff
@@ -95,16 +96,16 @@ def test_extract_release_from_non_hashed_image_id(
         self, image_id, expected_release
     ):  # pylint: disable=W0212
         """Tests extracting release from non hashed image id."""
-        cloud = LXD(tag="test")
+        cloud = LXDVirtualMachine(tag="test")
```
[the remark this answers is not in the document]
lucasmoura: This function should only be run on `LXDVirtualMachine` instances. I will better document that on the test

pycloudlib/lxd/cloud.py
lucasmoura: To be honest, I thought that maybe one of ours scripts were always expecting a string to be returned here, but I am not sure if that is okay. @paride would it be okay if this method returns `None`  for xenial images on vms ?

lucasmoura:
@OddBloke Code Updated. I will just wait on @paride response regarding the `image_serial` change

pycloudlib/lxd/cloud.py
paride: @lucasmoura the bootspeed measurement script does indeed expect a string, but I agree with Dan in that it should handle `None`, so +1 for Dan's suggestion. (In practice this is a non-issue as the bootspeed jobs do not measure Xenial, but I like to know it's working in case we want to enable it.)

TheRealFalcon (approved):

lucasmoura:
@OddBloke Code update with @paride approval on the updated `image_serial`

OddBloke (approved):
Thanks Lucas, this LGTM now!

PR #56, opened 2020-11-18
Run lxc commands as non root by default

lucasmoura:
In ubuntu-advantage-tools, we require some LXD commands to be run as `non-root`. This PR changes the default behavior of LXD to run exec commands as non-root.

This will impact some `cloud-init` tests, but we already have a [PR](https://github.com/canonical/cloud-init/pull/664) to handle that

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -23,6 +23,7 @@ class LXDInstance(BaseInstance):
         super().__init__(key_pair=key_pair)
 
         self._name = name
+        self._user_id = None
 
     def __repr__(self):
         """Create string representation for class."""
@@ -33,7 +34,9 @@ class LXDInstance(BaseInstance):
         if self.key_pair:
             return super()._run_command(command, stdin)
 
-        base_cmd = ['lxc', 'exec', self.name, '--']
+        base_cmd = [
+            'lxc', 'exec', self.name, '--', 'sudo', '-u', self.username, '--'
+        ]
         return subp(base_cmd + list(command), rcs=None)
 
     @property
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
Actually...lxd is the exception for us as all other clouds already execute non-root. This may actually simplify things for us :slightly_smiling_face:

TheRealFalcon (approved):
Travis is failing as this will break cloud-init tests, but once https://github.com/canonical/cloud-init/pull/664 lands it should be a one-line fix. Since we've pinned our pycloudlib version, I'm not concerned these failures.

blackboxsw (approved):
Once https://github.com/canonical/cloud-init/pull/694 lands, this looks good to land

OddBloke (approved):
The conclusion we came to in https://github.com/canonical/cloud-init/pull/664 was that `pycloudlib` should consistently execute commands as an unprivileged user by default. This aligns LXD with the other clouds in pycloudlib which use SSH.

OddBloke (changes requested):
Per https://github.com/canonical/cloud-init/pull/694#discussion_r531159622, running commands as UID 1000 is _not_ the same as running commands as `self.username`. I believe this will need modification to look up the UID of `self.username` before running `lxc exec` for the first time.

lucasmoura:
@OddBloke I have updated the code to get the `user id`  before running any `lxc exec` commands.

The solution I have is to retry the `id -u` command until the user was created. With that solution, I think we can drop from LXD the call for [_wait_for_execute](https://github.com/canonical/pycloudlib/blob/master/pycloudlib/instance.py#L369), since the `user_id` property is kind of covering that already

pycloudlib/lxd/instance.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, name, key_pair=None):
         super().__init__(key_pair=key_pair)
 
         self._name = name
+        self._user_id = None
```
OddBloke, line 26:
```suggestion
```

OddBloke (approved):
Thanks!

PR #57, opened 2020-11-18
Allow _wait_for_cloudinit to work with LXD VMs

TheRealFalcon:
I tried using LXD VMs in cloud-init integration tests, but was running into these issues. I don't think we'd expect to see these errors unless it's a cloud-init VM starting, so I think we can safely ignore them, even if we set `raise_on_failure` to True. If it's not one of these two errors, then we still get the actual cloud-init error.

@OddBloke You initially added the `raise_on_failure` stuff. Is what I'm saying here make sense?


TheRealFalcon:
Nevermind, I forced push a few changes till we were back to the initial implementation :smile:

PR #58, opened 2020-11-19
Add additional LXD VM error to ignore

TheRealFalcon:
When `raise_on_error` is false, we still need to wait until the LXD VM is ready and ignore some of these startup errors. Also adding `Error: Instance is not running` to the errors list

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -337,28 +337,29 @@ class LXDInstance(BaseInstance):
             When `True`, if the process waiting for cloud-init exits non-zero
             then this method will raise an `OSError`.
         """
-        if self.is_vm and raise_on_failure:
+        if self.is_vm:
             retries = [30, 45, 60, 75, 90, 105]
 
             for sleep_time in retries:
                 try:
                     super()._wait_for_cloudinit(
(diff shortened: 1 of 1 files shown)


pycloudlib/lxd/instance.py
```diff
@@ -337,28 +337,29 @@ def _wait_for_cloudinit(self, *, raise_on_failure: bool):
...
-                    if "Failed to connect to lxd-agent" in str(e):
-                        time.sleep(sleep_time)
-                        continue
-
                     # We are getting those types of messages when booting
                     # lxd vms. However, if we wait a moment before trying
                     # again, the error fades away. That's why we are treating
                     # it here.
-                    if "cloud-init failed to start: out: ......." in str(e):
+                    lxd_errors = [
```
blackboxsw, line 353: I waffled between making this error prefix list more visible like a global or a class attr `retry_on_errors` ... But since it's only used here, I guess that doesn't make much sense on this pass. If we have a more general need for different clouds to handle/retry error types we can address that then.

blackboxsw (approved):
This was a whoops in initial implementation. Thanks for the fix. My inline comment need not be addressed until we have a need on more than one cloud to retry on specific launch errors. 
LGTM

PR #69, opened 2020-11-26
Update travis to run cloudinit test with daily ppa

lucasmoura:
Our current travis build is running the cloud-init integration tests with the current cloud-init version. This is problematic
because some of the tests on cloud-init assume that we are using the most recent version of the package. To fix that, we
will be using the cloud-init package from our daily ppa to run the integration tests


OddBloke (approved):
Thanks!

PR #70, opened 2020-12-04
Add wait_for_stop implementations

TheRealFalcon:
Azure and LXD need wait_for_stop implementations.

### DIFF
--- a/pycloudlib/azure/instance.py
+++ b/pycloudlib/azure/instance.py
@@ -1,6 +1,7 @@
 # This file is part of pycloudlib. See LICENSE file for license information.
 """Azure instance."""
 
+import time
 from pycloudlib.instance import BaseInstance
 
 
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
We should also make our timeout/retry logic more consistent/configurable through the whole API (https://pypi.org/project/retry/ might help?), but I don't think this is the PR to do it.

lucasmoura (approved):

pycloudlib/azure/instance.py
```diff
@@ -30,7 +31,16 @@ def wait_for_delete(self):
...
+        for _ in range(100):
+            power_state = self._client.virtual_machines.get(
+                resource_group_name=self._instance['rg_name'],
+                vm_name=self.name,
+                expand='instanceView'
+            ).instance_view.statuses[1].display_status
+            if power_state == 'VM stopped':
+                return
+            time.sleep(1)
+        raise TimeoutError
```
lucasmoura, line 43: Maybe we could alert the user the context of the TimeoutError, like indicating that it was during a `wait_for_stop`  operation

OddBloke (approved):
Thanks!

PR #77, opened 2020-12-10
Some Azure updates

TheRealFalcon:
- Add ability to make more granular changes to vm_params when creating
an instance
- When deprovisioning an image for snapshot generation,
add option to keep the original user instead of automatically
removing it
- Small bugfix to not throw exception if image offer is None

### DIFF
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -11,7 +11,7 @@ import pycloudlib.azure.util as util
 from pycloudlib.cloud import BaseCloud
 from pycloudlib.azure.instance import AzureInstance
 from pycloudlib.key import KeyPair
-from pycloudlib.util import get_timestamped_tag
+from pycloudlib.util import get_timestamped_tag, update_nested
 
 
 class Azure(BaseCloud):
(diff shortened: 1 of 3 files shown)


pycloudlib/azure/cloud.py
```diff
@@ -376,7 +376,7 @@ def _create_vm_parameters(self, name, image_id, nic_id, user_data):
         return vm_parameters
 
     def _create_virtual_machine(
-            self, image_id, nic_id, user_data, name, **kwargs
+        self, image_id, nic_id, user_data, name, vm_params=None
```
TheRealFalcon, line 379: @blackboxsw or @lucasmoura Would this break anything in ua-client? Are you guys overriding Azure vm_params via here (and line 404) anywhere?

blackboxsw: Safe from uaclient standpoint, integration tests  only currently call launch with image_id and user_data params and snapshot(instance=X).

pycloudlib/azure/util.py
```diff
@@ -179,7 +179,7 @@ def is_pro_image(image_id, registered_image):
     if img_dict.get("publisher") == "Canonical":
         offer = img_dict["offer"]
     elif registered_image is not None:
-        offer = registered_image["offer"]
+        offer = registered_image["offer"] or ""
```
blackboxsw, line 182: Any reason we pefer this syntax to registered_image.get("offer", "") ?

TheRealFalcon: Yeah, I tried that syntax, but `offer` actually exists as a key, but the value is `None`. The return statement fails trying to iterate across None.

blackboxsw (approved):
LGTM!

PR #78, opened 2020-12-14
Add groovy to azure

TheRealFalcon:


mitechie (approved):
Did we need them on any of the other clouds?

TheRealFalcon:
We shouldn't. I didn't test all the clouds, but Dan recently fixed LXD, Oracle doesn't provide Groovy images, so that leaves ec2 and gce, and they both rely on simple streams.

PR #79, opened 2020-12-16
Allow GCE cloud to create ssh keys

lucasmoura:
Currently, there is no way for the GCE cloud instance to create ssh keys to be used on the instances. We are now adding that possibility. Also, we are changing the user related to the ssh key from admin to ubuntu, to be consistent with the ssh behavior of other cloud instances


blackboxsw (approved):
LGTM

PR #80, opened 2020-12-17
Make GCE cloud set environment variables

lucasmoura:
Right now, we need to explicitly set some environment variables to successfully create GCE instances. We are now setting those variables inside the GCE cloud

### DIFF
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -27,7 +27,8 @@ class GCE(BaseCloud):
     _type = 'gce'
 
     def __init__(
-        self, tag, timestamp_suffix=True, project=None, region=None, zone=None
+        self, tag, timestamp_suffix=True, credentials_path=None, project=None,
+        region="us-west2", zone="a"
     ):
         """Initialize the connection to GCE.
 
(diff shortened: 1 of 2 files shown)


pycloudlib/gce/cloud.py
```diff
@@ -27,28 +27,35 @@ class GCE(BaseCloud):
...
         super().__init__(tag, timestamp_suffix)
         self._log.debug('logging into GCE')
 
+        if not all([credentials_path, project, region, zone]):
+            raise ValueError(
+                'credentials_path, project, region, and zone are required!')
+
+        os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = str(credentials_path)
+        os.environ["GOOGLE_CLOUD_PROJECT"] = str(project)
+
```
blackboxsw, line 53: With this we are mandating that the optional parameters are required.
I think instead we should "export" the params that are provided with something like the following:

```suggestion
if credentials_path:
    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = str(credentials_path)
if project:
    os.environ["GOOGLE_CLOUD_PROJECT"] = str(project)
```

blackboxsw (changes requested):
I think we need to still allow for folks to optional omit project and credentials_path

ubuntu-server-builder (approved):
+1 Thanks Lucas.

blackboxsw (approved):
LGTM

PR #84, opened 2021-01-06
instance: convert SSH key passphrase error to a warning

### LINKED ISSUE #82: RFC: Allow passphrased keys to be specified to pycloudlib
TL;DR: Scroll down to "I think there are three paths forward".

Currently, pycloudlib's default behaviour is to use `~/.ssh/id_rsa` as the SSH key that is provided to, and used to access, all machines under test, with one exception: pycloudlib assumes that a key named with the current username exists in EC2, and will not create it if it is absent. pycloudlib also explicitly checks that the given key does not have a passphrase on it: it raises an exception before we even attempt to connect to the remote SSH host if it does.

This poses a problem for me: on my primary system, I use a non-default key for my work SSH access (`~/.ssh/id_rsa.canonical`) and I have it protected (at rest) with a passphrase. I spent some time yesterday trying to implement support for my usecase in pycloudlib with limited success; this issue captures my findings/thinking.

paramiko can actually use my passphrased key without issues, if I remove the guard we have implemented. I run an SSH agent and populate it with my key at login, so I don't have to type the passphrase every time I use the key; paramiko's [`connect` docs](http://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect) states they will attempt auth with "Any key we can find through an SSH agent", after attempting the specified key. The first use of my key fails (because there's no passphrase to unlock it), and then it is successfully used when paramiko falls back to the SSH agent.

My initial implementation path was to completely rely on paramiko's [default connection behaviour](http://docs.paramiko.org/en/stable/api/client.html#paramiko.client.SSHClient.connect): it will use an SSH agent if available, falling back to looking at `~/.ssh/id_rsa`, which closely approximates our current behaviour. However, this runs into issues: for most clouds we need to provide the public key file's contents to the cloud platform on launch: if we don't have any reference to the key (because we're relying on our SSH agent), then we can't readily do that. (paramiko _does_ have an API for querying the SSH agent, but it's not clear how you would select the correct key without already having either the public key material or its filename; in either case, there's then no need to use the SSH agent.) So this isn't the way to go.

I think there are three paths forward:

1. continue to disallow the use of passphrased keys (i.e. change nothing). I don't consider this viable, we should not be forcing our users to create and/or use less secure SSH keys.
2. don't explicitly disallow the use of passphrased keys (i.e. remove the guard from `_ssh_connect`). This would enable my use-case (via agent fallback), but would mean that use of passphrased keys in the absence of an agent will lead to a ten minute retry loop before failing (vs. the immediate failure we have now).
3. handle the SSH key lifecycle entirely within pycloudlib: instead of using the user's existing keys at all, pycloudlib could generate SSH keys for each instance it launches, cleaning them up after use. These single-use keys don't need the protection that a user's primary keys do, so we could guarantee they have no passphrase.

I think that looking at cost/reward, (2) is probably our best path forward in the short/medium term. We can ensure that none of the keys used for automated test running have passphrases, and users for whom this would change behaviour (i.e. those with passphrased default keys) will (a) already be running into issues with the current codebase, and (b) be present to observe the retry looping and therefore able to intervene.

(Probably better than removing the guard is to instead have it emit some sort of warning, something like: "Specified key requires a passphrase; it will need to be added to a running ssh-agent for tests to run successfully".)

OddBloke:
This allows users to specify keys with passphrases, but clearly warns
them about what may happen if they haven't added the key to an SSH agent
(which removes the need for a passphrase when this code runs).

Fixes #82.

### DIFF
--- a/pycloudlib/instance.py
+++ b/pycloudlib/instance.py
@@ -308,8 +308,12 @@ class BaseInstance(ABC):
             paramiko.RSAKey.from_private_key_file(
                 self.key_pair.private_key_path)
         except PasswordRequiredException:
-            self._log.error('RSA Key requires password!')
-            raise
+            self._log.warning(
+                "The specified key (%s) requires a passphrase. If you have not"
+                " added this key to a running SSH agent, you will see failures"
+                " to connect after a long timeout.",
+                self.key_pair.private_key_path,
+            )
         except SSHException:
             pass
 


OddBloke:
With the following diff applied to the cloud-init integration testing codebase:

```diff
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -93,6 +93,7 @@ class IntegrationCloud(ABC):
         self.cloud_instance = self._get_cloud_instance()
         if settings.KEYPAIR_NAME is not None:
             self.cloud_instance.key_pair.name = settings.KEYPAIR_NAME
+        self.cloud_instance.use_key("/home/daniel/.ssh/id_rsa.canonical.pub", name="canonical")
         self._released_image_id = self._get_initial_image()
         self.snapshot_id = None
```

I see that my key is not rejected, and I instead get this message in my log output:

```
2021-01-06 18:05:49 WARNING   pycloudlib.instance:instance.py:311 The specified key (/home/daniel/.ssh/id_rsa.canonical) requires a passphrase. If you have not added this key to a running SSH agent, you will see failures to connect after a long timeout.
```

TheRealFalcon (approved):

PR #87, opened 2021-01-07
Get defaults from CLI for GCE and OCI if not provided

TheRealFalcon:
Stop requiring to pass in project for gce and compartment for oci

### DIFF
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -58,7 +58,7 @@ class GCE(BaseCloud):
                 "setup?\nCommand attempted: '{}'".format(' '.join(command))
             )
             try:
-                result = subp(command)
+                result = subp(command, rcs=())
             except FileNotFoundError as e:
                 raise Exception(exception_text) from e
             if not result.ok:
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -6,10 +6,12 @@ authentication into the cloud, finding an image, and launching an
 instance. It however, does not allow any further actions from occuring.
 """
 
+import json
 import logging
 import os
 import time
 from itertools import count
+from pathlib import Path
 
 import googleapiclient.discovery
 
@@ -19,6 +21,9 @@ from pycloudlib.gce.instance import GceInstance
 
 logging.getLogger('googleapiclient.discovery').setLevel(logging.WARNING)
 
+DEFAULT_CREDENTIALS_PATH = Path.home() / '.config' / 'gcloud' / \
+    'application_default_credentials.json'
+
 
 class GCE(BaseCloud):
     """GCE Cloud Class."""
@@ -26,7 +31,8 @@ class GCE(BaseCloud):
     _type = 'gce'
 
     def __init__(
-        self, tag, timestamp_suffix=True, credentials_path=None, project=None,
+        self, tag, timestamp_suffix=True,
+        credentials_path=DEFAULT_CREDENTIALS_PATH, project=None,
         region="us-west2", zone="a"
     ):
         """Initialize the connection to GCE.
@@ -36,16 +42,26 @@ class GCE(BaseCloud):
             timestamp_suffix: bool set True to append a timestamp suffix to the
                 tag
             credentials_path: path to credentials file for GCE
+                Defaults to default gcloud path.
             project: GCE project
+                Defaults to project supplied in 'credentials_path'
             region: GCE region
             zone: GCE zone
         """
         super().__init__(tag, timestamp_suffix)
         self._log.debug('logging into GCE')
 
-        if credentials_path:
+        credentials_path = Path(credentials_path)
+        if credentials_path.exists():
             os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = str(
                 credentials_path)
+            if project is None:
+                with credentials_path.open() as f:
+                    config = json.load(f)
+                if 'project_id' in config:
+                    project = config['project_id']
+                elif 'quota_project_id' in config:
+                    project = config['quota_project_id']
 
         if project:
             os.environ["GOOGLE_CLOUD_PROJECT"] = str(project)
@@ -55,6 +71,7 @@ class GCE(BaseCloud):
         self.compute = googleapiclient.discovery.build(
             'compute', 'v1', cache_discovery=False
         )
+
         self.project = project
         self.region = region
         self.zone = '%s-%s' % (region, zone)
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -6,23 +6,20 @@ authentication into the cloud, finding an image, and launching an
 instance. It however, does not allow any further actions from occuring.
 """
 
-import json
 import logging
 import os
 import time
 from itertools import count
-from pathlib import Path
 
 import googleapiclient.discovery
 
 from pycloudlib.cloud import BaseCloud
 from pycloudlib.gce.util import raise_on_error
 from pycloudlib.gce.instance import GceInstance
+from pycloudlib.util import subp
+
 
 logging.getLogger('googleapiclient.discovery').setLevel(logging.WARNING)
-
-DEFAULT_CREDENTIALS_PATH = Path.home() / '.config' / 'gcloud' / \
-    'application_default_credentials.json'
 
 
 class GCE(BaseCloud):
@@ -31,8 +28,7 @@ class GCE(BaseCloud):
     _type = 'gce'
 
     def __init__(
-        self, tag, timestamp_suffix=True,
-        credentials_path=DEFAULT_CREDENTIALS_PATH, project=None,
+        self, tag, timestamp_suffix=True, credentials_path=None, project=None,
         region="us-west2", zone="a"
     ):
         """Initialize the connection to GCE.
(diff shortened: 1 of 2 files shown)


pycloudlib/gce/cloud.py
```diff
@@ -47,8 +48,22 @@ def __init__(
         self.compute = googleapiclient.discovery.build(
             'compute', 'v1', cache_discovery=False
         )
-        if not all([project, region, zone]):
-            raise ValueError('project, region, and zone are required!')
+        if not all([region, zone]):
+            raise ValueError('region, and zone are required!')
+        if not project:
+            command = ['gcloud', 'config', 'get-value', 'project']
```
lucasmoura, line 54: Another approach we could follow is to read the `project` directly from the json provided by `GOOGLE_APPLICATION_CREDENTIALS`. In that way, we would not need to setup up `gcloud`.

TheRealFalcon:
@lucasmoura Thanks for the tips. I've pushed a new version that uses the google credentials if provided, and also provides a default (the default for gcloud) that can be overridden.

pycloudlib/gce/cloud.py
```diff
@@ -20,14 +22,18 @@
 
 logging.getLogger('googleapiclient.discovery').setLevel(logging.WARNING)
 
+DEFAULT_CREDENTIALS_PATH = Path.home() / '.config' / 'gcloud' / \
+    'application_default_credentials.json'
```
OddBloke, line 26: I'm not 100% sure that this is the correct path: it does exist on my desktop, but if I initialise a new `gcloud` inside a container (to the point where I can list my instances), I don't see it:

```
# tree .config/gcloud/
.config/gcloud/
├── access_tokens.db
├── active_config
├── cache
│   └── daniel.watkins@canonical.com
├── config_sentinel
├── configurations
│   └── config_default
├── credentials.db
├── gce
├── legacy_credentials
│   └── daniel.watkins@canonical.com
│       └── adc.json
└── logs
    └── 2021.01.08
        ├── 17.27.25.441108.log
        ├── 17.27.37.404385.log
        ├── 17.27.41.272001.log
        ├── 17.28.11.613997.log
        ├── 17.28.16.418870.log
        ├── 17.28.24.243745.log
        └── 17.28.26.915831.log
```

`adc.json` _does_ look like the same file format (and `adc` is presumably an abbreviation for `application_default_credentials`?).

pycloudlib/gce/cloud.py
```diff
@@ -37,16 +43,26 @@ def __init__(
...
 
-        if credentials_path:
+        credentials_path = Path(credentials_path)
+        if credentials_path.exists():
             os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = str(
                 credentials_path)
+            if project is None:
+                with credentials_path.open() as f:
+                    config = json.load(f)
+                if 'project_id' in config:
```
OddBloke, line 62: I'm not sure this is correct either; in both my desktop and in-container ADC file, I have only 4 keys: `client_id`, `client_secret`, `refresh_token`, and `type`.

When prompted to set my project in the container, it was set in `.config/gcloud/configurations/config_default` (which appears to be selected by `default` being in `.config/gcloud/active_config`).

(I believe that this file is used for authenticating your user to the Google APIs, which is orthogonal to the project being used: users have access to many projects.)

pycloudlib/oci/cloud.py
```diff
@@ -37,6 +38,19 @@ def __init__(
...
+            exception = Exception(
+                "Could not obtain OCI compartment id. Has the CLI client been "
+                "setup?\nCommand attempted: '{}'".format(' '.join(command))
+            )
+            try:
+                result = subp(command)
+            except FileNotFoundError as e:
+                raise exception from e
+            if not result.ok:
+                raise exception
```
OddBloke, line 52: This is nit-level, so don't worry too much about it: will this include the output from the failing command? (That would save folks from having to re-run the command to see what the exact error was.)

OddBloke (changes requested):
Thanks for this, James! I think your initial implementation of the Google stuff is closer to what we want: comments inline, of course.

TheRealFalcon:
Thanks for the review @OddBloke . In UAC, they don't install gcloud for their pycloudlib usage, so I thought I could cover multiple use cases this way, but it looks like file locations and keys don't stay consistent across versions. Since shelling out seems less likely to change than file/key locations, I'll go back to shelling out. If you don't have gcloud installed, the old way should still work.

pycloudlib/gce/cloud.py
```diff
@@ -50,6 +52,21 @@ def __init__(
 
         if project:
             os.environ["GOOGLE_CLOUD_PROJECT"] = str(project)
+        else:
```
TheRealFalcon, line 55: @lucasmoura Still not exactly sure how this env var is being used, so not sure if I should be setting it if we obtain the project id from the gcloud tool.

pycloudlib/gce/cloud.py
```diff
@@ -50,6 +52,21 @@ def __init__(
...
+            command = ['gcloud', 'config', 'get-value', 'project']
+            exception_text = (
+                "Could not obtain GCE project id. Has the CLI client been "
+                "setup?\nCommand attempted: '{}'".format(' '.join(command))
+            )
+            try:
+                result = subp(command)
+            except FileNotFoundError as e:
+                raise Exception(exception_text) from e
+            if not result.ok:
```
OddBloke, line 65: Hmm, looking at this again, can we actually take this `if` path? `not result.ok` implies a non-zero return code (of course), but we aren't passing `rcs` to `subp`, so it will default to `(0,)`: https://github.com/canonical/pycloudlib/blob/8a9a79792ead5b680dd558988732b3b533d60b0f/pycloudlib/util.py#L147 which I believe will in turn mean that we get a `RuntimeError` raised any time `result.ok` would be `False`: https://github.com/canonical/pycloudlib/blob/8a9a79792ead5b680dd558988732b3b533d60b0f/pycloudlib/util.py#L193-L201

TheRealFalcon: Yeah, that looks right. I tested the FileNotFoundException but couldn't think of a simple way to test the non-0 condition. The way we layer Result with subp isn't the most intuitive.

pycloudlib/gce/cloud.py
blackboxsw: @TheRealFalcon the GOOGLE_CLOUD_PROJECT env  is used if gcloud config wasn't set up on the jenkins worker node, so I think we still need both paths as some clean instances would not have those credentials setup locally.

lucasmoura: I think @TheRealFalcon has a point here. Looking at the docs, we look first for the `project-id` on the `GOOGLE_CLOUD_PROJECT` variable. If we do not find it, then we search for the `GOOGLE_CLOUD_CREDENTIALS`

https://googleapis.dev/python/google-cloud-core/latest/config.html

Because we must setup the `GOOGLE_CLOUD_CREDENTIALS`, I think we can skip setting this env variable

pycloudlib/gce/cloud.py
TheRealFalcon: Updated the `rcs` argument so resolving this

lucasmoura (approved):

OddBloke:
When running this in combination with https://github.com/canonical/cloud-init/pull/757, I see the following error when I run `CLOUD_INIT_PLATFORM=gce pytest tests/integration_tests/modules/test_write_files.py`:

```pytb
  File "/home/daniel/dev/cloud-init/tests/integration_tests/conftest.py", line 236, in class_client
    with _client(request, fixture_utils, session_cloud) as client:
  File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/daniel/dev/cloud-init/tests/integration_tests/conftest.py", line 210, in _client
    with session_cloud.launch(
  File "/home/daniel/dev/cloud-init/tests/integration_tests/clouds.py", line 165, in launch
    pycloudlib_instance = self._perform_launch(kwargs)
  File "/home/daniel/dev/cloud-init/tests/integration_tests/clouds.py", line 135, in _perform_launch
    pycloudlib_instance = self.cloud_instance.launch(**launch_kwargs)
  File "/home/daniel/dev/pycloudlib/pycloudlib/gce/cloud.py", line 206, in launch
    operation = self.compute.instances().insert(
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/googleapiclient/_helpers.py", line 130, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/googleapiclient/http.py", line 844, in execute
    resp, content = _retry_request(
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/googleapiclient/http.py", line 164, in _retry_request
    resp, content = http.request(uri, method, *args, **kwargs)
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/google_auth_httplib2.py", line 189, in request
    self.credentials.before_request(
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/google/auth/credentials.py", line 133, in before_request
    self.refresh(request)
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/google/oauth2/credentials.py", line 200, in refresh
    access_token, refresh_token, expiry, grant_response = _client.refresh_grant(
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/google/oauth2/_client.py", line 248, in refresh_grant
    response_data = _token_endpoint_request(request, token_uri, body)
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/google/oauth2/_client.py", line 124, in _token_endpoint_request
    _handle_error_response(response_body)
  File "/home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/google/oauth2/_client.py", line 60, in _handle_error_response
    raise exceptions.RefreshError(error_details, response_body)
google.auth.exceptions.RefreshError: ('invalid_grant: Bad Request', '{\n  "error": "invalid_grant",\n  "error_description": "Bad Request"\n}')
```

Earlier in the run, I see:

```
------ Captured log setup ------
2021-01-13 15:48:48 INFO      integration_testing:clouds.py:73 Detected image: image_id=focal os=ubuntu release=focal
2021-01-13 15:48:50 WARNING   google.auth._default:_default.py:346 No project ID could be determined. Consider running `gcloud config set project` or setting the GOOGLE_CLOUD_PROJECT environment variable
2021-01-13 15:48:50 INFO      integration_testing:clouds.py:73 Detected image: image_id=focal os=ubuntu release=focal
```

And in the warnings summary, I see:

```
  /home/daniel/.virtualenvs/cloud-init/lib/python3.8/site-packages/google/auth/_default.py:69: UserWarning: Your application has authenticated using end user credentials from Google Cloud SDK without a quota project. You might receive a "quota exceeded" or "API not enabled" error. We recommend you rerun `gcloud auth application-default login` and make sure a quota project is added. Or you can use service accounts instead. For more information about service accounts, see https://cloud.google.com/docs/authentication/
    warnings.warn(_CLOUD_SDK_CREDENTIALS_WARNING)
```

After running that command, not only do the tests run successfully, but I both have a `~/.config/gcloud/application_default_credentials.json` _and_ it has `quota_project_id` set.

So perhaps what we actually need to do is check for `quota_project_id` in there and if it's missing guide people to run `gcloud auth application-default login`? (Apologies for all the flipping back and forth between things here. :disappointed:)

TheRealFalcon:
@OddBloke I don't think we provide guidance for setting up any cloud CLIs, do we? We kind of leave that as an exercise for the user. Since you were given a warning telling you what you need to do, I'm thinking we don't need to do anything else atm. While I'm not providing giving guidance for setting up the various CLIs, I think that's beyond the scope of this particular PR.

Thoughts?

OddBloke (approved):
This LGTM, thanks!

PR #91, opened 2021-01-12
Fix regex that extracts release from image id

lucasmoura:
On LXD, we use a regex to extract the ubuntu release from a image id. We try to find either the base release
name on the image id, like focal, or the version number, like 20.4. However, if the image id has a timestamp on it we could accidentally match the timestamp value instead of a valid version string. The reason for that is because
we are not escaping the dot part of the version string.

For example, for this image id:
`local:ubuntu-behave-image-build--vm-focal1610458065934836-snapshot`

Using the following regex:
`(.*ubuntu.*(?P<release>(focal|bionic|xenial|20\.04|18\.04|16\.04)).*)`

We are matching the timestamp part `16104`


blackboxsw (approved):
LGTM

PR #94, opened 2021-01-20
Set service account email for gce instance

lucasmoura:
When we are launching an image, we do not bind it to a service account, even if we are using a service
account credential to launch the images. This is problematic because some of our vms need that to work properly. For example,
the recent gcp pro machines requires that bind to allow uaclient to auto-attach on them. To fix that, we are now
performing that bind if the user provides a credential file to the gce client and that credential file has an account_email on it

### DIFF
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -6,6 +6,7 @@ authentication into the cloud, finding an image, and launching an
 instance. It however, does not allow any further actions from occuring.
 """
 
+import json
 import logging
 import os
 import time
@@ -42,10 +43,11 @@ class GCE(BaseCloud):
         """
         super().__init__(tag, timestamp_suffix)
         self._log.debug('logging into GCE')
+        self.credentials_path = credentials_path
 
-        if credentials_path:
+        if self.credentials_path:
             os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = str(
-                credentials_path)
+                self.credentials_path)
 
         if project:
             os.environ["GOOGLE_CLOUD_PROJECT"] = str(project)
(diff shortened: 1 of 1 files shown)


mitechie:
I've got this on the agenda for a sync with google as well because needing to do a manual set/config to successfully use an image is a problem.

pycloudlib/gce/cloud.py
```diff
@@ -59,6 +61,17 @@ def __init__(
...
+
+    def _set_service_account_email(self):
+        """Set service account email if credentials provided."""
+        json_credentials = {}
+
+        if self.credentials_path:
+            with open(self.credentials_path, 'r') as f:
+                json_credentials = json.load(f)
+
+        self._service_account_email = json_credentials.get("client_email")
```
OddBloke, line 74: My `application_default_credentials.json` does not have a `client_email` key.

More broadly, I'm concerned about us relying on the credentials JSON: we've seen before that it doesn't have a predictable interface, so code written to use it is going to be unreliable.

OddBloke (commented):
I am not convinced that this functionality should be included in `pycloudlib`: this is handling a specific case in a specific way, which will not be necessary for most VM launches.

Put another way: should we be attaching a service account in this specific way to all GCE VMs we launch? My feeling is no: as it isn't required for "usual" VM functioning (which we know because we haven't needed it yet :p), then we shouldn't do it by default.

A couple of potential alternatives: you could expand the API of `GCE` to enable the setting of a service account and do that before `launch` is called, or you could [attach a service account to the VM once launched](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#changeserviceaccountandscopes).

Arguably, if a service account is required for Pro images, you should be testing any way that a service account could be presented to a GCE Pro instance: it should work if one is specified at launch as in this PR, but it should probably also work if one is attached 20 seconds after boot (or via an entirely different mechanism that we aren't yet aware of). You might also want a negative test: if a service account isn't presented to the instance, what happens? Hard-coding this behaviour into pycloudlib means that you can't readily test any behaviour but this specific one.

What do you think?

mitechie:
> You might also want a negative test: if a service account isn't presented to the instance, what happens? Hard-coding this behaviour into pycloudlib means that you can't readily test any behaviour but this specific one.

:+1:

lucasmoura:
@OddBloke I think those are all good points. I will close this PR and work on attaching the service account directly on uaclient. This will also allows us to test the negative scenario as you mentioned. Thanks for the review :)

Issue #95, opened 2021-01-20
Cannot launch hirsute LXD VM instances


OddBloke:
When I attempt to do so, I see the following traceback:

```pytb
Traceback (most recent call last):
  File "/home/daniel/scratch/cloud-init/tests/integration_tests/clouds.py", line 285, in _perform_launch
    profile_list = launch_kwargs['profile_list']
KeyError: 'profile_list'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/daniel/scratch/cloud-init/tests/integration_tests/conftest.py", line 142, in setup_image
    client = session_cloud.launch()
  File "/home/daniel/scratch/cloud-init/tests/integration_tests/clouds.py", line 165, in launch
    pycloudlib_instance = self._perform_launch(kwargs)
  File "/home/daniel/scratch/cloud-init/tests/integration_tests/clouds.py", line 287, in _perform_launch
    profile_list = self._get_or_set_profile_list(release)
  File "/home/daniel/scratch/cloud-init/tests/integration_tests/clouds.py", line 316, in _get_or_set_profile_list
    self._profile_list = self.cloud_instance.build_necessary_profiles(
  File "/home/daniel/dev/pycloudlib/pycloudlib/lxd/cloud.py", line 608, in build_necessary_profiles
    base_release = self._extract_release_from_image_id(release)
  File "/home/daniel/dev/pycloudlib/pycloudlib/lxd/cloud.py", line 596, in _extract_release_from_image_id
    return self._image_info(image_id)[0]["release"]
IndexError: list index out of range
```

OddBloke:
This issue does not present on LXD container launches.

PR #99, opened 2021-01-21
lxd: use `lxc image info` for release name determination

OddBloke:
Both images: and ubuntu[-daily]: Ubuntu LXD images are published with metadata that describes the OS and release of the contained image.  This metadata is retained across image snapshots.

Instead of relying on naming conventions and regexes to determine the release of a LXD image, this PR calls `lxc image info` and parses the emitted YAML to find the apropos release.

This also includes a couple of prerequisite changes: there was a naming inconsistency that made it difficult to reason about this issue, and we were handling image IDs inconsistently (some public APIs would prepend the daily remote if no remote was specified, others wouldn't).

Fixes: #95

### DIFF
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -138,15 +138,15 @@ class _BaseLXD(BaseCloud):
 
     # pylint: disable=R0914,R0912,R0915
     def _prepare_command(
-            self, name, release, ephemeral=False, network=None, storage=None,
+            self, name, image_id, ephemeral=False, network=None, storage=None,
             inst_type=None, profile_list=None, user_data=None,
             config_dict=None):
         """Build a the command to be used to launch the LXD instance.
 
         Args:
             name: string, what to call the instance
-            release: string, [<remote>:]<release>, what release to launch
-                     (default remote: )
+            image_id: string, [<remote>:]<image identifier>, the image to
+                      launch
             ephemeral: boolean, ephemeral, otherwise persistent
             network: string, optional, network name to use
             storage: string, optional, storage name to use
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
Nice refactor

lucasmoura (commented):
I am just running some uaclient tests to assure the refactor will not affect anything, but that refactoring looks great. Thanks for that @OddBloke

pycloudlib/lxd/cloud.py
```diff
@@ -567,45 +572,68 @@ def _image_info(self, image_id, image_hash_key=None):
...
+        Returns:
+            A dict produced by loading the YAML emitted by ``lxc image info
+            <image_id>``, or the empty dict if either the command or YAML load
+            fails.
+        """
+        raw_image_info = subp(["lxc", "image", "info", image_id], rcs=())
+        if raw_image_info.ok:
+            try:
+                return yaml.safe_load(raw_image_info)
+            except yaml.error.YAMLError:
```
TheRealFalcon, line 591: Is this right? My editor says this is an error and I think it's just `yaml.YAMLError` rather than `yaml.error.YAMLError`

lucasmoura (approved):
Everything is working as expected in `uaclient`

OddBloke:
On Fri, Jan 22, 2021 at 11:10:22AM -0800, James Falcon wrote:
> Is this right? My editor says this is an error and I think it's just
> `yaml.YAMLError` rather than `yaml.error.YAMLError`

It is _valid_:

```py
In [1]: import yaml

In [2]: yaml.error.YAMLError
Out[2]: yaml.error.YAMLError

In [3]: yaml.YAMLError
Out[3]: yaml.error.YAMLError
```

but that's not the same as right; let me fix.

PR #100, opened 2021-01-21
Set service account email for gce instance

lucasmoura:
When we are launching an image, we do not bind it to a service account, even if we are using a service account credential to launch the images. This is problematic because some of our vms need that to work properly. For example, the recent gcp pro machines requires that bind to allow uaclient to auto-attach on them. To fix that, we are now performing that bind if the user provides a service_account_email to the gce client.

@OddBloke I think this is a less restrictive approach than the one I had before

### DIFF
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -27,7 +27,7 @@ class GCE(BaseCloud):
 
     def __init__(
         self, tag, timestamp_suffix=True, credentials_path=None, project=None,
-        region="us-west2", zone="a"
+        region="us-west2", zone="a", service_account_email=None
     ):
         """Initialize the connection to GCE.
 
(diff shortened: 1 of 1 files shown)


OddBloke (approved):
Thanks Lucas, this seems like a good solution. I've confirmed (with a local modification to cloud-init's test framework to specify a service account) that this does launch instances which can fetch service account tokens:

```sh
$ curl "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token" -H "Metadata-Flavor: Google"
{"access_token":"<snip>","expires_in":3599,"token_type":"Bearer"}
```

OddBloke:
FWIW, I just realised that I had a local modification to pycloudlib (to set scopes on the service account, which I initially thought was needed to fix problems I was seeing which were, in fact, PEBKAC); if I remove that, I see:

```sh
$ curl "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token" -H "Metadata-Flavor: Google"
"No service account scopes specified."
```

No change required if this is expected, just wanted to flag up that my testing wasn't exactly as described.

lucasmoura:
@OddBloke I think this expected since we are not passing any `scope` when performing the service account binding. But I do think we can solve this if another optional parameter, if necessary

Issue #106, opened 2021-01-27
oci 2.18.0 breaks pycloudlib


OddBloke:
Specifically, from https://github.com/oracle/oci-python-sdk/blob/master/CHANGELOG.rst#breaking-11:

> * Parameter vcn_id changed from required to optional in methods list_dhcp_options, list_local_peering_gateways, list_route_tables, list_security_lists, list_subnets and list_internet_gateways in the virtual network client. If the VCN ID is not provided, then the list includes information of all VCNs in the specified compartment.

which means that we can no longer pass `vcn_id` positionally to `list_subnets`.

PR #107, opened 2021-01-27
align GCE and Oracle launch wait behaviour with other clouds

OddBloke:
GCE simply had a bug: it didn't call `instance.wait()`.

OCI's behaviour is changed somewhat: we now unconditionally wait for the instance to be `RUNNING` and `wait=True` controls whether `instance.wait()` is called.

### DIFF
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -233,7 +233,11 @@ class GCE(BaseCloud):
             result['networkInterfaces'][0]['accessConfigs'][0]['natIP']
         )
 
-        return self.get_instance(result['id'], name=result['name'])
+        instance = self.get_instance(result['id'], name=result['name'])
+        if wait:
+            instance.wait()
+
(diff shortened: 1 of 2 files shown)


OddBloke:
I've tested this with the cloud-init integration testing framework locally against both GCE and OCI: it fixed the GCE failures found in security fix validation, and those same tests ran successfully on OCI (with a manual workaround for #106 applied).

TheRealFalcon (approved):
+1 either way, but is there a reason we wait unconditionally for instance to be running, rather than  wait for both things?

OddBloke:
> +1 either way, but is there a reason we wait unconditionally for instance to be running, rather than wait for both things?

I _thought_ that was standard, but that might just be GCE and Oracle now. I'll land this and double-check.

OddBloke:
> I _thought_ that was standard, but that might just be GCE and Oracle now. I'll land this and double-check.

#108

PR #108, opened 2021-01-28
oci/gce: only perform cloud-native wait in launch if wait=True

OddBloke:
I previously made OCI's wait=True behaviour consistent with GCE, but
that means they are inconsistent with the other clouds supported by
pycloudlib.  This commit aligns their behaviour with the other clouds.

Specifically, the "wait for instance start" calls made to the cloud are
no longer unconditional, and instead are only executed if wait=True
before instance.wait() is called.

### DIFF
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -220,8 +220,6 @@ class GCE(BaseCloud):
         ).execute()
         raise_on_error(operation)
 
-        self._wait_for_operation(operation, operation_type='zone')
-
         result = self.compute.instances().get(
             project=self.project,
             zone=self.zone,
@@ -229,12 +227,9 @@ class GCE(BaseCloud):
         ).execute()
         raise_on_error(result)
 
-        self._log.info(
-            result['networkInterfaces'][0]['accessConfigs'][0]['natIP']
-        )
-
         instance = self.get_instance(result['id'], name=result['name'])
         if wait:
+            self._wait_for_operation(operation, operation_type='zone')
             instance.wait()
 
         return instance
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -204,13 +204,13 @@ class OCI(BaseCloud):
 
         instance_data = self.compute_client.launch_instance(
             instance_details).data
-        instance_data = wait_till_ready(
-            func=self.compute_client.get_instance,
-            current_data=instance_data,
-            desired_state='RUNNING',
-        )
         instance = self.get_instance(instance_data.id)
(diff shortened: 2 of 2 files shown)


OddBloke:
I've tested this and it works locally with cloud-init's integration testing framework for both GCE and OCI.

pycloudlib/gce/cloud.py
```diff
@@ -220,21 +220,16 @@ def launch(self, image_id, instance_type='n1-standard-1', user_data=None,
         ).execute()
         raise_on_error(operation)
 
-        self._wait_for_operation(operation, operation_type='zone')
```
TheRealFalcon, line 223: The call before this creates the instance, and the call after uses the information created to create our own instance object. If the first isn't finished, will the following run correctly?

pycloudlib/oci/cloud.py
```diff
@@ -204,13 +204,13 @@ def launch(self, image_id, instance_type='VM.Standard2.1', user_data=None,
 
         instance_data = self.compute_client.launch_instance(
             instance_details).data
-        instance_data = wait_till_ready(
-            func=self.compute_client.get_instance,
-            current_data=instance_data,
-            desired_state='RUNNING',
-        )
         instance = self.get_instance(instance_data.id)
```
TheRealFalcon, line 207: If we haven't waiting for the launch call, will an `id` be available if the launch hasn't completed?

TheRealFalcon (commented):
So looking at this again, I might take back what I said in the previous PR :grimacing: 

I think we should respect wait when we can, but it looks like in both of these cases, removing the wait could potentially make us attempt to use something that isn't fully initialized yet.

pycloudlib/gce/cloud.py
OddBloke: I believe so. Anecdotally, I've seen this succeed. More broadly, I think the lifecycle of the VM for an instance is decoupled from the lifecycle of the resource (in the RESTful-API sense) in GCE: https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert says it "Creates an instance resource in the specified project using the data included in the request." I would expect the _resource_ creation to be complete when this call returns, including the ID assignment.

(Separately, the instance resource has multiple states in which it can be: https://cloud.google.com/compute/docs/instances/instance-life-cycle; these are represented within the resource.)

TheRealFalcon (approved):
Ok, great. Thanks for the clarification and for this PR.

OddBloke:
> Ok, great. Thanks for the clarification and for this PR.

Thanks for the review, and for prompting me to check this more than anecdotally!

PR #109, opened 2021-01-28
oci: fix API call broken in oci 2.18.0

OddBloke:
Per their changelog[0], list_subnets went from taking `vcn_id`
positionally to taking it as a keyword argument: this commit follows
that.

[0] https://github.com/oracle/oci-python-sdk/blob/master/CHANGELOG.rst#breaking-11

Fixes: #106


OddBloke:
I've tested this with the cloud-init integration tests.

PR #113, opened 2021-01-30
Initial openstack support

TheRealFalcon:
At this point I think the basic functionality is usable and reviewable.

Everything in the base_api.py test works, so the basics all work. I intentionally didn't add an example since all clouds should share the same base API right now. We should really only need one example unless we want to showcase specialized use cases of a specific cloud.

### DIFF
--- a/docs/clouds/openstack.md
+++ b/docs/clouds/openstack.md
@@ -1,0 +1,21 @@
+# Openstack
+
+## Credentials
+
+No connection information is directly passed to pycloudlib but rather relies **clouds.yaml** or **OS_** environment variables. See [the openstack configuration docs](https://docs.openstack.org/python-openstackclient/victoria/configuration/index.html) for more information.
+
+## SSH Keys
+
+Openstack can't launch instances unless an openstack managed keypair already exists. Since pycloudlib also manages keys, pycloudlib will attempt to use or create an openstack ssh keypair based on the pycloudlib keypair. If a key is provided to pycloudlib with the same name and public key that already exists in openstack, that key will be used. If no key information is provided, an openstack keypair will be created with the current user's username and public key.
+
+## Image ID
+
+The image id to use for a launch must be manually passed to pycloudlib rather than determined from release name. Given that each openstack deployment can have a different setup of images, it's not practical given the information we have to guess which image to use for any particular launch.
+
+## Network ID
+
+Network ID must be manually passed to pycloudlib. Since there can be multiple networks and no concept of a default network, we can't choose which network to create an instance on.
+
+## Floating IPs
+
+A floating IP is allocated and used per instance created. The IP is then deleted when the instance is deleted.
--- a/docs/clouds/openstack.md
+++ b/docs/clouds/openstack.md
@@ -2,7 +2,7 @@
 
 ## Credentials
 
-No connection information is directly passed to pycloudlib but rather relies **clouds.yaml** or **OS_** environment variables. See [the openstack configuration docs](https://docs.openstack.org/python-openstackclient/victoria/configuration/index.html) for more information.
+No connection information is directly passed to pycloudlib but rather relies on **clouds.yaml** or **OS_** environment variables. See [the openstack configuration docs](https://docs.openstack.org/python-openstackclient/victoria/configuration/index.html) for more information.
 
 ## SSH Keys
 
--- a/pycloudlib/openstack/cloud.py
+++ b/pycloudlib/openstack/cloud.py
@@ -1,0 +1,224 @@
+"""Openstack cloud type."""
+import base64
+from operator import itemgetter
+
+import openstack
+
+from pycloudlib.cloud import BaseCloud
+from pycloudlib.openstack.instance import OpenstackInstance
+
+
+class Openstack(BaseCloud):
+    """Openstack cloud class."""
+
+    _type = 'openstack'
+
+    def __init__(self, tag, network, timestamp_suffix=True):
+        """Initialize the connection to openstack.
+
+        Requires valid pre-configured environment variables or clouds.yaml.
+        See https://docs.openstack.org/python-openstackclient/pike/configuration/index.html
+
+        Args:
+            tag: Name of instance
+            network: Name of the network to use (from openstack network list)
+            timestamp_suffix: bool set True to append a timestamp suffix to the
+                tag
+        """  # noqa: E501
+        super().__init__(tag, timestamp_suffix)
+        self.network = network
+        self._openstack_keypair = None
+        self.conn = openstack.connect()
+
+    def delete_image(self, image_id):
+        """Delete an image.
+
+        Args:
+            image_id: string, id of the image to delete
+        """
+        self.conn.delete_image(image_id, wait=True)
+
+    def released_image(self, release, **kwargs):
+        """Find the id of the latest released image for a particular release.
+
+        Args:
+            release: string, Ubuntu release to look for
+            arch: string, architecture to use
+            root_store: string, root store to use
+
+        Returns:
+            string, id of latest image
+
+        """
+        self._log.warning('Released images NOT supported for openstack')
+        return self.daily_image(release)
+
+    def daily_image(self, release, **kwargs):
+        """Find the id of the latest daily image for a particular release.
+
+        Args:
+            release: string, Ubuntu release to look for
+            arch: string, architecture to use
+            root_store: string, root store to use
+
+        Returns:
+            string, id of latest image
+
+        """
+        self._log.debug('finding daily Ubuntu image for %s', release)
+        images = self.conn.compute.images(details=True)
+        images = [image for image in self.conn.compute.images()
+                  if release in image['name']]
+        sorted_images = sorted(images, key=itemgetter('updated_at'))
+        if not sorted_images:
+            raise ValueError('No image found for release {}'.format(release))
+        return sorted_images[-1]['id']
+
+    def image_serial(self, image_id):
+        """Find the image serial of the latest daily image for a particular release.
+
+        Args:
+            image_id: string, Ubuntu image id
+
+        Returns:
+            string, serial of latest image
+
+        """
+        raise NotImplementedError
+
+    def get_instance(self, instance_id) -> OpenstackInstance:
+        """Get an instance by id.
+
+        Args:
+            instance_id:
+
+        Returns:
+            An instance object to use to manipulate the instance further.
+
+        """
+        return OpenstackInstance(
+            self.key_pair,
+            instance_id,
+        )
+
+    def launch(self, image_id, instance_type='m1.small', user_data=None,
+               wait=True, **kwargs) -> OpenstackInstance:
+        """Launch an instance.
+
+        Args:
+            image_id: string, image ID to use for the instance
+            instance_type: string, type (flavor) of instance to create
+            user_data: used by cloud-init to run custom scripts/configuration
+            wait: wait for instance to be live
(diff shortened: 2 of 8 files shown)


pycloudlib/openstack/cloud.py
```diff
@@ -0,0 +1,159 @@
...
+from pycloudlib.cloud import BaseCloud
+from pycloudlib.openstack.instance import OpenstackInstance
+
+
+class Openstack(BaseCloud):
+    """ Openstack cloud class."""
+
+    _type = 'openstack'
+
+    def __init__(self, tag, network, timestamp_suffix=True):
```
TheRealFalcon, line 16: I think we have no choice but to require a network parameter. Pycloudlib can't auto-determine the network to use because there could be many defined for any particular openstack deployment and randomly choosing one seems like a bad idea.

pycloudlib/openstack/cloud.py
```diff
@@ -0,0 +1,159 @@
...
+            root_store: string, root store to use
+
+        Returns:
+            string, id of latest image
+
+        """
+        self._log.warning('Released images NOT supported for openstack')
+        return self.daily_image(release)
+
+    def daily_image(self, release, **kwargs):
```
TheRealFalcon, line 40: I initially tried to use simplestreams here, but realized that many openstack deployments (including ours) are probably out of date with the latest images.

pycloudlib/openstack/cloud.py
```diff
@@ -0,0 +1,159 @@
...
+                "but its public key doesn't match the public key passed "
+                "in.\n"
+                "{name} key: {openstack_key}\n"
+                "Passed in key: {passed_key}".format(
+                    name=name,
+                    openstack_key=openstack_keypair.public_key,
+                    passed_key=public_key_content
+                )
+            )
+        # TODO: Figure out if I want to incorpoate this...
```
TheRealFalcon, line 140: If anybody has strong opinions about this, please chime in.

TheRealFalcon:
I recently noticed that on the CLI, I don't need to provide the network information like I do through the API. I'm curious how the CLI knows which network to create the instance on and if we could use that same knowledge to acquire the network to use when using the API.

TheRealFalcon:
I'm also going on the assumption that we want a floating IP associated with instances that we launch. Obviously not every openstack deployment in the wild will require that, but I don't think that's something we need to worry about yet.

pycloudlib/openstack/cloud.py
OddBloke: Yeah, that's definitely an issue we'll hit with OpenStack deployments. It's also entirely possible, of course, that there are no Ubuntu images available in any given deployment.

I'm concerned that this heuristic could cause some confusing test behaviours, both between platforms (i.e. a current EC2 image vs a stale OpenStack image) and between OpenStack deployments (most likely one which is running glance-simplestreams-sync vs one which is not). We could also very easily end up launching non-base non-official images without noticing.

I wonder if instead of trying to determine this automatically, we should accept that we can't reliably launch the images that pycloudlib usually does and require OpenStack consumers to provide an explicit image ID in all cases? My thinking is that the image ID to use is deployment-specific, so is best determined by the consumers who know what OpenStack deployment they're running against.

(IIRC, what we did in cloud-test-framework is have an OpenStack cloud class which holds the cross-deployment behaviour and logic, and then subclasses of that which define the deployment-specific behaviour: ScalingStack and then the OpenStack public clouds we supported. Someone trying to use "generic OpenStack" would have to fill in the unknowns to proceed, but we still had a shared understanding of how to operate against shared deployments stored in the codebase.)

A couple of other thoughts: I believe that glance-simplestreams-sync does include additional metadata against the images, which we might be able to use to make this heuristic less likely to hit false positives (i.e. less likely to return non-official Ubuntu images); we might even be able to use it to cross-reference streams (i.e. gather a list of all fingerprints that are valid for $series according to streams and then find the most recent corresponding image). We could also consider using streams to populate the target deployment with an image ourselves: that's probably the only way we can be sure that we'll have an official Ubuntu image on every OpenStack deployment we encounter.

docs/clouds/openstack.md
```diff
@@ -0,0 +1,21 @@
+# Openstack
```
TheRealFalcon, line 1: Added some documentation here, but not sure how I'm supposed to generate it. I tried the makefile in the docs directory, but the output looks very not right.

pycloudlib/openstack/tests/test_cloud.py
```diff
@@ -0,0 +1,40 @@
...
+import pytest
+from openstack.connection import Connection
+
+from pycloudlib.openstack.cloud import Openstack
+
+
+@mock.patch('pycloudlib.key.KeyPair.public_key_content',
+            new_callable=mock.PropertyMock,
+            return_value='pretend public key')
+@mock.patch('openstack.connect', spec=Connection)
```
OddBloke, line 13: I'm not sure this `spec` is doing what you want here:

```py
In [1]: import unittest.mock as mock
In [3]:  from openstack.connection import Connection
In [4]: p = mock.patch("openstack.connect", spec=Connection)
In [5]: m = p.start()
In [6]: m()  # this is what `openstack.connect()` will return
Out[6]: <MagicMock name='connect()' id='140181990886512'>
In [7]: m  # and I think you actually want it to return this
Out[7]: <openstack.connection.Connection at 0x7f7ea9c62fa0>
```

pycloudlib/openstack/tests/test_cloud.py
```diff
@@ -0,0 +1,40 @@
...
+class TestOpenstackKeypair:
+    """Tests covering _get_openstack_keypair."""
+
+    def test_keypair_doesnt_exist(self, m_openstack, m_keypair):
```
OddBloke, line 17:
```suggestion
    def test_keypair_doesnt_exist(self, m_openstack, _m_keypair):
```

pycloudlib/openstack/tests/test_cloud.py
```diff
@@ -0,0 +1,40 @@
...
+        """Test no pre-existing openstack keypair."""
+        m_openstack.return_value.get_keypair.return_value = None
+        cloud = Openstack(tag='test', network=None)
+        cloud._get_openstack_keypair()
+        assert 1 == cloud.conn.create_keypair.call_count
+
+    def test_keypairs_match(self, m_openstack, m_keypair):
```
OddBloke, line 24: `m_keypair` threw me off here: I assumed that would be a mock of `pycloudlib.key.KeyPair` (making this test very confusing :p). Perhaps `m_public_key_content`? Still a little confusing, given it's a property mock (whereas you'd never call `public_key_content` of course).

OddBloke (commented):
Thanks for this work, James! By and large, this looks good. I have some thoughts inline.

pycloudlib/openstack/cloud.py
```diff
@@ -0,0 +1,189 @@
...
+            self._openstack_keypair = self._get_openstack_keypair()
+        if user_data:
+            user_data = base64.b64encode(user_data.encode()).decode()
+        else:
+            user_data = ''
+
+        instance = self.conn.compute.create_server(
+            name=self.tag,
+            image_id=image_id,
+            flavor_id=self.conn.compute.find_flavor(instance_type).id,
```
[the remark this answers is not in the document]
TheRealFalcon: Ah, I didn't realize that about flavors. I'll update it accordingly.

pycloudlib/openstack/instance.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @property
+    def ip(self):
+        """Return IP address of instance."""
+        return self.floating_ip.floating_ip_address
+
+    def console_log(self):
+        """Return the instance console log."""
+        # Returning nothing for me but might work in other contexts?
+        self.conn.compute.get_server_console_output(self.server)
```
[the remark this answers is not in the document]
TheRealFalcon: :facepalm: I guess that explains why it was returning nothing!

TheRealFalcon:
Addressed the code comments. Still need to address the test comments. @OddBloke , I addressed the floating IP issue a little differently than we initially discussed, so you might want to take a look.

TheRealFalcon:
Updated the test contents, but now I have new floating ip tests to write :wink:

TheRealFalcon:
Added new tests. Should be ready for re-review.

pycloudlib/openstack/instance.py
```diff
@@ -0,0 +1,151 @@
...
+        if not connection:
+            connection = openstack.connect()
+        self.conn = connection
+
+        self.server = self.conn.compute.get_server(instance_id)
+
+        self.delete_floating_ip = False
+        self.floating_ip = self._get_existing_floating_ip()
+        if self.floating_ip is None:
+            self._create_and_attach_floating_id()
```
OddBloke, line 40: A nit, but `self.floating_ip = self._create_and_attach_floating_id()` feels a little easier to follow; it would give `_get_...` and `_create_...` a (more) similar API and means that `self.floating_ip` is only set in one place.

pycloudlib/openstack/tests/test_instance.py
```diff
@@ -0,0 +1,66 @@
...
+        }
+    ]
+}
+
+# While the actual results from openstack will have a lot more values
+# in the dict, 'floating_ip_address' is all we care about
+NETWORK_IPS = [
+    {'floating_ip_address': '10.0.0.4'},
+    {'floating_ip_address': '10.0.0.3'},
+]
```
OddBloke, line 36: Another nit, but if the code under test incorrectly did the equivalent of `chain(*[d.values() for d in NETWORK_IPS])`, we wouldn't catch that with this input.

pycloudlib/openstack/tests/test_instance.py
```diff
@@ -0,0 +1,66 @@
...
+    'pycloudlib.openstack.instance.OpenstackInstance.'
+    '_create_and_attach_floating_id'
+)
+class TestAttachFloatingIp:
+    """Ensure we create/use floating IPs accordingly."""
+
+    def test_existing_floating_ip(self, m_create):
+        """Test that if a server has an existing floating IP, we use it."""
+        m_connection = mock.Mock()
+        m_server = m_connection.compute.get_server.return_value = mock.Mock()
```
OddBloke, line 49: A nit; I don't think we need this final assignment:

```suggestion
        m_server = m_connection.compute.get_server.return_value
```

```py
In [6]: from unittest import mock                                                                                                                                                                                                                                                         

In [7]: m_connection = mock.Mock()                                                                                                                                                                                                                                                        

In [8]: m_server = m_connection.compute.get_server.return_value                                                                                                                                                                                                                           

In [9]: m_server                                                                                                                                                                                                                                                                          
Out[9]: <Mock name='mock.compute.get_server()' id='140123963984144'>

In [10]: m_server = m_connection.compute.get_server.return_value = mock.Mock()                                                                                                                                                                                                            

In [11]: m_server                                                                                                                                                                                                                                                                         
Out[11]: <Mock name='mock.compute.get_server()' id='140123963995232'>
```

OddBloke (commented):
Floating IP stuff LGTM, I've left a few inline comments, all of which are nits. I'm going to run some tests locally before approving but, assuming that's all good, this will be good to land.

docs/clouds/openstack.md
OddBloke: https://github.com/canonical/pycloudlib/pull/125

OddBloke (approved):
Thanks for all the work!

Issue #112, opened 2021-01-28
LXD connections are attempted via SSH iff a non-default `key_pair` is specified


OddBloke:
Due to this code: https://github.com/canonical/pycloudlib/blob/1ac9d4c82fdfd5cb1407f70b8a2b17e02953569d/pycloudlib/lxd/instance.py#L31-L39

This also interacts badly with LXD's SSH key defaults. For most clouds, the `key_pair` used by default is https://github.com/canonical/pycloudlib/blob/1ac9d4c82fdfd5cb1407f70b8a2b17e02953569d/pycloudlib/cloud.py#L30-L32 LXD, however, has a different default: https://github.com/canonical/pycloudlib/blob/1ac9d4c82fdfd5cb1407f70b8a2b17e02953569d/pycloudlib/lxd/cloud.py#L49-L50 This means that in the default case, LXD will `exec`. In itself, this is not a problem.

However, it also means that if a consumer chooses to specify a non-default key (via `.use_key`), LXD's behaviour _changes_. `self.key_pair` is no longer `None`, and so LXD will start using SSH to access instances.

This is biting us in the cloud-init integration testing: we have a configuration option to specify an SSH key (and we call `use_key` if provided). If it isn't set, then we want to use pycloudlib's default (i.e. we don't call `use_key` at all). This means that users who do specify the configuration option see test failures on LXD: we rely on the fact that LXD does not require functioning networking to access the instance for some tests, and so using SSH (as pycloudlib does for such users) is guaranteed to fail.

OddBloke:
It's clear to me that we should decouple LXD's choice of execution method from the presence of an SSH key: what we have currently is action at a distance, and took me a while to figure out.

It's less clear to me what the correct way of selecting execution method is. I don't think it belongs on `execute`: that's part of the super-class's API, so introducing a LXD-specific flag for it isn't ideal (and, indeed, ISTR we've tried that before).

Given that a parameter isn't an option, the two choices that I can think of are (a) a flag on the `LXDInstance` instance (`if self.use_ssh:`) which is documented as something consumers can toggle; or (b) have `_run_command` only handle one of the paths, and have a separate callable on the `LXDInstance` class (something like `instance.run_ssh_command(...)` or `instance.exec_command(...)`, depending on what we choose to be the default behaviour) which handles the other path.

(If we choose (b) we likely should default to SSH: that would let us drop LXD's `_run_command` implementation altogether, and make `execute`s behaviour consistent across clouds.)

[...]

PR #114, opened 2021-02-05
lxd/instance: use SSH by default, allow switching to `exec`

OddBloke:
Analysis performed (and captured in #112) indicates that accessing LXD
instances via SSH rather than via `lxc exec` leads to faster test
execution, and also brings LXD's default behaviour in line with the
default behaviour of other clouds.

This commit introduces a `execute_via_ssh` parameter and instance
variable, which are used by `execute()` to determine whether to connect
to the instance via SSH, or use `lxc exec`.  The default value of
`execute_via_ssh` is `True`.

(Prior to this commit, we would execute via SSH but only if we had a key
manually set against the instance, leading to inconsistencies in
behaviour depending on local configuration.  This also addresses that.)

Fixes: #112

### DIFF
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -37,18 +37,6 @@ class _BaseLXD(BaseCloud):
     _daily_remote = 'ubuntu-daily'
     _releases_remote = 'ubuntu'
 
-    def __init__(self, tag, timestamp_suffix=True):
-        """Initialize LXD cloud class.
-
-        Args:
-            tag: string used to name and tag resources with
-            timestamp_suffic: Append a timestamped suffix to the tag string.
-        """
-        super().__init__(tag, timestamp_suffix)
-
-        # User must manually specify the key pair to be used
-        self.key_pair = None
-
     def clone(self, base, new_instance_name):
         """Create copy of an existing instance or snapshot.
 
@@ -121,21 +109,7 @@ class _BaseLXD(BaseCloud):
             The existing instance as a LXD instance object
 
         """
-        instance = LXDInstance(instance_id)
-
-        if self.key_pair:
-            local_path = "/tmp/{}-authorized-keys".format(instance_id)
-
-            instance.pull_file(
-                remote_path="/home/ubuntu/.ssh/authorized_keys",
-                local_path=local_path
-            )
-
-            with open(local_path, "r") as f:
-                if self.key_pair.public_key_content in f.read():
-                    instance.key_pair = self.key_pair
-
-        return instance
+        return LXDInstance(instance_id, key_pair=self.key_pair)
 
     def _normalize_image_id(self, image_id: str) -> str:
         if ':' not in image_id:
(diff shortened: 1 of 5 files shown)


OddBloke:
The corresponding cloud-init changes are at https://github.com/canonical/cloud-init/pull/802

OddBloke:
Looks like I need some test changes; I think the code is good to review while I look at them.

OddBloke:
Ah, and it also looks like I've introduced an error condition in the "default SSH key" codepath (which I don't take on my local machine), so I'll also need to address that.

pycloudlib/lxd/cloud.py
```diff
@@ -121,21 +109,7 @@ def get_instance(self, instance_id):
...
-
-        if self.key_pair:
-            local_path = "/tmp/{}-authorized-keys".format(instance_id)
-
-            instance.pull_file(
-                remote_path="/home/ubuntu/.ssh/authorized_keys",
-                local_path=local_path
-            )
-
-            with open(local_path, "r") as f:
```
lucasmoura, line 134: I know that this code is not optimal at all, but the reason I have added it here is that we could change the ssh keys here, but still have launched instances that are using an old key. In that situation, we would provided an unusable instance.

However, I know that this is not optimal because, for example, if we can't exec into the machine, that `pull_file` call will not work. So maybe we can drop this check and live with this possibility.

But if you think there is a smarter way to handle that, no problem at all

OddBloke: Thanks for the comment, this is the code I was hoping to discuss! You're correct, the issue I found is that we don't necessarily have `exec` access when this runs (and, indeed, we definitely don't when this runs in the demo scripts, which was causing CI failure). Adding an `instance._wait_for_execute()` call in `get_instance` doesn't seem like the right thing to do: if the instance isn't expected to be accessible, that would cause a long delay.

I think this logic is less necessary than it was previously, however, because `instance.key_pair` no longer determines our execute behaviour: even if `instance.key_pair` is `None`, we will still attempt to SSH in (using any keys available to paramiko).

However, I don't think we can say it's entirely unnecessary: while we don't need it to determine the SSH key to use (and, actually, never did because paramiko will try all keys by default), it _did_ serve a purpose previously: if SSH execution _could_ work (determined via `exec`), the LXDInstance would be configured to execute via SSH instead of via `exec`, otherwise it would be configured to use `exec`. We lose that behaviour here: we will always return a `LXDInstance` configured to execute via SSH. 

cloud-init's tests only have a single place which uses `get_instance` (when `CLOUD_INIT_EXISTING_INSTANCE_ID` is set). It actually has a similar issue to the one we're trying to solve here: previously with a local SSH key configured, it would have always configured SSH access (even to instances for which SSH wouldn't work: the existence of an authorized key does not indicate, e.g., working networking), and would otherwise have always configured `exec` access. I need to update https://github.com/canonical/cloud-init/pull/802 to handle things differently: we should set `instance.execute_via_ssh` after the instance has been instantiated. We would still not require the "execution method auto-detection" which is being removed here.

Does UA client have test code which would be affected by this removal? Or can we safely remove it (and modify the docstring here to explain the execution behaviour)?

lucasmoura: I did not known about the fact that `paramiko`  would try to access the machine with all of the keys that is has seen. If that is the case, I don't see any problem removing the code and it should not affect uaclient at all.

Thanks for all the provided context @OddBloke :)

lucasmoura (approved):

TheRealFalcon (approved):

PR #115, opened 2021-02-05
pycloudlib: increase a variety of retry frequencies

### LINKED ISSUE #111: SSH connection retrying can take significantly longer than the apparent 10 minutes
[`_ssh_connect` sleeps for 10 seconds between failures, up to 60 times](https://github.com/canonical/pycloudlib/blob/master/pycloudlib/instance.py#L317). However, depending on the failure mode, `client.connect` can take 10s of seconds to return (e.g. when a connection is timing out, rather than being immediately refused). As a result, instead of waiting for `60*(10 + miniscule function calling overhead)` seconds, it's possible to wait for several multiples of that amount of time to see a full failure.

(This could be avoided by retrying only if less than 10 minutes of wall clock time have passed.)

OddBloke:
When investigating the relative performance of execution methods for #114, I found that a lot of our slow launch speeds were due to long sleeps between retries. This PR tweaks a variety of them, and fixes #111 as well.

### DIFF
--- a/pycloudlib/instance.py
+++ b/pycloudlib/instance.py
@@ -375,7 +375,7 @@ class BaseInstance(ABC):
         test_instance_command = "whoami"
         result = self.execute(test_instance_command)
         if result.failed:
-            retries = 10
+            retries = 100
             while retries:
                 result = self.execute(test_instance_command)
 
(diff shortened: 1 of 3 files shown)


TheRealFalcon (approved):
Thanks!

PR #117, opened 2021-02-11
Update lxc-setup script on lxd vm instances

lucasmoura:
When we create Xenial and Bionic lxd vm instances through pycloudlib, we need to provide a script to install the lxd-agent
on those machines. Recently, one of the services that was being started in this script was dropped from lxd, lxd-agent-9p. We are
now updating the script to remove references to that service.

This is the new script being used:
```sh
#!/bin/sh
if ! grep lxd_config /proc/mounts; then
    mkdir -p /run/lxdagent
    mount -t 9p config /run/lxdagent
    VIRT=$(systemd-detect-virt)
    case $VIRT in
        qemu|kvm)
            (cd /run/lxdagent/ && ./install.sh)
            umount /run/lxdagent
            systemctl start lxd-agent
            ;;
        *)
    esac
fi
```

And we can see that lxd has dropped support for the `lxd-agent-9p` service on this commit:
https://github.com/lxc/lxd/commit/2d013dfbeb4d10575989ab07e0a69022a6f9aa0b

### DIFF
--- a/pycloudlib/lxd/defaults.py
+++ b/pycloudlib/lxd/defaults.py
@@ -21,8 +21,7 @@ LXC_SETUP_VENDORDATA = textwrap.dedent(
               ICAgIFZJUlQ9JChzeXN0ZW1kLWRldGVjdC12aXJ0KQogICAgY2FzZSAkVklSVCBpbgogICAgICAg
               IHFlbXV8a3ZtKQogICAgICAgICAgICAoY2QgL3J1bi9seGRhZ2VudC8gJiYgLi9pbnN0YWxsLnNo
               KQogICAgICAgICAgICB1bW91bnQgL3J1bi9seGRhZ2VudAogICAgICAgICAgICBzeXN0ZW1jdGwg
-              c3RhcnQgbHhkLWFnZW50LTlwIGx4ZC1hZ2VudAogICAgICAgICAgICA7OwogICAgICAgICopCiAg
-              ICBlc2FjCmZpCg==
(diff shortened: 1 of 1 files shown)


blackboxsw (changes requested):
I think your commit message should reference the upstream lxd changeset that broke due to removal of the `/lib/systemd/system/lxd-agent-9p.service`

`https://github.com/lxc/lxd/commit/2d013dfbeb.`


Otherwise LGTM

blackboxsw:
Hrm I'm holding off on approval of this as I can't reproduce this issue locally (outside of ua-client integration tests).  Once I can properly reproduce with pycloudlib alone, i can +1 this. Until then, I'll hold on approviing.

lucasmoura:
@blackboxsw pycloudlib is not raising any failures if cloud-init fails to run since #93. This might explain why you are not seeing this error. But if you are explicitly checking the status of cloud-init and it is not failing, please let me know and I will work on it

blackboxsw:
Ok was able to reproduce the error today, right the issue was raise_on_error wasn't happening:
```bash
# /var/lib/cloud/scripts/per-once/setup-lxc.sh

LXD agent has been installed, reboot to confirm setup.
To start it now, unmount this filesystem and run: systemctl start lxd-agent
Failed to start lxd-agent-9p.service: Unit lxd-agent-9p.service not found.
```

Changing the service name s/lxd-agent-9p/lxd-agent/ works.

blackboxsw (approved):
LGTM!

PR #119, opened 2021-02-18
Update .gitignore

TheRealFalcon:
Turns out my language server will index everything not in `.gitignore`, and `.mypy_cache` was causing a big increase in memory usage.


OddBloke (approved):
Thanks!

PR #120, opened 2021-02-19
Support 'instance_type' for Azure instances

TheRealFalcon:
Azure was the only cloud that didn't include `instance_type` in it's launch.

Note that I DIDN'T put it at the end to keep as to keep it consistent with the other launch APIs. If people were passing args in a particular order to Azure launch, this will break them.

### DIFF
--- a/examples/az.py
+++ b/examples/az.py
@@ -40,6 +40,7 @@ def demo():
 
     instance = client.launch(
         image_id=image_id,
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
LGTM. Can we please add an example to examples/az.py with calling a different instance type?

passing instance_type=`Standard_DS2_v2` WFM.
Approved with a doc change.

Issue #75, opened 2020-12-10
`lxd/defaults.py` should have a "default" profile so we don't have to define a new one every 6 months


OddBloke:
focal+ are all the same currently (and we don't expect that to change); let's save ourselves some busywork.

(This is a more general version of #62, because I'm about to fix the issue for hirsute.)

PR #122, opened 2021-02-23
Drop support for xenial vstock image

lucasmoura:
Currently, LXD xenial virtual machines use a special image to allow supporting installing lxd-agent on the vm and
run commands with lxc exec. Since we are now running all the commands with ssh by default, we don't need to support that xenial image anymore. The implication now is that xenial vm will not be able to run lxc exec commands

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -13,7 +13,9 @@ class LXDInstance(BaseInstance):
     _type = 'lxd'
     _is_vm = None
 
-    def __init__(self, name, key_pair=None, execute_via_ssh=True):
+    def __init__(
+        self, name, key_pair=None, execute_via_ssh=True, series=None
+    ):
         """Set up instance.
 
         Args:
@@ -24,6 +26,7 @@ class LXDInstance(BaseInstance):
 
         self._name = name
         self.execute_via_ssh = execute_via_ssh
+        self.series = series
 
     def __repr__(self):
         """Create string representation for class."""
@@ -333,3 +336,24 @@ class LXDInstance(BaseInstance):
                 return
             time.sleep(1)
         raise TimeoutError
+
+
+class LXDVirtualMachineInstance(LXDInstance):
+    """LXD Virtual Machine backed instance."""
+
+    def _run_command(self, command, stdin):
+        """Run command in the instance."""
+        if self.execute_via_ssh:
+            return super()._run_command(command, stdin)
+
+        if self.series == "xenial":
+            msg = (
+                "Many xenial images do not support executing commands"
+                " via exec due to missing kernel support: you may see"
+                " unavoidable failures.\nSee"
+                " https://github.com/canonical/pycloudlib/issues/132 for"
+                " details."
+            )
+            self._log.warning(msg)
+
+        return super()._run_command(command, stdin)
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -6,7 +6,7 @@ import warnings
 import yaml
 
 from pycloudlib.cloud import BaseCloud
-from pycloudlib.lxd.instance import LXDInstance
+from pycloudlib.lxd.instance import LXDInstance, LXDVirtualMachineInstance
 from pycloudlib.util import subp
 from pycloudlib.constants import LOCAL_UBUNTU_ARCH
 from pycloudlib.lxd.defaults import base_vm_profiles, LXC_PROFILE_VERSION
@@ -35,6 +35,7 @@ class _BaseLXD(BaseCloud):
     _type = 'lxd'
     _daily_remote = 'ubuntu-daily'
     _releases_remote = 'ubuntu'
+    _lxd_instance_cls = LXDInstance
 
     def clone(self, base, new_instance_name):
         """Create copy of an existing instance or snapshot.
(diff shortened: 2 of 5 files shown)


pycloudlib/lxd/instance.py
```diff
@@ -333,3 +334,85 @@ def wait_for_stop(self):
...
+        Returns:
+            A string containing the base release from the instance
+        """
+        instance_info = self._lxc_instance_info()
+        release = None
+        try:
+            profiles = str(instance_info["Profiles"])
+
+            pattern_search = re.search(
+                r"pycloudlib-vm-(?P<release>.*)",
```
OddBloke, line 408: Once we fix #75, everything from focal and later should be sharing a profile; this determination will break.

Given that we know exactly the image ID that we're using in `_BaseLXD.launch`, we could pass that in as a parameter to the instance instantiation instead, and then call `self.cloud._extract_release_from_image_id` instead of performing this determination?

pycloudlib/lxd/instance.py
```diff
@@ -333,3 +334,85 @@ def wait_for_stop(self):
...
+            msg = (
+                "The xenial instance does not support executing commads"
+                " with exec."
+            )
+            raise Exception(msg)
+
+        base_cmd = [
+            'lxc', 'exec', self.name, '--', 'sudo', '-u', self.username, '--'
+        ]
+        return subp(base_cmd + list(command), rcs=None)
```
OddBloke, line 372: No need to repeat this implementation here, I think. We can probably do something like:

```py
if self.series == "xenial" and not self.execute_via_ssh:
   # raise/log here
super()._run_command(command, stdin)
```

That way we only have the actual `_run_command` code defined in a single place.

OddBloke (commented):
Thanks for this Lucas! Apologies my review has been so delayed. Overall, this looks like what we've talked about: thanks! I have some inline comments on some of the specifics.

OddBloke:
I've successfully run (albeit with the now-expected failures) the cloud-init integration tests using this code locally.

pycloudlib/lxd/instance.py
lucasmoura: This is true during a launch situation, but if we are using getting an existing instance through `cloud.get_instance` we would not have the `image ID` to get that information from. That's why I am extracting this from `lxc info` command.

But you are totally right that the regex will no longer work. I am fixing that now

pycloudlib/lxd/instance.py
lucasmoura: I don't know If I am missing something here, but I think this will not hold if we are using `xenial containers` which support both operations.

OddBloke: This implementation is in `LXDVirtualMachineInstance`, it shouldn't be called in the container situation.

pycloudlib/lxd/instance.py
OddBloke: > This is true during a launch situation, but if we are using getting an existing instance through `cloud.get_instance` we would not have the `image ID` to get that information from. That's why I am extracting this from `lxc info` command.

If we're getting an existing instance through `cloud.get_instance`, there's no guarantee that it will have a pycloudlib profile applied to it, I think? So I think we're always going to have corner cases around `get_instance`.

If we don't handle this, what's the consequence? I think it will only mean that we don't emit a warning log message, once my comment above is taken into account.

lucasmoura: Yes, I agree that the `cloud.get_instance` will have corner cases if the user created the vm outside pycloudlib and are now trying to use in pycloudlib. However, what will happen in that situation is just that we not alert the user that `exec`  commands will not work in xenial vms.

The reason I made this is to try to be explicit about warning users of possible problems in the code. But maybe I am overthinking this. Specially now that we are just emitting a warning instead of raising an `Exception`. If you agree that this solution is too complicated to just alert the user about potential xenial problems, just let me know and I will drop it

pycloudlib/lxd/instance.py
lucasmoura: Ohhh okay, now I see your point. I thought you meant adding that part to the `_run_command`  we have at `LXDInstance`. The proposed change makes sense. I will update the code

pycloudlib/lxd/instance.py
```diff
@@ -333,3 +334,82 @@ def wait_for_stop(self):
...
+        Returns:
+            A string containing the base release from the instance
+        """
+        instance_info = self._lxc_instance_info()
+        release = None
+        try:
+            profiles = str(instance_info["Profiles"])
+
+            pattern_search = re.search(
+                r"pycloudlib-vm-(?P<release>.*)-v.*",
```
OddBloke, line 405: This has been updated for the change in profile naming (thanks!), but this still doesn't address the fact that this will stop working entirely once we fix #75 (i.e. once we have a profile that's used for more than a single Ubuntu release, and therefore won't include a release name in its name). This determination also won't work for instances launched without pycloudlib (or even with pycloudlib, if the consumer specifies their own profile on launch).

Given this, I'd lean towards dropping this: it's quite a lot of code to "just" emit a warning, and it will likely have to be removed or substantially modified when #75 happens.

I think we agreed that you'd switch to just using the image ID for this determination, if we have it available?

lucasmoura: Got it. I have dropped all of the code that performed that verification and added the `image id` way of checking the release

OddBloke (commented):
One inline comment with a proposed minor improvement; LMK what you think.

pycloudlib/lxd/instance.py
```diff
@@ -333,3 +336,21 @@ def wait_for_stop(self):
...
+
+    def _run_command(self, command, stdin):
+        """Run command in the instance."""
+        if self.execute_via_ssh:
+            return super()._run_command(command, stdin)
+
+        if self.series == "xenial":
+            msg = (
+                "The xenial instance does not support executing commads"
+                " with exec."
```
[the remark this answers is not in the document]
lucasmoura: I think this is a good idea. Let me open that issue and update the code

lucasmoura:
@OddBloke Code updated

OddBloke (approved):
Thanks, sorry for all the delay getting it reviewed!

lucasmoura:
No worries at all @OddBloke

PR #124, opened 2021-02-25
Update naming convetion for LXD vm profiles

lucasmoura:
Currently, when we update a LXD vm profile we may have not see it reflected on subsequent runs because we do not generate a new profile if there is one with the same name on the system. To avoid that, we are now adding a version suffix to the profile name while also creating a md5sum test that will fail if we update the profile content but do not update the md5sum in the test.

PS: The way our test is structure does not prevent that a user may modify the profile, change the md5sums but do not bump the version name on the profile. However, I am proposing a guideline to incrementally add new profile version to the md5sum dict. I believe this documentation can help us reviewing future PRs that will change the profile. But if anyone has an idea to programmatically check that the version was bumped when a md5sum was modified, please let me know and I will implement it

### DIFF
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -9,7 +9,7 @@ from pycloudlib.cloud import BaseCloud
 from pycloudlib.lxd.instance import LXDInstance
 from pycloudlib.util import subp
 from pycloudlib.constants import LOCAL_UBUNTU_ARCH
-from pycloudlib.lxd.defaults import base_vm_profiles
+from pycloudlib.lxd.defaults import base_vm_profiles, LXC_PROFILE_VERSION
 
 
 class UnsupportedReleaseException(Exception):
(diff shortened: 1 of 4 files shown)


paride (changes requested):
In general this LGTM. I just found one issue: when a new release is added the unit test does not fail if `version_to_md5sum` is not updated adding it. This can easily be tested by dropping the `hirsute` entry from `version_to_md5sum` in your branch. I think the test should fail in this case, or that another small test should be added checking for the presence of a profile hash for all the releases.

lucasmoura:
@paride Good catch. I have updated the code to address that issue and I have also found another problem with the way we verify if a profile already exists

paride (approved):
LGTM, I'll let you do the merge so you can decide what commit type to use.

Nit: there's a typo in the ee4d4c6415e35816c07ad23ebf04730944036a58 commit message (`lxx` instead of `lxc`)

PR #128, opened 2021-03-03
Add banner timeout for paramiko connect

lucasmoura:
We are observing that a lot of tests that use pycloudlib are failing to due errors reading SSH protocol banner. Since we
cannot consistently reproduce those errors, we believe this should be related to a timeout issue. To address that we are now setting
the banner_timeout variable when we run paramiko connect method.

As an example, here is one Jenkins test that is being affected by this issue:
https://jenkins.canonical.com/server-team-jenkins-private/blue/rest/organizations/jenkins/pipelines/github-ci/pipelines/ubuntu-advantage-client/branches/PR-1414/runs/6/nodes/128/steps/153/log/?start=0

### DIFF
--- a/pycloudlib/instance.py
+++ b/pycloudlib/instance.py
@@ -35,6 +35,7 @@ class BaseInstance(ABC):
         self.port = '22'
         self.username = 'ubuntu'
         self.connect_timeout = 60
+        self.banner_timeout = 60
 
     @property
     @abstractmethod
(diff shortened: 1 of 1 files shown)


pycloudlib/instance.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, key_pair):
         self.port = '22'
         self.username = 'ubuntu'
         self.connect_timeout = 60
+        self.banner_timeout = 60
```
blackboxsw, line 38: While it would be nice to provide config overrides for this timeout, I think this is something that needs to be addressed separately as all these instance attrs should be configurable at some point in the future

blackboxsw (approved):

PR #134, opened 2021-03-19
Xenial LXD VMs: do file pull/push via sftp

paride:


pycloudlib/lxd/instance.py
```diff
@@ -195,6 +195,20 @@ def pull_file(self, remote_path, local_path):
...
+            super().pull_file(remote_path, local_path)
+            return
+
+        if self.series == "xenial":
+            msg = (
+                "Many xenial images do not support pulling files via"
+                " `lxc file pull` due to missing kernel support: you may"
+                " see unavoidable failures.\nSee"
+                " https://github.com/canonical/pycloudlib/issues/132 for"
+                " details."
```
OddBloke, line 208: As we now have this message three times, I wonder if we should capture the shared text in a separate variable? (Then it would be `.format("lxc file pull")` or something similar.)

paride: There is also "pulling files" / "pushing files" / "executing commands", but the warning can probably be made more generic ("the required functionality (format-string) requires lxd-agent ...") and save it in a variable, which can also be used by the tests.

paride: Refactored the warning a bit.

pycloudlib/lxd/tests/test_instance.py
```diff
@@ -36,12 +39,33 @@ def test_exec_with_run_command_on_xenial_machine(
...
+        assert any(self._missing_agent_msg in L for L in caplog.messages)
+        assert _m_subp.call_count == 1
+
+    @mock.patch("pycloudlib.lxd.instance.subp")
+    def test_file_pull_with_agent_on_xenial_machine(
+        self,
+        _m_subp,
+        caplog
+    ):
+        """Test exec does not work with xenial vm."""
```
OddBloke, line 51:
```suggestion
        """Test pull does not work with xenial vm."""
```

pycloudlib/lxd/tests/test_instance.py
```diff
@@ -36,12 +39,33 @@ def test_exec_with_run_command_on_xenial_machine(
...
+        assert any(self._missing_agent_msg in L for L in caplog.messages)
+        assert _m_subp.call_count == 1
+
+    @mock.patch("pycloudlib.lxd.instance.subp")
+    def test_file_push_with_agent_on_xenial_machine(
+        self,
+        _m_subp,
+        caplog
+    ):
+        """Test exec does not work with xenial vm."""
```
OddBloke, line 65:
```suggestion
        """Test push does not work with xenial vm."""
```

pycloudlib/lxd/instance.py
```diff
@@ -13,6 +13,12 @@ class LXDInstance(BaseInstance):
     _type = 'lxd'
     _is_vm = None
 
+    _missing_agent_msg = (
+        "Many Xenial images do not support `%s` due to missing lxd-agent:"
+        " you may see unavoidable failures.\n"
+        "See https://github.com/canonical/pycloudlib/issues/132 for details."
+    )
+
```
OddBloke, line 21: This is fine here, but it could probably just be a module-level constant: we don't need to be able to modify it within the inheritance hierarchy here.

pycloudlib/lxd/tests/test_instance.py
```diff
@@ -36,12 +39,33 @@ def test_exec_with_run_command_on_xenial_machine(
...
         instance._run_command(["test"], None)
-        expected_msg = (
-            "Many xenial images do not support executing commands"
-            " via exec due to missing kernel support: you may see"
-            " unavoidable failures.\nSee"
-            " https://github.com/canonical/pycloudlib/issues/132 for"
-            " details."
-        )
-        assert expected_msg in caplog.messages
+        assert any(self._missing_agent_msg in L for L in caplog.messages)
```
OddBloke, line 42: Could this be simplified to `assert self._missing_agent_msg in caplog.text`? (This applies below, too.)

pycloudlib/lxd/tests/test_instance.py
```diff
@@ -36,12 +39,33 @@ def test_exec_with_run_command_on_xenial_machine(
...
+        assert _m_subp.call_count == 1
```
OddBloke, line 43: The leading underscore on `_m_subp` is a convention to indicate that it's unused: as we're now using it, we should remove the underscore. (This applies below too, of course.)

OddBloke (commented):
Thanks Paride! A few test clean-ups suggested, as well as some other thoughts.

pycloudlib/lxd/tests/test_instance.py
paride: Not really, as `caplog.text` is a list of strings and `self._missing_agent_msg in caplog.text` check if `self._missing_agent_msg` matches one of the list elements (exactly), while we want to to do a substring match here.

Not saying there isn't a smarter way to do so :-)

paride: I mixed up `.messages` and `.text`, your suggestion works. Thanks!

pycloudlib/lxd/instance.py
paride: Done, let me know if you had anything different in mind.

(I'm doing separate commits for these fixups, we'll squash before merging.)

pycloudlib/lxd/tests/test_instance.py
paride: Underscore dropped.

Issue #129, opened 2021-03-03
lxd.ip: intermittent error: IndexError on failed `lxc list {} -c 4`


blackboxsw:
LXDInstance.ip should either perform better error handling or better failure logging would be helpful in triage if we get an error from `lxc list {} -c 4 --format csv`

The code doesn't check `result.ok|failed` it just exits if some content is seen on result and fails to parse it.

```bash
      Traceback (most recent call last):
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1414@2/.tox/behave-lxd-18.04/lib/python3.6/site-packages/behave/model.py", line 1329, in run
          match.run(runner.context)
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1414@2/.tox/behave-lxd-18.04/lib/python3.6/site-packages/behave/matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "features/steps/steps.py", line 70, in given_a_machine
          image_name=context.series_image_name[series],
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1414@2/features/cloud.py", line 140, in launch
          user_data=user_data,
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1414@2/features/cloud.py", line 669, in _create_instance
          name=instance_name, image_id=image_name, user_data=user_data
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1414@2/.tox/behave-lxd-18.04/lib/python3.6/site-packages/pycloudlib/lxd/cloud.py", line 282, in launch
          instance.start(wait)
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1414@2/.tox/behave-lxd-18.04/lib/python3.6/site-packages/pycloudlib/lxd/instance.py", line 317, in start
          self.wait()
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1414@2/.tox/behave-lxd-18.04/lib/python3.6/site-packages/pycloudlib/instance.py", line 97, in wait
          self._wait_for_execute()
        File "/jenkins/servers/private/workspace/_ubuntu-advantage-client_PR-1414@2/.tox/behave-lxd-18.04/lib/python3.6/site-packages/pycloudlib/instance.py", line 376, in _wait_for_execute
          result = self.execute(test_instance_command)
[...]

PR #135, opened 2021-03-25
lxd.ip: better instance name match and parsing of lxc ls output

blackboxsw:
lxd.ip: better instance name match and parsing of lxc ls output

Use exact name matching via lxc ls "^<name>%" to avoid matching
multiple vms with common name prefix.

Add try/except blocked to handle, ignore and retry on unexpected
responses while lxc brings up the instance networking.

Add some logging for unparseable output so we can better handle
those cases if they result in irresolvable instance state.

Raise TimeoutError if our 150 retries are exhausted

Fixes: #129

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -81,21 +81,37 @@ class LXDInstance(BaseInstance):
         Returns:
             IP address assigned to instance.
 
+        Raises: TimeoutError when exhausting retries trying to parse lxc list
+            for ip addresses.
         """
         retries = 150
 
         while retries != 0:
-            command = 'lxc list {} -c 4 --format csv'.format(self.name)
-            result = subp(command.split()).stdout
-
-            if result != '':
-                break
-
+            command = [
+                 'lxc', 'list', '^{}$'.format(self.name), '-c4',
+                 '--format', 'csv'
+            ]
+            result = subp(command)
+            if result.ok and result.stdout:
+                ip_address = None
+                try:
+                    # Expect "<ip> (<interface>)" when network fully configured
+                    ip_address, _dev = result.stdout.split()
+                except ValueError:
+                    self._log.debug(
+                        "Unable to parse output of cmd: %s. Expected"
+                        " <ip> (<interface>), got: %s. Retrying %d time(s)...",
+                        command, result.stdout, retries
+                    )
+                if ip_address:
+                    return ip_address
             retries -= 1
             time.sleep(1)
-
-        ip_address = result.split()[0]
-        return ip_address
+        raise TimeoutError(
+            "Unable to determine IP address after 150 retries."
+            " exit:{} stdout: {} stderr: {}".format(
+                result.return_code, result.stdout, result.stderr)
+        )
 
     @property
     def ephemeral(self):
--- a/pycloudlib/lxd/tests/test_instance.py
+++ b/pycloudlib/lxd/tests/test_instance.py
@@ -1,7 +1,11 @@
 """Tests for pycloudlib.lxd.instance."""
+import re
 from unittest import mock
 
+import pytest
+
 from pycloudlib.lxd.instance import LXDInstance, LXDVirtualMachineInstance
+from pycloudlib.result import Result
 
 
 class TestExecute:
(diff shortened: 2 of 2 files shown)


OddBloke (changes requested):
Looking at the help for `lxc list`, I see in the Filters section that:

> A single keyword like "web" which will list any instance with a name starting by "web".

which I see locally:

```
$ lxc list bright-moose -c n4 --format csv
bright-moose,10.236.135.57 (eth0)
bright-moose-2,10.236.135.43 (eth0)
```

However, I also see in that section:

> A regular expression on the instance name. (e.g. .*web.*01$).

and from my local testing, we can match on the exact name using `^` and `$`:

```
$ lxc list ^bright-moose$ -c n4 --format csv
bright-moose,10.236.135.57 (eth0)
```

So that looks like a preferable path forward, to me.

blackboxsw:
Thanks @OddBloke I RTFM'd and that makes the code way simpler. take a look when you get a chance

pycloudlib/lxd/instance.py
```diff
@@ -81,21 +81,37 @@ def ip(self):
...
+            command = [
+                 'lxc', 'list', '^{}$'.format(self.name), '-c4',
+                 '--format', 'csv'
+            ]
+            result = subp(command)
+            if result.ok and result.stdout:
+                ip_address = None
+                try:
+                    # Expect "<ip> (ethX)" when network fully configured
+                    ip_address, _dev = result.stdout.split()
```
OddBloke, line 99: I don't believe I've seen a case where this split has failed: do we have a motivating error case for this?

pycloudlib/lxd/instance.py
```diff
@@ -81,21 +81,37 @@ def ip(self):
...
+                except ValueError:
+                    self._log.debug(
+                        "Unable to parse output of cmd: %s. Expected"
+                        " <ip> (ethX), got: %s. Retrying %d time(s)...",
```
OddBloke, line 103: Interfaces in VMs generally don't start with `eth`:

```suggestion
                        " <ip> (<interface>), got: %s. Retrying %d time(s)...",
```

pycloudlib/lxd/tests/test_instance.py
```diff
@@ -69,3 +72,81 @@ def test_file_push_with_agent_on_xenial_machine(
...
+        if isinstance(stdout, list):
+            # Create a list of responses based on stdout being a list
+            m_subp.side_effect = [
+                Result(stdout=out, stderr=stderr, return_code=return_code)
+                for out in stdout
+            ]
+        else:
+            m_subp.return_value = Result(
+                stdout=stdout, stderr=stderr, return_code=return_code
+            )
```
OddBloke, line 139: If you modified the non-list parameters to be single item lists, could this be simplified to:

```suggestion
        m_subp.side_effect = [
            Result(stdout=out, stderr=stderr, return_code=return_code)
            for out in stdout
        ]
```

?

pycloudlib/lxd/tests/test_instance.py
```diff
@@ -69,3 +72,81 @@ def test_file_push_with_agent_on_xenial_machine(
...
+        instance = LXDInstance(name="my_vm")
+        lxc_mock = mock.call(
+            ["lxc", "list", "^my_vm$", "-c4", "--format", "csv"]
+        )
+        if isinstance(expected, Exception):
+            with pytest.raises(type(expected)) as exc_info:
```
OddBloke, line 145: `raises` has a `match` kwarg which we could use here: https://docs.pytest.org/en/stable/reference.html#pytest-raises

OddBloke: I suspect that you could further use that to simplify this test by using a noop contextmanager per https://docs.pytest.org/en/stable/example/parametrize.html#parametrizing-conditional-raising. That's perhaps an interesting exercise, but is very much a nit.

OddBloke (approved):
One inline log change I'd like to see, plus some test comments; feel free to land this after addressing the log issue though

pycloudlib/lxd/tests/test_instance.py
blackboxsw: While it felt dirty to stuff different types into the parametrized values, if we try to use side_effect for the mocked subp, we have to bloat the side_effect * sleeps (which is 150) to avoid mock stopiteration issues. That said, I don't like type checking in the unittest. I'll use the one item list for stdouts. and check len to determine if we need to use side_effect vs a single uniform return_value mocked.

pycloudlib/lxd/tests/test_instance.py
blackboxsw: +1 on the first suggestion and applied. Noted on second, but given the smell of me stuffing strings vs Exceptions into the parametrized "expected" param I suspect this exercise is more than it's worth. That said there are many other places where the noop context mgr could be used in testing here, ua-client and cloud-init so I'll keep that link/reference handy.

blackboxsw:
Ran successful lxd integration tests for uaclient on xenial and bionic without regressions here. (on top of this CI run)

PR #140, opened 2021-05-06
Drop trusty support for launching lxd instances

lucasmoura:
We have dropped the support for trusty in [ubuntu-advantage-client](https://github.com/canonical/ubuntu-advantage-client/pull/1611). Since this is the only project we are aware that had the need for trusty lxd support, we are dropping that support from pycloudlib.

PS: I have kept the trusty check on LXD VMs, to alert the user that a trusty vm cannot be launched, but I can drop this as well if needed

### DIFF
--- a/docs/clouds/azure.md
+++ b/docs/clouds/azure.md
@@ -47,8 +47,8 @@ Azure requires an SSH key to be uploaded before using it. See the SSH Key page f
 To find latest daily Azure image for a release of Ubuntu:
 
 ```python
-azure.daily_image('trusty')
-"Canonical:UbuntuServer:14.04.0-LTS",
+azure.daily_image('xenial')
+"Canonical:UbuntuServer:16.04-DAILY-LTS"
 ```
 
 The return Azure image can then be used for launching instances.
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -19,7 +19,6 @@ class Azure(BaseCloud):
     _type = 'azure'
 
     UBUNTU_RELEASE = {
-        "trusty": "Canonical:UbuntuServer:14.04.0-LTS",
         "xenial": "Canonical:UbuntuServer:16.04-DAILY-LTS",
         "bionic": "Canonical:UbuntuServer:18.04-DAILY-LTS",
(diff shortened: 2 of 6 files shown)


paride:
This LGTM, one question: is this meant to be LXD specific for some reason, or should it be a more general post-Trusty cleanup? There are a few more places where Trusty appears (line numbers are from your branch):
```
pycloudlib/tests/test_instance.py
105:                # `cloud-init status --help` on trusty returns non-zero
144:                # mirror trusty's behaviour)

pycloudlib/azure/cloud.py
22:        "trusty": "Canonical:UbuntuServer:14.04.0-LTS",

pycloudlib/lxd/cloud.py
698:        if release == "trusty":
699:            # trusty is not supported on LXD vms
701:                release="trusty",

pycloudlib/lxd/tests/test_cloud.py
182:    def test_trusty_image_not_supported_when_launching_vms(
185:        """Tests searching for trusty image for launching LXD vms."""
192:                release="trusty",
195:        assert "Release trusty is not supported for LXD vms" == str(

docs/clouds/azure.md
50:azure.daily_image('trusty')
```
then in `pycloudlib/instance.py` there is this hacky wait logic:
```
        has_wait = "--wait" in result.stdout                                        
                                                                                    
        if has_wait:                                                                
            cmd = ["cloud-init", "status", "--wait", "--long"]                      
        else:                                                                       
            # runlevel 'N 2' supports distros without recent cloud-init             
            # (e.g. trusty).                                                        
            runlevel_result = (                                                     
                '[ "$(runlevel)" = "N 2" ] && '                                     
                "[ -f /run/cloud-init/result.json ]"                                
            )                                                                       
            cmd = (                                                                 
                "i=0; while [ $i -lt {} ] && i=$(($i+1)); do {} && exit 0;"         
                " sleep 1; done; exit 1".format(                                    
                    self.boot_timeout, runlevel_result                              
                )                                                                   
            )                                                                       
```
which is a workaround for `cloud-init status --wait` being missing in the other releases. Do we need to keep it?

lucasmoura:
@paride good point on that. I have kept some of trusty checks for LXD vms, but thinking of it now, It seems unnecessary.
Also, thanks for spotting the other trusty references here, I have removed them as well

paride (approved):
LGTM, one nit: the commit message still refers to LXD, while the commit now drop Trusty support in general. I'm leaving the merge to you in case you want to change that.

PR #142, opened 2021-05-12
gce: raise IndexError if no image found

OddBloke:
This matches the behaviour of other platforms, which raise either
IndexError or ValueError to indicate that no daily/release image has
been found.  This fixes specifying a non-default GCE image in
cloud-init's integration testing.

(This isn't ideal, of course, but this commit at least makes it
consistent.)


TheRealFalcon (approved):
Thanks!

PR #143, opened 2021-05-12
ec2: wait for console log output

OddBloke:
On EC2, console log output is not available immediately after launching
an instance: we've seen it take up to 2 minutes for the API to start
returning content.  Instead of, arguably incorrectly, claiming that
there is no console output in these cases, we retry for up to 3 minutes
until we get _some_ content.

### DIFF
--- a/pycloudlib/ec2/instance.py
+++ b/pycloudlib/ec2/instance.py
@@ -109,12 +109,15 @@ class EC2Instance(BaseInstance):
             The console log or error message
 
         """
-        try:
-            # OutputBytes comes from platform._decode_console_output_as_bytes
+        start = time.time()
+        while time.time() < start + 180:
             response = self._instance.console_output()
-            return response['OutputBytes']
-        except KeyError:
-            return 'No Console Output [%s]' % self._instance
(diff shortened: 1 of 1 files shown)


OddBloke:
Marked wip because I know @TheRealFalcon wanted to consider a parameter for this behaviour.

pycloudlib/ec2/instance.py
```diff
@@ -109,11 +109,15 @@ def console_log(self):
...
+        while time.time() < start + 180:
             response = self._instance.console_output()
-            return response['OutputBytes']
-        except KeyError:
+            try:
+                return response['Output']
+            except KeyError:
+                self._log.debug("Console output not yet available; sleeping")
+                time.sleep(5)
+        else:
```
TheRealFalcon, line 120: I don't think this else is necessary here as there's no break in the look.

TheRealFalcon (approved):
I was thinking of adding a `wait` parameter, but defaulted to `True`. I'm ok merging as-is, and if I don't want to wait I can always add the parameter later.

PR #146, opened 2021-05-12
openstack: wait for console log output

TheRealFalcon:
Openstack wants to be cool like ec2 (see #143).


OddBloke (approved):
Thanks! Couple of thoughts inline, but nothing blocking.

PR #147, opened 2021-05-25
Make OCI's availability domain required (SC-59)

### LINKED ISSUE #138: Null availability_domain for OCI
https://github.com/canonical/pycloudlib/blob/main/pycloudlib/oci/cloud.py#L185 used to give a valid value, but now it's just giving 'Null'.

I didn't have time to debug, so as a quick workaround, I hardcoded an availability zone:
`availability_domain = 'qIZq:PHX-AD-3' `

With this, tests run again. We should look to see if there's a better long-term solution.

TheRealFalcon:
availability_domain used to be able to be inferred from the subnet.
Now it is required to be passed manually.
https://docs.oracle.com/en-us/iaas/tools/python/2.38.3/api/core/models/oci.core.models.Subnet.html#oci.core.models.Subnet.availability_domain

Fixes #138

### DIFF
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -20,8 +20,8 @@ class OCI(BaseCloud):
     _type = 'oci'
 
     def __init__(
-        self, tag, timestamp_suffix=True, compartment_id=None,
-        availability_domain=None, config_path='~/.oci/config',
+        self, tag, *, availability_domain, timestamp_suffix=True,
+        compartment_id=None, config_path='~/.oci/config'
     ):
         """
         Initialize the connection to OCI.
@@ -40,6 +40,8 @@ class OCI(BaseCloud):
             config_path: Path of OCI config file
         """
         super().__init__(tag, timestamp_suffix)
+        self.availability_domain = availability_domain
+
         if not compartment_id:
             command = ['oci', 'iam', 'compartment', 'get']
             exception_text = (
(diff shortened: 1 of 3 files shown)


pycloudlib/oci/cloud.py
```diff
@@ -21,7 +21,7 @@ class OCI(BaseCloud):
 
     def __init__(
         self, tag, timestamp_suffix=True, compartment_id=None,
-        config_path='~/.oci/config',
+        availability_domain=None, config_path='~/.oci/config',
```
lucasmoura, line 24: I know this the default value here is probably to maintain backwards compatibility, but since this a necessary parameter now I think we should not have a default value for it.

Also, I think we should also update the `oci` example file to include that change as well

TheRealFalcon: Yeah, I have always been of the opinion that the common arguments between the clouds (tag and timestamp_suffix) should go first in `__init__`. Since timestamp_suffix has a default, everything after it would also have to have a default, including "required" arguments.

I know we already have another cloud or two that breaks this, but I was hoping to change it to be consistent, not the other way around. I fine doing whatever the team thinks is best though.

Thanks for the reminder about the example

lucasmoura: That's a good point about the consistency. Fair enough. I don't see a problem keeping it as it is to enforce that

lucasmoura (approved):

Issue #110, opened 2021-01-28
SSH connection retrying is handled at multiple levels


OddBloke:
When attempting to run a command via SSH on a host, we call `_ssh_(command, ...)` which, in turn, calls `_ssh_connect`. In [`_ssh_connect`, we retry 60 times, sleeping 10 seconds in-between](https://github.com/canonical/pycloudlib/blob/master/pycloudlib/instance.py#L315-L337):

```py
        retries = 60
        last_exception = None
        while retries:
            try:
                ...
                return client
            except ... as e:
                ...
                retries -= 1
                time.sleep(10)
```

If this fails, however, then `_ssh` _also_ retries: https://github.com/canonical/pycloudlib/blob/ef76d90f8412a932e92212bc2ec0680c109fb324/pycloudlib/instance.py#L263-L269

600s is plenty of time for an SSH connection to happen if it's going to: we should rationalise this waiting so we only perform it in one place.

TheRealFalcon:
This was fixed in #148

PR #148, opened 2021-05-28
Move SSH retry logic to boot wait method (SC-58)

TheRealFalcon:
List of changes are in the respective commits (please don't squash merge them). If this PR is approved I'll take care of getting PR commits getting squashed into the right commit.

In general, moved all the SSH retry logic to the wait for boot section. The connection timeout should take care of the random network blips.

I also provided a base method for restart. Every cloud that issued a restart signal then waited for boot (including the fancy ec2 implementation) had occasional reboot issues. Are there any situations we know of where we'd need to explicitly send a restart signal rather than just a shutdown then start?

I also fixed some cloud-specific boot issues.

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -271,8 +271,18 @@ class LXDInstance(BaseInstance):
         """
         self._log.debug('restarting %s', self.name)
 
-        self.shutdown(wait=True, force=force)
-        self.start(wait=wait)
+        # Since restart is always blocking, use restart if wait is True
+        cmd = ["lxc", "restart", self.name]
+        if force:
+            cmd.append("--force")
+        if wait:
+            subp(cmd)
+            self.wait()
+        else:
+            # If wait is False, this is a faster way to restart without
+            # blocking
+            self.shutdown(wait=True, force=force)
+            self.start(wait=False)
 
     def restore(self, snapshot_name):
         """Restore instance from a specific snapshot.
@@ -371,10 +381,38 @@ class LXDInstance(BaseInstance):
         """Wait for instance stop."""
         self._log.debug('waiting for stop: %s', self.name)
         for _ in range(100):
-            result = subp(
-                'lxc list {} -c s --format csv'.format(self.name).split()
-            )
+            result = subp([
+                'lxc', 'list', '^{}$'.format(self.name), '-cs',
+                '--format', 'csv'
+            ])
+
             if result == 'STOPPED':
+                return
+            time.sleep(1)
+        raise TimeoutError
+
+    def _wait_for_instance_start(self):
+        """Wait for the cloud instance to be up.
+
+        LXD VMs need to install systemd units upon initialization. There is
+        no easy way to do this and also enable them on boot, so an LXD VM
+        will reboot as part of its initialization process. It is possible
+        we have connected the VM before this reboot occurs, so then any
+        following SSH connections will fail.
+
+        The VM doesn't accurately report the number of processes until
+        the initialization is fully complete, so block until our number
+        of processes isn't -1.
+        """
+        processes = -1
+        for _ in range(300):
+            try:
+                processes = int(subp(
+                    'lxc list -c N {} -f csv'.format(self.name).split()
+                ))
+            except ValueError:
+                pass
+            if processes > -1:
                 return
             time.sleep(1)
         raise TimeoutError
--- a/pycloudlib/oci/instance.py
+++ b/pycloudlib/oci/instance.py
@@ -99,6 +99,12 @@ class OciInstance(BaseInstance):
         Args:
             wait: wait for the instance to shutdown
         """
+        # Because it's like 1995 or something...
+        # But seriously, without a sync (or 5), if you power down too quickly
+        # after first boot, all the keys created in /etc/ssh will turn into
+        # zero-byte files and you'll be permanently kicked out of your
+        # instance with your only hope being serial console access.
+        self.execute('sync; sync; sync; sync; sync')
         self.compute_client.instance_action(self.instance_data.id, 'STOP')
         if wait:
             self.wait_for_stop()
--- a/pycloudlib/instance.py
+++ b/pycloudlib/instance.py
@@ -156,6 +156,8 @@ class BaseInstance(ABC):
         Returns:
             Result object
 
+        Raises SSHException if there are any problem with the ssh connection
+
         """
         if isinstance(command, str):
             command = ['sh', '-c', command]
@@ -197,6 +199,8 @@ class BaseInstance(ABC):
         Args:
             remote_path: path on remote instance
             local_path: local path
+
+        Raises SSHException if there are any problem with the ssh connection
         """
         self._log.debug('pulling file %s to %s', remote_path, local_path)
 
@@ -209,6 +213,8 @@ class BaseInstance(ABC):
         Args:
             local_path: local path
             remote_path: path on remote instance
+
+        Raises SSHException if there are any problem with the ssh connection
         """
         self._log.debug('pushing file %s to %s', local_path, remote_path)
 
@@ -225,6 +231,7 @@ class BaseInstance(ABC):
         Returns:
             result from script execution
 
+        Raises SSHException if there are any problem with the ssh connection
         """
         # Just write to a file, add execute, run it, then remove it.
         shblob = '; '.join((
@@ -264,15 +271,11 @@ class BaseInstance(ABC):
 
         """
         cmd = shell_pack(command)
-        for _ in range(10):
-            try:
-                client = self._ssh_connect()
-                fp_in, fp_out, fp_err = client.exec_command(cmd)
-                break
-            except (ConnectionResetError, NoValidConnectionsError) as e:
-                last_error = e
-        else:
-            raise last_error  # noqa
+        client = self._ssh_connect()
+        try:
+            fp_in, fp_out, fp_err = client.exec_command(cmd)
+        except (ConnectionResetError, NoValidConnectionsError) as e:
+            raise SSHException from e
         channel = fp_in.channel
 
         if stdin is not None:
@@ -316,37 +319,21 @@ class BaseInstance(ABC):
         except SSHException:
             pass
 
-        start = time.time()
-        end = start + 600
-        last_exception = None
-        while True:
-            try:
-                client.connect(
-                    username=self.username,
-                    hostname=self.ip,
-                    port=int(self.port),
-                    timeout=self.connect_timeout,
-                    banner_timeout=self.banner_timeout,
-                    key_filename=self.key_pair.private_key_path,
-                )
-                self._ssh_client = client
-                return client
-            except (ConnectionRefusedError, AuthenticationException,
-                    BadHostKeyException, ConnectionResetError, SSHException,
-                    OSError) as e:
-                last_exception = e
-            if time.time() > end:
-                break
-            self._log.info(
-                "%s\nRetrying SSH connection to %s@%s:%s (%ds left)",
-                last_exception, self.username, self.ip, self.port,
-                end - time.time()
+        try:
+            client.connect(
+                username=self.username,
+                hostname=self.ip,
+                port=int(self.port),
+                timeout=self.connect_timeout,
+                banner_timeout=self.banner_timeout,
+                key_filename=self.key_pair.private_key_path,
             )
-            time.sleep(1)
-
-        self._log.error('Failed ssh connection to %s@%s:%s after 10 minutes',
-                        self.username, self.ip, self.port)
-        raise last_exception
+        except (ConnectionRefusedError, AuthenticationException,
+                BadHostKeyException, ConnectionResetError, SSHException,
+                OSError) as e:
+            raise SSHException from e
+        self._ssh_client = client
+        return client
 
     def _sftp_connect(self):
         """Connect to instance via SFTP."""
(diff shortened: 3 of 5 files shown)


TheRealFalcon:
Ran the following (via cloud-init's integration tests) on all supported clouds as a means to verify this. The first `execute` and `pull_file` are just ensuring these commands work as expected.

```
def test_stuff(setup_image, session_cloud: IntegrationCloud):
    for i in range(5):
        with session_cloud.launch() as client:
            for j in range(10):
                print(client.execute('cloud-init --version'))
                print(client.pull_file('/var/log/cloud-init.log', '/tmp/delete_me'))
                client.execute('echo hi > /tmp/test')
                client.restart()
                assert not client.execute('ls /tmp/test').ok
```

pycloudlib/ec2/instance.py
```diff
@@ -127,65 +123,6 @@ def delete(self, wait=True):
         if wait:
             self.wait_for_delete()
 
-    def restart(self, wait=True, **kwargs):
```
TheRealFalcon, line 130: Do we know any reason why EC2 had this particular implementation?

orndorffgrant: After reading through this implementation, I don't _think_ we're losing anything by switching to the stop-then-start implementation. But it looks like @paride was the last one to touch this method so he may have a more definitive opinion (though it appears the last significant edit was about 2 years ago)

orndorffgrant: Looking around, the main downside I've found to stop/start is that the IP address will change, but I don't think that will affect us.
The other downside is that you lose anything stored on "ephemeral"/"instance" storage, but you have to go out of your way to use that, and I don't think we're using it at all (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-volumes)

pycloudlib/oci/instance.py
```diff
@@ -83,22 +83,18 @@ def delete(self, wait=True):
...
 
         Args:
             wait: wait for the instance to shutdown
         """
+        # Because it's like 1995 or something...
+        # But seriously, without a sync (or 5), if you power down too quickly
+        # after first boot, all the keys created in /etc/ssh will turn into
+        # zero-byte files and you'll be permanently kicked out of your
+        # instance with your only hope being serial console access.
+        self.execute('sync; sync; sync; sync; sync')
```
orndorffgrant, line 97: LOL - is the number of syncs here specific? As in: would 4 work? or 3? Or do we need all 5 to ensure this will work?

lucasmoura (commented):
I have tested this change on all types of uaclient integration tests and the only one that failed is the vm tests.

However, the lxd test are actually slower after introducing that change. I have used this [scenario](https://github.com/canonical/ubuntu-advantage-client/b) to test it:

Running locally without this changes: 2.50min
With the change: 3.30min

This is the average of running that test 3 times for each scenario.

pycloudlib/lxd/instance.py
```diff
@@ -379,6 +379,32 @@ def wait_for_stop(self):
             time.sleep(1)
         raise TimeoutError
 
+    def _wait_for_instance_start(self):
```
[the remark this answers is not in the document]
TheRealFalcon: That's unfortunate. I am also on 4.14 (latest/stable)

Curious what happens when you try `lxc list -c N <vm_name> -f csv` while the test is waiting for the VM. Does it ever return anything meaningful? Are you able to shell/ssh to it while the command isn't returning valid info?

I'm assuming you're using the `pycloudlib-vm-bionic-v2` profile? It works for me when I use that profile.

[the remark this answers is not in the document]
orndorffgrant: @lucasmoura @TheRealFalcon I played around with uaclient tests and this these pycloudlib changes a bit and discovered a couple things.

1. `tox -e behave-vm-20.04` succeeds for uaclient with these changes
2. `tox -e behave-vm-18.04` times out with these changes due to this code when bringing up the temporary image-build vm
3. `tox -e behave-vm-18.04` _without_ these changes works but the temporary image-build vm never gets a positive process count.
    a. I was unable to `lxc exec` anything on the instance manually, but, weirdly, our test code is able to successfully run commands on the instance. 
    b. later in the same test, it brings up a second vm. The second vm does get a positive process count and I can `lxc exec` on it.
    c. the difference between our commands to bring up the image-build vm and the second vm are that we include some cloud-init user-data for the image-build vm. That must be breaking something

So I think this might count as a bug in our uaclient tests :upside_down_face: and these changes are probably fine.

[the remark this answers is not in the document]
orndorffgrant: I confirmed that the test failures should be considered a bug in uaclient - not this PR - and I put up a uaclient PR to fix it: https://github.com/canonical/ubuntu-advantage-client/pull/1665

pycloudlib/lxd/instance.py
```diff
@@ -379,6 +379,32 @@ def wait_for_stop(self):
...
+
+        The VM doesn't accurately report the number of processes until
+        the initialization is fully complete, so block until our number
+        of processes isn't -1.
+        """
+        processes = -1
+        for _ in range(300):
+            try:
+                processes = int(subp(
+                    'lxc list -c N {} -f csv'.format(self.name).split()
```
blackboxsw, line 399: We need to be more strict in any calls to lxc list because it does a prefix match on names which may match multiple VMs in CI with common prefix.

Also in what cases do we need "split" with self.name?

```suggestion
                    ['lxc', 'list', '-cN',  '^{}$'.format(self.name),  '-f', 'csv']
```

blackboxsw (changes requested):
Thanks for this @TheRealFalcon. Minor nits on lxc ls calls needing to specify a strict regex. Also i'm thining we probably want to retain the general functionality to use the platform's restart operation instead of discrete shutdown/start because (in theory) the platform should be doing this better/more effeciently than our separate discrete calls. 

Also, given that LXC does have a blocking restart operation,  I also have a patch that seems to improve LXC test performance for simple launch, restart by about 30%.
```diff
commit 0363231e0916321d79e5e79761acd4cb4acec5c9 (HEAD)
Author: Chad Smith <chad.smith@canonical.com>
Date:   Fri Jun 4 12:55:37 2021 -0600

    lxd: restart should use platform restart logic when wait=True
    
    Otherwise use separate inst.shutdown(wait=True) start(wait=False).
    
    Using the LXD platform restart operation directly actually improves
    overall performance by about 30% as the platform handles blocking
    on shutdown/restart better than two separate lxc cmds.

diff --git a/pycloudlib/lxd/instance.py b/pycloudlib/lxd/instance.py
index 69fe1f1..eec7e6c 100644
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -270,9 +270,14 @@ class LXDInstance(BaseInstance):
             force: boolean, force instance to shutdown before restart
         """
         self._log.debug('restarting %s', self.name)
-
-        self.shutdown(wait=True, force=force)
-        self.start(wait=wait)
+        cmd = ["lxc", "restart", self.name]
+        if force:
+            cmd.append("--force")
+        if wait:
+            subp(cmd)
+        else:
+            self.shutdown(wait=True, force=force)
+            self.start(wait=wait)
 
     def restore(self, snapshot_name):
         """Restore instance from a specific snapshot.
@@ -372,7 +377,10 @@ class LXDInstance(BaseInstance):
         self._log.debug('waiting for stop: %s', self.name)
         for _ in range(100):
             result = subp(
-                'lxc list {} -c s --format csv'.format(self.name).split()
+                 [
+                     'lxc', 'list', '^{}$'.format(self.name), '-cs',
+                     '--format', 'csv'
+                 ]
             )
             if result == 'STOPPED':
                 return
diff --git a/pycloudlib/lxd/tests/test_instance.py b/pycloudlib/lxd/tests/test_instance.py
index b1c4ca6..cf99dc1 100644
--- a/pycloudlib/lxd/tests/test_instance.py
+++ b/pycloudlib/lxd/tests/test_instance.py
@@ -8,6 +8,42 @@ from pycloudlib.lxd.instance import LXDInstance, LXDVirtualMachineInstance
 from pycloudlib.result import Result
 
 
+class TestRestart:
+    """Tests covering pycloudlib.lxd.instance.Instance.restart."""
+
+    @pytest.mark.parametrize(
+        "wait,force,cmd",
+        (
+             (True, True, ["lxc", "restart", "my_vm", "--force"]),
+             (True, False, ["lxc", "restart", "my_vm"]),
+             # When wait is false, call shutdown and start
+             (False, False, []), (False, True, [])
+        )
+    )
+    @mock.patch("pycloudlib.lxd.instance.LXDInstance.start")
+    @mock.patch("pycloudlib.lxd.instance.LXDInstance.shutdown")
+    @mock.patch("pycloudlib.lxd.instance.subp")
+    def test_restart_calls_lxc_cmd_with_force_param(
+        self, subp, shutdown, start, wait, force, cmd
+    ):
+        """When wait=True, honor force param on lxc restart cmdline.
+
+        When wait is False will wait on shutdown and not wait on start.
+        """
+        instance = LXDInstance(name="my_vm")
+        instance.restart(force=force, wait=wait)
+        if wait:
+            assert [mock.call(cmd)] == subp.call_args_list
+            assert 0 == shutdown.call_count
+            assert 0 == start.call_count
+        else:
+            assert 0 == subp.call_count
+            assert [
+                mock.call(wait=True, force=force)
+            ] == shutdown.call_args_list
+            assert [mock.call(wait=wait)] == start.call_args_list
+
+
 class TestExecute:
     """Tests covering pycloudlib.lxd.instance.Instance.execute."""
 
```

pycloudlib/lxd/instance.py
TheRealFalcon: > Also in what cases do we need "split" with self.name?

The split functions on the whole string (post formatting). Only reason is I find it easier to define the command as a string and then split it afterwords, but defining the command as a list works too.

TheRealFalcon:
> as that operation will be exercising different platform logic relative to the intsances under test.

I'm fine removing the behavior, but I'm not sure I understand this reasoning. Why do we care about exercising different platform logic? From the perspective of cloud-init, I care about what happens while the instance is booting. Whether that happened due to a restart signal vs a shutdown/start doesn't really matter to me. Should it? I'm assuming it's the same for UA, but maybe there are use cases I'm not understanding.

TheRealFalcon:
@blackboxsw Thanks! I think I updated everything per your review. Can you do a final review (but don't merge...I still want to squash that last commit in if it's good).

pycloudlib/lxd/instance.py
blackboxsw: ahh right +1

blackboxsw:
u> > as that operation will be exercising different platform logic relative to the intsances under test.
> 
> I'm fine removing the behavior, but I'm not sure I understand this reasoning. Why do we care about exercising different platform logic? From the perspective of cloud-init, I care about what happens while the instance is booting. Whether that happened due to a restart signal vs a shutdown/start doesn't really matter to me. Should it? I'm assuming it's the same for UA, but maybe there are use cases I'm not understanding.


Thanks @TheRealFalcon, this where I'm coming from, and maybe it just equates to us over-engineering something that is not needed. I'm happy to be dissauded or steered better.

Here are three ideas causing tension for me in priority order:
 1. the need for pycloudlib to make an opinionated decision about how to "best" perform a cloud operation as suited for the test need
 2. the need to generalize pycloudlib cloud and instance APIs to make reusing the same test against multiple clouds easy
 3. the ability to trigger discrete cloud operations if the platform does something that breaks cloud-init/UA/image creation etc.
 4. desire to also represent a platform's operations as closely to the cloud's API/SDK as possible to ease adoption for cloud-specific test use cases

Are those 4 goals of pycloudlib correct/reasonable to the current project?

#3 is the reason I made the comment about not exercising specific platform logic for restart vs stop/start. We don't have any testcases now that explicitly need a restart as separate from stop/start , so maybe it's something we only address when we have a need.  If/when a case does come up where the platform logic for a discrete restart vs stop/start on CloudX causes cloud-init/UA/networking to fall over, we'll then need to instrument a mechanism in pycloudlib to call that operation vs the start/stop.

Since we have no need for a explicity `restart` vs `stop/start`,  I'm good with pycloudlib generalizing restart to be an alias for start/stop.

That said, I think we also need to keep wary of #1 for each cloud/platform (as we have been) and we should use a cloud's specific discrete `restart` operation by default if it makes it easier, faster and/or more stable.
I think in terms of lxc, `lxc restart` will make for the most efficient/correct restart operation for that platform, in Ec2, this may be different and I think we should address that separately if needed and we determine that our current approach is suffering from stability issues.

TheRealFalcon:
Yeah, those goals make sense. That helps a lot. Thanks for explaining your thoughts. It sounds like there's room for practicality in dealing with each cloud.

> Since we have no need for a explicity restart vs stop/start, I'm good with pycloudlib generalizing restart to be an alias for start/stop. 

Well, I already removed the commit so let's keep it as-is.  :joy:

If we see further connection issues after this change, we can always change things. There are already a few clouds where we use stop/start for the restart case. We can keep it that way unless we see a strong reason not to. A generic form of the EC2 reboot tracking may also make sense (if we can figure out why it doesn't always work).

blackboxsw (commented):
Will one of the commits reference https://github.com/canonical/pycloudlib/issues/110

TheRealFalcon:
https://github.com/canonical/pycloudlib/pull/148/commits/b94253cb7d8e121c5e75ff86acdbd838a7c5c1bf fixes it. Should I add it to the commit message too?

blackboxsw (approved):
+1. Finally ran integration tests with this changeset without degradation of performance. Had wanted to spot check LXC integration test behavior to confirm it didn't affect stability of runs.

PR #149, opened 2021-06-04
Run Pycloudlib CI GH action only on pull_request

TheRealFalcon:
Specifying `pull_request` will still build every time there is a new push to the pull request. When we specify `push`, we get double builds for every PR.


paride (approved):
+1, thanks!

PR #150, opened 2021-06-10
update lxd xenial profile to not install agent

lucasmoura:
On xenial we don't need the lxd-agent anymore, however we do need additional cloud-init config to be added into the profile. We are not splitting the cloud-init behavior from lxd-agent one

### DIFF
--- a/pycloudlib/lxd/defaults.py
+++ b/pycloudlib/lxd/defaults.py
@@ -3,7 +3,7 @@
 import textwrap
 
 
-LXC_PROFILE_VERSION = "v2"
+LXC_PROFILE_VERSION = "v3"
 
 
 # For Xenial and Bionic vendor-data required to setup lxd-agent in a vm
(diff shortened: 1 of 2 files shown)


pycloudlib/lxd/defaults.py
```diff
@@ -47,25 +47,34 @@
...
     config_device = ""
     vendordata = "config: {}"
-    if install_agent:
+    if config_cloudinit:
         # We need to mount the config drive so that cloud-init finds the
         # vendor-data instructing it to install the agent
         config_device = "config: {source: cloud-init:config, type: disk}"
+    if install_agent:
+        # We need to mount the config drive so that cloud-init finds the
+        # vendor-data instructing it to install the agent
```
orndorffgrant, line 60: Looks like this comment got copy/pasted and could use updating or removal

orndorffgrant (approved):
This worked for me! I was able to run xenial vm tests. One tiny comment change requested and then good to go!

PR #151, opened 2021-06-11
Don't run wait_for_instance_start on xenial LXD vms

lucasmoura:
On xenial LXD vms, we don't install the lxd-agent. Therefore we cannot properly run the _wait_for_instance_start method. To fix that, we will not run that method on xenial LXD vms


orndorffgrant (approved):
This fixed the xenial vm issue for me when running uaclient behave tests

PR #152, opened 2021-06-11
Sync OCI on both shutdown and restart (SC-120)

TheRealFalcon:
Whoops, when I removed the "start/stop" commit on the timeout PR, I forgot to update OCI to also sync before restart.

Also updated the description to be slightly less aggressive :joy:


blackboxsw (approved):
whoopsie

PR #153, opened 2021-06-11
lxd.vm: check lxd ls state == RUNNING on xenial

blackboxsw:
Xenial VMs do not have a lxd-agent, we can't see process
count for VMs.

Fallback to check lxc ls state as "RUNNING" on Xenial VMs.

Add LXDInstance.wait_for_state() method and adapt lxc.wait_for_stop
to use that shared method.

# to test 
python3  examples/lxd.py  # watch for DEBUG:pycloudlib.instance:waiting for (STOPPED|RUNNING): pycloudlib-vm  logs

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -377,19 +377,29 @@ class LXDInstance(BaseInstance):
         Not used for LXD.
         """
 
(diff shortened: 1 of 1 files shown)


lucasmoura (approved):
Confirmed that `wait_for_state` is working as expected

PR #155, opened 2021-07-01
ec2: Increase console wait time to 5 minutes

TheRealFalcon:
We're still getting failures on ec2 due to console logs not being available. E.g., https://jenkins.ubuntu.com/server/view/cii-pytest/job/cloud-init-integration-ec2-focal-daily/5/console

I want to try upping the wait time to 5 minutes, and if that doesn't work, restricting the test to only run on LXD until we can fix what's wrong here.


paride (approved):
LGTM! Good to merge if CI agrees.

PR #156, opened 2021-07-06
Allow hotplug of NIC on Openstack and EC2 (SC-254)

TheRealFalcon:
Needed for integration testing https://github.com/canonical/cloud-init/pull/936

I expect there may some change in the base interface as new clouds are added.

### DIFF
--- a/pycloudlib/ec2/instance.py
+++ b/pycloudlib/ec2/instance.py
@@ -1,9 +1,9 @@
 # This file is part of pycloudlib. See LICENSE file for license information.
 """EC2 instance."""
-
 import string
 import time
 
+import botocore
 from paramiko.ssh_exception import (
     SSHException
 )
@@ -67,11 +67,12 @@ class EC2Instance(BaseInstance):
         """Return id of instance."""
         return self._instance.image_id
 
-    def add_network_interface(self):
+    def add_network_interface(self) -> str:
         """Add network interface to instance.
 
         Creates an ENI device and attaches it to the running instance. This
-        is effectively a hot-add of a network device.
+        is effectively a hot-add of a network device. Returns the IP address
+        of the added network interface as a string.
 
         See the AWS documentation for more info:
         https://boto3.readthedocs.io/en/latest/reference/services/ec2.html?#EC2.Client.create_network_interface
@@ -79,7 +80,7 @@ class EC2Instance(BaseInstance):
         """
         self._log.debug('adding network interface to %s', self.id)
         interface_id = self._create_network_interface()
-        self._attach_network_interface(interface_id)
+        return self._attach_network_interface(interface_id)
 
     def add_volume(self, size=8, drive_type='gp2'):
         """Add storage volume to instance.
@@ -264,7 +265,7 @@ class EC2Instance(BaseInstance):
             }]
         )
 
-    def _attach_network_interface(self, interface_id):
+    def _attach_network_interface(self, interface_id: str) -> str:
         """Attach ENI device to an instance.
 
         This will attach the interface at the next available index.
@@ -274,6 +275,8 @@ class EC2Instance(BaseInstance):
 
         Args:
             interface_id: string, id of interface to attach
+        Returns:
+            IP address of the added interface
         """
         device_index = self._get_free_nic_index()
         args = {
@@ -294,6 +297,9 @@ class EC2Instance(BaseInstance):
                         'DeleteOnTermination': True
                     }
                 )
+                return nic.private_ip_address
+        raise Exception('Could not attach NIC with AttachmentId: {}'.format(
+            response.get('AttachmentId', None)))
 
     def _create_ebs_volume(self, size, drive_type):
         """Create EBS volume.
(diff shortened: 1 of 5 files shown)


pycloudlib/ec2/instance.py
```diff
@@ -294,6 +294,9 @@ def _attach_network_interface(self, interface_id):
                         'DeleteOnTermination': True
                     }
                 )
+                return nic.private_ip_address
+        raise Exception('Could not nic with AttachmentId: {}'.format(
```
blackboxsw, line 298: mino typo
```suggestion
        raise Exception('Could not attach NIC with AttachmentId: {}'.format(
```

pycloudlib/ec2/instance.py
```diff
@@ -67,7 +67,7 @@ def image_id(self):
         """Return id of instance."""
         return self._instance.image_id
 
-    def add_network_interface(self):
+    def add_network_interface(self) -> str:
```
blackboxsw, line 70: Let's also type hint _attach_network_interface and document the expected return of a private_ip_address?

pycloudlib/ec2/instance.py
```diff
@@ -294,6 +294,9 @@ def _attach_network_interface(self, interface_id):
                         'DeleteOnTermination': True
                     }
                 )
+                return nic.private_ip_address
+        raise Exception('Could not nic with AttachmentId: {}'.format(
+            response['AttachmentId']))
```
blackboxsw, line 299: Have we seen the failure case here? I'm wondering if 'AttachmentId' may not exist in failure cases and if we need to safe guard that assumption with a `.get`

pycloudlib/openstack/instance.py
```diff
@@ -156,3 +164,40 @@ def wait_for_delete(self):
...
+        ports = self.conn.network.ports()
+        for port in ports:
+            for ip in port['fixed_ips']:
+                if ip['ip_address'] == ip_address:
+                    return port
+        raise Exception('Could not find port with IP: {}'.format(ip_address))
+
+    def remove_network_interface(self, ip_address: str):
+        """Remove nic from running instance."""
+        port = self._get_port_id_by_ip(ip_address)
```
blackboxsw, line 194: We aren't handling the potential Exception here if invalid IP is provided. Don't we want to handle this Exception by logging a warning as below or just let it break the CI run?

pycloudlib/openstack/instance.py
```diff
@@ -39,6 +41,7 @@ def __init__(self, key_pair, instance_id, connection=None):
         if self.floating_ip is None:
             self.floating_ip = self._create_and_attach_floating_id()
             self.delete_floating_ip = True
+        self.added_local_ports = []
```
blackboxsw, line 44: Do we eventually want visibility to the ports attached to the running instance via `compute.server_interfaces` so that other pre-existing ports could be removed via pycloudlib (in the event of an instance created and referenced via 
 pycloudlib's `Openstack.get_instance`?

blackboxsw (changes requested):
Thanks @TheRealFalcon couple of comments. Take what you will

pycloudlib/openstack/instance.py
TheRealFalcon: I don't think so. If we're trying to remove a NIC and fail, I think we have a problem that needs to be dealt with. I used a warning further down because that's our own accounting of the ports and likely just a harmless bug on our end if it's not there.

pycloudlib/openstack/instance.py
TheRealFalcon: I'm not entirely sure what you're asking here.

pycloudlib/openstack/cloud.py
```diff
@@ -78,8 +78,12 @@ def get_instance(self, instance_id) -> OpenstackInstance:
         return OpenstackInstance(
             key_pair=self.key_pair,
             instance_id=instance_id,
+            network_id=self._get_network_id()
         )
 
+    def _get_network_id(self):
+        return self.conn.network.find_network(self.network).id
```
lucasmoura, line 85: Can `find_network` fail ? And if it can, we could have a better error message here instead of probably `NoneType has no attribute named id`

pycloudlib/openstack/instance.py
```diff
@@ -156,3 +164,40 @@ def wait_for_delete(self):
...
+        port = self.conn.network.create_port(
+            network_id=self.network_id,
+            node_id=self.server.id
+        )
+        self.added_local_ports.append(port.id)
+        interface = self.conn.compute.create_server_interface(
+            server=self.server.id,
+            port_id=port.id
+        )
+        return interface['fixed_ips'][0]['ip_address']
```
lucasmoura, line 182: Same thing here regarding logging a better error message, if the error can happen of course

TheRealFalcon: If the line above fails in some way, it should raise it's own error. If it succeeds, then these fields should all be present. If they're not present for some reason, then we'll need to dive in and debug why not, and I don't think any additional exception wrapping would be advantageous in that case.

TheRealFalcon:
@lucasmoura , I addressed one of your comments, and I believe the other one is ok as-is.

lucasmoura (approved):

PR #157, opened 2021-07-13
lxd: use shutdown to delete ephemeral instances

lucasmoura:
Ephemeral instances are properly deleted by lxd when they are stopped. Because of that we are now shutting down ephemeral instances when a user calls delete on them.

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -130,7 +130,7 @@ class LXDInstance(BaseInstance):
         except IndexError:
             return False
 
-        return bool(info_type == 'ephemeral')
+        return bool("ephemeral" in info_type)
 
     @property
     def state(self):
@@ -180,7 +180,13 @@ class LXDInstance(BaseInstance):
             wait: wait for delete
         """
         self._log.debug('deleting %s', self.name)
-        subp(['lxc', 'delete', self.name, '--force'])
+
+        if self.ephemeral:
+            # We don't need to wait here, since the
+            # instance will be deleted once it is stopped
+            self.shutdown(wait=False)
+        else:
+            subp(['lxc', 'delete', self.name, '--force'])
 
         if wait:
             self.wait_for_delete()
--- a/pycloudlib/lxd/tests/test_instance.py
+++ b/pycloudlib/lxd/tests/test_instance.py
@@ -188,3 +188,35 @@ class TestIP:
             assert expected == instance.ip
             assert [lxc_mock] * (1 + sleeps) == m_subp.call_args_list
         assert sleeps == m_sleep.call_count
+
+
+class TestDelete:
+    """Tests covering pycloudlib.lxd.instance.Instance.delete."""
+
+    @pytest.mark.parametrize(
+        "is_ephemeral", ((True), (False))
+    )
+    @mock.patch("pycloudlib.lxd.instance.LXDInstance.shutdown")
+    @mock.patch("pycloudlib.lxd.instance.subp")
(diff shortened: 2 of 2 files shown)


renanrodrigo:
lgtm

blackboxsw (changes requested):
It looks like we have a bug, not in this PR, but in ephemeral detection not properly discovering ephemeral type instances. We need to fix that in this branch if we are going to rely on it.
Basically Type from lxc info is probably either "container (ephemeral)" or "virtual-machine (ephemeral)", maybe our match algorithm should just look for the string "ephemeral" in info_type to avoid strict matching of strings that might change (or new instance types in the future).



To reproduce failure:

I can apply the following diff[1]
and then run
```bash
tox -e py3
.tox/py3/bin/python3 examples/lxd.py
.... OOPS doesn't look ephemeral to me
```



References: 
[1] diff: 
```diff
diff --git a/examples/lxd.py b/examples/lxd.py
index 5b480cf..cda2de6 100755
--- a/examples/lxd.py
+++ b/examples/lxd.py
@@ -98,53 +98,14 @@ def launch_options():
     Finally, an instance with custom configurations options.
     """
     lxd = pycloudlib.LXDContainer('example-launch')
-    kvm_profile = textwrap.dedent(
-        """\
-        devices:
-          kvm:
-            path: /dev/kvm
-            type: unix-char
-        """
-    )
-
-    lxd.create_profile(
-        profile_name="kvm",
-        profile_config=kvm_profile
-    )
-
-    lxd.launch(
-        name='pycloudlib-kvm', image_id=RELEASE,
-        profile_list=['default', 'kvm']
-    )
-    lxd.delete_instance('pycloudlib-kvm')
-
-    lxd.launch(
+    inst = lxd.launch(
         name='pycloudlib-ephemeral',
         image_id='ubuntu:%s' % RELEASE,
         ephemeral=True
     )
+    assert inst.ephemeral, "OOPS doesn't look ephemeral to me"
     lxd.delete_instance('pycloudlib-ephemeral')
 
-    lxd.launch(
-        name='pycloudlib-custom-hw',
-        image_id='images:ubuntu/xenial',
-        network='lxdbr0',
-        storage='default',
-        inst_type='t2.micro',
-        wait=False
-    )
-    lxd.delete_instance('pycloudlib-custom-hw')
-
-    lxd.launch(
-        name='pycloudlib-privileged',
-        image_id=RELEASE,
-        config_dict={
-            'security.nesting': 'true',
-            'security.privileged': 'true'
-        }
-    )
-    lxd.delete_instance('pycloudlib-privileged')
-
 
 def basic_lifecycle():
     """Demonstrate basic set of lifecycle operations with LXD."""
@@ -219,12 +180,12 @@ def launch_virtual_machine():
 
 def demo():
     """Show examples of using the LXD library."""
-    basic_lifecycle()
+  #  basic_lifecycle()
     launch_options()
-    launch_multiple()
-    modify_instance()
-    snapshot_instance()
-    launch_virtual_machine()
+   # launch_multiple()
+    #modify_instance()
+  # snapshot_instance()
+  # launch_virtual_machine()
 
 
 if __name__ == '__main__':
```

lucasmoura:
@blackboxsw good catch. I have updated the code to address it

blackboxsw (approved):
+1 thanks. delete now works.  I've put up a separate https://github.com/canonical/pycloudlib/pull/158 for `LXDInstance.shutdown` which hits a timeout on ephemeral instances because it tries to wait until LXC reports the instance in stopped state, but what it should expect is that the instance disappears in this case.

PR #158, opened 2021-07-13
Lxd shutdown no wait

blackboxsw:
lxd: ephemeral instance shutdown should not wait for STOPPED state

Since lxd will destroy an ephemeral instance after shutdown, don't
call wait_for_stop() as it'll timeout on inability to find any
instance information from `lxc info <name>`

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -21,19 +21,31 @@ class LXDInstance(BaseInstance):
     _is_ephemeral = None
 
     def __init__(
-        self, name, key_pair=None, execute_via_ssh=True, series=None
+        self,
+        name,
+        key_pair=None,
+        execute_via_ssh=True,
+        series=None,
+        ephemeral=None,
     ):
         """Set up instance.
 
         Args:
             name: name of instance
             key_pair: SSH key object
+            execute_via_ssh: Boolean True to use ssh instead of lxc exec for
+                all operations.
+            series: Ubuntu release name: xenial, bionic etc.
+            ephemeral: Boolean True if instance is ephemeral. If left
+                unspecified, ephemeral type will be determined and cached by
+                the ephemeral method.
         """
         super().__init__(key_pair=key_pair)
 
         self._name = name
         self.execute_via_ssh = execute_via_ssh
         self.series = series
+        self._is_ephemeral = ephemeral
 
     def __repr__(self):
         """Create string representation for class."""
@@ -128,14 +140,14 @@ class LXDInstance(BaseInstance):
 
             try:
                 info_type = re.findall(r'Type: (.*)', result)[0]
+                self._is_ephemeral = bool("ephemeral" in info_type)
             except IndexError:
                 self._log.debug(
                     'Unable to parse lxc show %s to determine ephemeral type.'
                     ' Assuming not ephemeral.',
                     self.name
                 )
-                return False
-        self._is_ephemeral = bool("ephemeral" in info_type)
+                self._is_ephemeral = False
         return self._is_ephemeral
 
     @property
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -18,6 +18,7 @@ class LXDInstance(BaseInstance):
 
     _type = 'lxd'
     _is_vm = None
+    _is_ephemeral = None
 
     def __init__(
         self, name, key_pair=None, execute_via_ssh=True, series=None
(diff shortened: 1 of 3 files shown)


pycloudlib/lxd/instance.py
```diff
@@ -318,7 +318,7 @@ def shutdown(self, wait=True, force=False, **kwargs):
 
         subp(cmd)
 
-        if wait:
+        if wait and not self.ephemeral:
```
paride, line 321: This LGTM, but maybe it would be better to put this login inside `wait_for_stop()` https://github.com/canonical/pycloudlib/blob/0869d07b7894d985e183be957ae8318aae667c00/pycloudlib/lxd/instance.py#L406 just in case there are other places where we call it. Wdyt?

blackboxsw: @paride your suggestion makes sense. I'm thinking a bit more about this self.ephemeral [relies on calling 'lxc info self.name'](https://github.com/canonical/pycloudlib/tree/main/pycloudlib/lxd/instance.py which may return nothing if the lxc container has already stopped because lxc info will not find the container anymore if it already stopped and was deleted. Since a container won't/can't migrate from ephemeral  TO "not ephemeral". I'm thinking instead that this state could just be preserved on the instance itself at launch time. No need to shell out to lxc info to get that info. I'll adapt this branch with your suggestion and an ephemeral "fix" to see what you think.

paride:
CI spotted:
```
  File "examples/lxd.py", line 232, in <module>
    demo()
  File "examples/lxd.py", line 222, in demo
    basic_lifecycle()
  File "examples/lxd.py", line 171, in basic_lifecycle
    print(inst.ephemeral)
  File "/home/travis/build/canonical/pycloudlib/pycloudlib/lxd/instance.py", line 138, in ephemeral
    self._is_ephemeral = bool("ephemeral" in info_type)
UnboundLocalError: local variable 'info_type' referenced before assignment
```
however IIUC you'll drop the `lxd info` shell out, retailing the ephemeral status from the beginning.

pycloudlib/lxd/instance.py
paride: I think retaining the ephemeral status makes sense, given that it can't change. However it should probably be retained from the `lxc init` stage and not from `lxc launch`, as initialized-but-not-launched instanced already have the ephemeral flag set.

blackboxsw:
> CI spotted:
> 
> ```
> Traceback (most recent call last):
>   File "examples/lxd.py", line 232, in <module>
>     demo()
>   File "examples/lxd.py", line 222, in demo
>     basic_lifecycle()
>   File "examples/lxd.py", line 171, in basic_lifecycle
>     print(inst.ephemeral)
>   File "/home/travis/build/canonical/pycloudlib/pycloudlib/lxd/instance.py", line 138, in ephemeral
>     self._is_ephemeral = bool("ephemeral" in info_type)
> UnboundLocalError: local variable 'info_type' referenced before assignment
> ```
> 
> however IIUC you'll drop the `lxd info` shell out, retailing the ephemeral status from the beginning.

blackboxsw:
> > CI spotted:
> > ```
> > Traceback (most recent call last):
> >   File "examples/lxd.py", line 232, in <module>
> >     demo()
> >   File "examples/lxd.py", line 222, in demo
> >     basic_lifecycle()
> >   File "examples/lxd.py", line 171, in basic_lifecycle
> >     print(inst.ephemeral)
> >   File "/home/travis/build/canonical/pycloudlib/pycloudlib/lxd/instance.py", line 138, in ephemeral
> >     self._is_ephemeral = bool("ephemeral" in info_type)
> > UnboundLocalError: local variable 'info_type' referenced before assignment
> > ```


Thanks, resolved the CI issue, setting info_type also when we can't parse Type: from lxc info

pycloudlib/lxd/instance.py
blackboxsw: On further inspection we can't actually from the "lxc info <name>" logic because that data is not available to [LXD*(Cloud).get_instances](https://github.com/canonical/pycloudlib/tree/main/pycloudlib/lxd/cloud.py#L88-L98). In this case, via a cloud.get_instance call, the instance itself will still have to determine this at runtime.


```bash
$ lxc launch ubuntu-daily:bionic --ephemeral my-ephemeral
$ python3 -c 'import pycloudlib;   lxd = pycloudlib.LXDContainer('test-ephemeral'); instance = lxd.get_instance("my-ephemeral"); print("I am ephemeral: {}".format(instance.ephemeral))'
I am ephemeral True
```
```

lucasmoura (approved):

PR #160, opened 2021-08-11
Allow snapshot for LXD ephemeral instances (SC-321)

### LINKED ISSUE #159: LXD snapshot(): do not shutdown() and use `lxc publish --force`
Stemmed out of https://github.com/canonical/ubuntu-advantage-client/pull/1744#issuecomment-884433552

From my local testing ephemeral instances survive the stop issued by `lxc publish --force`. Also the instance is brought back to it previous RUNNING state after the snapshot is taken, I can't see code doing it in

https://github.com/canonical/pycloudlib/blob/dd27b604137cc5a20fb54e233ecb6a028b3d6891/pycloudlib/lxd/instance.py#L362

and I think it's desirable. My proposal is to remove the `self.shutdown(wait=True)` from `snapshot()` for LXD instances and always use `--force`.

lucasmoura:
During the snapshot call, we shutdown the base instance before creating the snapshot. This approach doesn't work for ephemeral instances, because they get deleted after the shutdown operation. We are now updating the snapshot function to allow ephemeral instances to be snapshot.

Fixes #159

### DIFF
--- a/examples/lxd.py
+++ b/examples/lxd.py
@@ -34,6 +34,33 @@ def snapshot_instance():
     inst.delete(wait=False)
 
 
+def image_snapshot_instance(ephemeral_instance=False):
+    """Demonstrate image snapshot functionality.
+
+    Create an snapshot image from a running instance an show
+    how to launch a new instance based of this image snapshot
+    """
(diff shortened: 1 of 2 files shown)


paride (approved):
LGTM, thanks!

PR #161, opened 2021-08-30
Sru issues (SC-373)

TheRealFalcon:
Unrelated issues found in latest cloud-init SRU. See individual commits for specific details. Don't squash merge.

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -371,6 +371,8 @@ class LXDInstance(BaseInstance):
             snapshot_name: name to call snapshot
         """
         self.clean()
+        if not self.ephemeral:
+            self.shutdown()
         if snapshot_name is None:
             snapshot_name = '{}-snapshot'.format(self.name)
         cmd = [
--- a/pycloudlib/openstack/instance.py
+++ b/pycloudlib/openstack/instance.py
@@ -91,7 +91,7 @@ class OpenstackInstance(BaseInstance):
         while time.time() < start + 180:
             response = self.conn.compute.get_server_console_output(self.server)
             if response:
-                return response
+                return response['output']
             self._log.debug("Console output not yet available; sleeping")
             time.sleep(5)
         return 'No console output'
--- a/pycloudlib/gce/instance.py
+++ b/pycloudlib/gce/instance.py
@@ -138,15 +138,17 @@ class GceInstance(BaseInstance):
 
     def wait_for_delete(self, sleep_seconds=300):
         """Wait for instance to be deleted."""
-        # Once instance is deleted, URL just 404s
         for _ in range(sleep_seconds):
             try:
-                self.instance.get(
+                result = self.instance.get(
                     project=self.project,
                     zone=self.zone,
                     instance=self.instance_id
                 ).execute()
+                if result['status'] == 'TERMINATED':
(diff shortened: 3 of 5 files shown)


pycloudlib/openstack/instance.py
```diff
@@ -91,7 +91,7 @@ def console_log(self):
         while time.time() < start + 180:
             response = self.conn.compute.get_server_console_output(self.server)
             if response:
-                return response
+                return response['output']
```
blackboxsw, line 94: Looks good had to validate what python openstack python client returned (as their get_and_munchify() function in /usr/lib/python3/dist-packages/openstack/cloud/meta.py led me to believe that the top-level "output" key wasn't going to be present)
```bash
(Pdb) inst.console_log.keys()
dict_keys(['output'])
```

blackboxsw (approved):
LGTM. Will await a ping if you have additional commits to add, or feel free to merge as is.

pycloudlib/lxd/instance.py
```diff
@@ -371,6 +371,7 @@ def snapshot(self, snapshot_name):
             snapshot_name: name to call snapshot
         """
         self.clean()
+        self.shutdown()
```
blackboxsw, line 374: Hrm, this might be a prob for ephemeral instances right because the instance will disappear after stop? Maybe we don't want to call this when ephemeral VM?
```suggestion
        if not self._is_ephemeral:
            self.shutdown()
```

blackboxsw (commented):
Looks good, though I have once concern about the impact of LXDInstance.shutdown on ephemeral vms and would that still work?

pycloudlib/lxd/instance.py
TheRealFalcon: The change is fine, but on UA, do you create snapshots like we do on cloud-init (note that I mean snapshot generally, not the lxd snapshot command)? Can you even take snapshots of an ephemeral VM?

pycloudlib/ec2/instance.py
```diff
@@ -111,13 +111,18 @@ def console_log(self):
 
         """
         start = time.time()
+        output = ''
         while time.time() < start + 300:
             response = self._instance.console_output()
             try:
-                return response['Output']
+                output += response['Output']
```
blackboxsw, line 118: @TheRealFalcon This logic now has the effect of looping every 5 seconds for 5 minutes and adding all previous repeated full console output to the boot. The end result that you will always wait 5 mins on `ec2.console_log` call and then get about 60 copies of the full console log from the vm. Is it worth instead just adding a wait_for_cloud_init=False boolean param for cases where we need to wait cloud-init to complete before checking the logs?

blackboxsw: An alternative could be that we just call instance.wait() in integration tests before calling instance.console_log()

TheRealFalcon: Bah, thanks. There's something in the back of my mind thinking that we consume the log. Not sure if that's because of another cloud or I just made it up in my head.

The test will need to be updated to poll multiple times then. We are already calling `.wait()` before we start our tests, and `.restart()` also calls `.wait()`. I can back out this commit.

TheRealFalcon:
I dropped that last commit. I'll update the test accordingly.

PR #162, opened 2021-10-04
gce: fallback to GOOGLE_CLOUD_PROJECT if project is not specified

paride:
When initializing GCE fallback to the GOOGLE_APPLICATION_CREDENTIALS and
GOOGLE_CLOUD_PROJECT environment variables if credentials_path or
project are not specified.


paride:
@lucasmoura I think I understand your comment (thanks!), but that only applies to `credentials_path`, while `project` gets used.

paride:
PR updated and tested.

lucasmoura (approved):

paride:
Thanks!

PR #163, opened 2021-10-17
Ensure instance is active before cloud-init status --wait (SC-509)

TheRealFalcon:


pycloudlib/instance.py
```diff
@@ -396,5 +396,15 @@ def _wait_for_execute(self):
     def _wait_for_cloudinit(self):
         """Wait until cloud-init has finished."""
         self._log.debug('_wait_for_cloudinit to complete')
+        if self.execute(['which', 'systemctl']).ok:
+            # We may have issues with cloud-init status early boot, so also
+            # ensure our cloud-init.target is active as an extra layer of
+            # protection against connecting before the system is ready
+            for _ in range(300):
+                if self.execute([
+                    'systemctl', 'is-active', 'cloud-init.target'
```
blackboxsw, line 405: As written, this means any integration test expecting to fail cloud-init will likely loop for 5 mins waiting on a state that'll never come.

I think this is OK/expected behavior and in those cases the integration test in question would need launch the instance with wait=False which would avoid this loop. And implement their own cloud-init --wait calls etc.

+1 if you agree.

TheRealFalcon: I think this is ok. This is called immediately after `_wait_for_execute()` which itself already has a 10 minute wait. After 10 minutes, what's 5 minutes more? :wink: 

But in all seriousness, the target will be active even if cloud-init services fail so realistically this shouldn't add any additional time that wasn't already spent in `cloud-init status --wait`.

PR #164, opened 2021-10-22
Add pycloudlib configuration file (SC-519)

TheRealFalcon:
```
Add a pycloudlib configuration file to obtain cloud-specific
credentials rather than embedding credential details in the API.

This is currently be backwards compatible, so credentials can still be
passed via API, and those passed via the API will take precedence
over credentials specified in the configuration file.
```

This allows us to pull individual cloud configuration out of the API. There's a few reasons I want to do this:
1. There have often been "works on my machine" problems when it comes to credentials. One person might have a newer version of one of the cli tools that expects things in different locations. Moving credentials into a separate configuration file (ideally) means there's no more wondering what needs to be specified to launch an instance on particular cloud.
2. Pycloudlib currently needs to be configured per project. E.g., UA tests are already work on certain jenkins machines, and re-setting things up so they work with cloud-init is a waste of time.
3. It's easier to follow the [liskov substitution principle](https://stackify.com/solid-design-liskov-substitution-principle/#:~:text=The%20Liskov%20Substitution%20Principle%20in,the%20objects%20of%20your%20superclass). E.g., openstack currently requires a 'network' parameter that other clouds don't require. This means we can't make a generic `get_cloud().launch()` kind of API, because different clouds require different arguments based on their auth.
4. I'd eventually like to build [a cli](https://gist.github.com/TheRealFalcon/9488910f1d32082ba7ad853674fc958b) (how I'm currently testing this branch) on top of pycloudlib, and this makes it a whole lot easier.

I haven't updated any tests or documentation as I would like to get buy-in on what's here first. I also realize that this config references other configs. Most of them can be pulled out into a single config, but we can iterate on that later, and I'd rather not make this PR even bigger.


TheRealFalcon:
45 files changed!? Over half of those are docs changes and removal of a directory we don't need (`make build` from the docs directory if you don't believe me). I intend to keep that as a separate commit.

There were a lot of small required test changes, so a lot of files got touched, but the non-test code changes are almost identical to the first round of reviews. I tested using the config file on all of the pycloudlib supported clouds.

blackboxsw (commented):
This looks really good, and I think a strict approach to defining a single conf entrypoint makes the most sense and avoids some of pitfalls we've seen with trying to be "too flexible" in supporting each cloud's means of obtaining credentials. One conf to rule them all, and proper pycloudlib docs to allow pycloudlib users to query the cloud to get those details where possible.

blackboxsw (commented):
Now that we have a potential of people providing unaltered or incorrect toml config with empty string as values. I think we probably want to handle "empty" values and ignore them properly on each cloud.

Ec2 currently fails in an unfriendly manner because pycloudlib.ec2.util _get_session passes the empty strings for secet_access_key, access_key_id and region into boto3.Session, which attempts to use the empty string instead of sourcing your local .aws/config files. 
```bash
    self.client = session.client('ec2')
  File "/home/csmith/src/pycloudlib/.tox/pylint/lib/python3.8/site-packages/boto3/session.py", line 258, in client
    return self._session.create_client(
  File "/home/csmith/src/pycloudlib/.tox/pylint/lib/python3.8/site-packages/botocore/session.py", line 847, in create_client
    client = client_creator.create_client(
  File "/home/csmith/src/pycloudlib/.tox/pylint/lib/python3.8/site-packages/botocore/client.py", line 86, in create_client
    client_args = self._get_client_args(
  File "/home/csmith/src/pycloudlib/.tox/pylint/lib/python3.8/site-packages/botocore/client.py", line 355, in _get_client_args
    return args_creator.get_client_args(
  File "/home/csmith/src/pycloudlib/.tox/pylint/lib/python3.8/site-packages/botocore/args.py", line 99, in get_client_args
    endpoint = endpoint_creator.create_endpoint(
  File "/home/csmith/src/pycloudlib/.tox/pylint/lib/python3.8/site-packages/botocore/endpoint.py", line 287, in create_endpoint
    raise ValueError("Invalid endpoint: %s" % endpoint_url)
ValueError: Invalid endpoint: https://ec2..amazonaws.com
csmith@uptown:~/src/pycloudlib$ 
```

Should we treat toml empty string values as None in pycloudlib before invoking cloud-specific libraries? Or should we just ensure ec2, gce, openstack oci adopt the right fallback when provided with an empty string value.

TheRealFalcon:
Re: empty or bad value checks, I think I'd like to wait and iterate on that. Especially if we decide that configuration file will be the _only_ valid way to pass credentials to pycloudlib, it will be a lot easier to do validation of required values.

paride (approved):
I finally managed to try test more things and what I tried worked smoothly. There is probably room for improvement in how some errors/failures are presented, but that only happens if the conf file is broken. We can have more iterations on this if needed, no need to block landing. Overall this is a very big improvement on the usability of the library!

PR #171, opened 2021-12-09
update tox.ini (SC-674)

TheRealFalcon:
Includes:
* Docs build
* Share testenvs when reasonable
* Share common config when reasonable
* Bump linting versions
* Rename py3 env to pytest
* Remove nose section
* Remove default recreate=true

Also update main makefile clean to also clean docs dir and print docs
make commands to stdout.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ clean:
 	$(SETUP) clean
 	rm -rf .tox .eggs *.egg-info build dist venv
 	@find . -regex '.*\(__pycache__\|\.py[co]\)' -delete
+	$(MAKE) -C docs clean
 
 install:
 	$(SETUP) install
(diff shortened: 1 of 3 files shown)


paride:
All nice changes, thanks! I can't spot right away why tox itself is failing in CI. Maybe Bionic's tox is too old? The rationale for testing in Bionic is that it's the oldest Ubuntu release we can reasonably want to use pycloudlib on.)

paride:
I can reproduce the failure on Bionic.

TheRealFalcon:
Yeah, looks like the bionic version uses a tox from 2015 :astonished: . Since we don't package pycloudlib into any OS repos, I think I'm ok using the pip version of tox here. I'm going to play with that unless you think we shoudln't.

paride (approved):
LGTM, thanks!

PR #173, opened 2021-12-13
tox: bump pylint to 2.12.2

paride:
This solves a E1136 false positive with Python 3.9 [1], but the code
needs some changes to test clean.

[1] https://github.com/PyCQA/pylint/issues/3882

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -11,7 +11,9 @@ jobs=4
 # C0122: misplaced-comparison-constant
 # C0301: line-too-long  (flake8 already catches this for us)
 # R0201: Method could be a function (no-self-use)
-disable=R0902, R0903, R0913, W0221, C0103, C0122, C0301, R0201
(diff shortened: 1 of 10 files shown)


paride:
Let me know if you prefer me to batch-fix C0209 (consider-using-f-string) instead of silencing it.

TheRealFalcon (approved):
I would like us to upgrade all of our string formatting to f-strings eventually. They're faster and cleaner than the other formatting options. I'm also aware of [a tool](https://github.com/ikamensh/flynt) that should make it easier when we do decide to do it.

This can (and probably should) be a separate PR though, so this PR Is good as-is.

paride:
https://github.com/ikamensh/flynt: nice! But I gave it a blind shot and it didn't detect/convert 100% of the format strings (a number of C0209 remain for some reason), and there are some "E501 line too long" to fix. Maybe worth a separate PR as you say.

PR #174, opened 2021-12-14
Black (SC-695)

TheRealFalcon:
```
Black

Apply black formatting, and update tox.ini, pyproject.toml, and docs
accordingly.
```

The configuration/docs changes are a separate commit. We might want to remove other tools in the future (i.e., flake8), but I'm leaving these in place for now.

Don't squash-merge.


TheRealFalcon:
I looked into using a github action to automatically apply black to a PR and add a commit on top if necessary, but that requires the use of github tokens, which won't really work with forked repositories.

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+line-length = 79
```
TheRealFalcon, line 2: I'd like to remove this restriction and change to the 88 black default (or even bump up to 100 (that'll be a fun post-standup)), but for now, lets minimize the change surface and keep the current restriction.

tox.ini
```diff
@@ -1,5 +1,6 @@
 [tox]
-envlist = flake8, pylint, pytest, docs
+# As it may be undesired to make formatting changes, by default only check
+envlist = flake8, pylint, pytest, black
```
paride, line 3: You dropped `docs` here

paride (changes requested):
Nice! I don't like the idea of addind a GH action to auto-commit the black changes (I know it's not part of this PR): I think that belongs to pre-commit, but let's add it in a separate PR.

One inline comment in `tox.ini`, for the rest LGTM, thanks!

pyproject.toml
paride: +1

paride (approved):
LGTM, thanks!

orndorffgrant (approved):
Looks good!

I also like the idea of a pre-commit hook for formatting more than the github action. That is what we do in ubuntu-advantage-client.

Also flake8 does provide some checks that black won't fix, such as finding unused imports. So it's probably best to keep around.

I'm not opinionated on line length as long as black is enforcing it :)

PR #176, opened 2021-12-16
Isort (SC-701)

TheRealFalcon:
Should have just done this at the same time as black.

As before, the config changes are separate commits so should be easier to review.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 # As it may be undesired to make formatting changes, by default only check
-envlist = flake8, pylint, pytest, black, docs
+envlist = flake8, pylint, pytest, black, isort, docs
 skipsdist = true
 
 [common]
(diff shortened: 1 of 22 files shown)


tox.ini
```diff
@@ -33,11 +33,18 @@ envdir = {[common]envdir}
 deps = {[common]deps}
 commands = {envpython} -m black . --check
 
+[testenv:isort]
+envdir = {[common]envdir}
+deps = {[common]deps}
+commands = {envpython} -m isort . --check-only
+
 [testenv:format]
 # Should also include isort in the future
```
paride, line 42: The future is now! :-)

paride (changes requested):
One comment line to drop, for the rest LGTM and ready to land.

paride (approved):
Thanks!

PR #177, opened 2022-01-12
lxd vm: wait up to 600s for the agent to come up

paride:
Apparently 300s are not enough when the host machine is busy, in
particular Bionic seems to be likely to hit the 300s timeout.


TheRealFalcon (approved):
Thanks!

PR #181, opened 2022-02-07
set max azure-cli-core version

holmanb:
In the logs of a failed test:
```
21:50:41     def get_azure_cli_credentials(resource=None, with_tenant=False):
21:50:41         """Return Credentials and default SubscriptionID of current loaded profile of the CLI.
21:50:41     
21:50:41         *Disclaimer*: This method is not working for azure-cli-core>=2.21.0 (released in March 2021).
21:50:41         It is now recommended to authenticate using https://pypi.org/project/azure-identity/ and AzureCliCredential.
```

I think this "fix" will get our tests passing, but since this has been depreciated it looks like we'll eventually need to switch to azure-identity.  Alternatively we could add the azure-identity dependency and modifications right now, if preferred.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ INSTALL_REQUIRES = [
     "azure-mgmt-resource >= 10.0.0, < 15",
     "azure-mgmt-network >= 11.0.0, < 16",
     "azure-mgmt-compute >= 13.0.0, < 17",
-    "azure-cli-core >= 2.9.1",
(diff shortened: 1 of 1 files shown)


paride (approved):
Interesting, I dug for that Disclaimer and it's a comment in the code, it's not even in the docs. I think we can live with this for now, we already have azure deps with capped versions.

blackboxsw (approved):
I'm good with pinning this version. We can then bump with intention when we really need to for new features.

PR #183, opened 2022-02-11
Stop using short options on "lxc list"

TheRealFalcon:
The '-f' flag is no longer available. Long options are better anyway
because they are more readable.


blackboxsw (approved):
While I don't see '-f' being dropped on either lxc --version 4.15  or lxc --version 4.22 I can't argue with long options being easier to read/maintain.

Issue #179, opened 2022-01-26
[config] key paths should be expanded (~ -> $HOME)


paride:
The `public_key_path` and `private_key_path` config file entries are not expanded via `.expanduser()`, so if there a `~` there is remains literal. I think they should be expanded instead.

I noticed this while using the `[ec2]` settings, but I don't see code treating the other clouds differently.

PR #185, opened 2022-03-03
gce: expand ~user in credentials_path

paride:
Fixes: #179

### DIFF
--- a/pycloudlib/cloud.py
+++ b/pycloudlib/cloud.py
@@ -4,6 +4,7 @@
 import getpass
 import io
 import logging
+import os
 from abc import ABC, abstractmethod
 
 import paramiko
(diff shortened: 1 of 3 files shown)


blackboxsw (commented):
Should we also be using os.path.expanduser in [KeyPair pubkey privkey handling](https://github.com/canonical/pycloudlib/blob/main/pycloudlib/key.py#L21-L25) too

paride:
Fun thing: the bug I filed was the the key paths, but I ended up fixing the gce credentials path. We want both.

PR updated, thanks for reviewing.

blackboxsw (approved):
Thanks for the @paride. I confirm your branch handles pycloudlib.toml config-based issues. I think we still should sort os.path.expanduser in [KeyPair pubkey privkey handling](https://github.com/canonical/pycloudlib/blob/main/pycloudlib/key.py#L21-L25) but likely as a separate PR. This KeyPair entrypoint is called from BaseCloud.use_key() or could be instantiated directly and needs safeguards for expanduser because those keys will likely come from the user's home dir.

PR #187, opened 2022-03-30
Update Bionic LXD VM profile

lucasmoura:
Currently, there is a known [regression](https://github.com/lxc/distrobuilder/commit/04f32caf6f0e874e856ef8fc88f1c85275211dfb) that prevents us from installing the lxd-agent on Bionic VMs. Since we need
the service running to boot the VM through pycloudlib, we are providing a temporary solution until that issue is properly fixed
on LXD

### DIFF
--- a/pycloudlib/lxd/defaults.py
+++ b/pycloudlib/lxd/defaults.py
@@ -1,11 +1,43 @@
 # This file is part of pycloudlib. See LICENSE file for license information.
 """LXD default values to be used by cloud and instance modules."""
+import base64
 import textwrap
 
-LXC_PROFILE_VERSION = "v3"
+LXC_PROFILE_VERSION = "v4"
+
+LXC_AGENT_PATH = "/lib/systemd/system/lxd-agent.service"
+#  pylint: disable=anomalous-backslash-in-string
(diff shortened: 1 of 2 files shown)


pycloudlib/lxd/defaults.py
```diff
@@ -1,11 +1,41 @@
...
+                systemctl start lxd-agent
+                ;;
+            *)
+        esac
+    fi
+    """.format(LXC_AGENT_PATH)  # noqa: W605
+)
 
 
-# For Xenial and Bionic vendor-data required to setup lxd-agent in a vm
```
renanrodrigo, line 8: We set it up for Xenial as well, right?

renanrodrigo (approved):
lgtm,
only black is failing afaics

paride:
@lucasmoura did we just forget about this one?

lucasmoura:
Yeah, I opened that PR when we had the kernel issue regarding launching LXD vms. I will double check to see if this is already fixed on LXD though

lucasmoura:
Looks like this is already fixed on newer versions of LXD. closing it now

PR #188, opened 2022-04-11
Bump EC2 instance type (SC-924)

TheRealFalcon:
# Proposed commit message
```
Bump EC2 instance type

Updates the default ec2 instance type to t3.micro as it supports
IPv6 metadata while also being faster and cheaper than t2.
```

### DIFF
--- a/pycloudlib/ec2/cloud.py
+++ b/pycloudlib/ec2/cloud.py
@@ -172,7 +172,7 @@ class EC2(BaseCloud):
     def launch(
         self,
         image_id,
-        instance_type="t2.micro",
+        instance_type="t3.micro",  # Using nitro instance for IPv6
         user_data=None,
         wait=True,
         vpc=None,


pycloudlib/ec2/instance.py
```diff
@@ -126,6 +126,12 @@ def delete(self, wait=True):
         if wait:
             self.wait_for_delete()
 
+    def enable_ipv6_instance_metadata(self):
```
blackboxsw, line 129: I don't think we want to typically run this after the instance launches because there could be a race with the instance coming up earlier than our IPv6 IMDS metadata makes it through AWS backplane, so cloud-init may have already passed the point where early IMDS detection passed on an initially IPv4-only IMDS.

I think we actually want this at launch which looks like we can provide via something like [create_instances kwargs MetadataOption={'HttpProtocolIpv6': 'enabled'}](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.ServiceResource.create_instances). 
I'm testing this suggestion now. That said, we might want an easy way to provide a simple param during ec2 instance launch to specify IMDS 'ipv4'|'ipv6'|'dual-stack' at instance launch time so we don't have to digup and remember these options at every call-site.

blackboxsw (commented):
Thanks @TheRealFalcon. If you have a use-case for why we want to set this up after initial instance creation,  then I think we can keep this method as defined. But, my initial thought was we'd really like to do this just as instance launches, which would either require passing kwargs into the create_instances call or even letting pycloudlib attempt to [setup  create_launch_templates for ipv4, dual-stack and ipv6-only metadata](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.create_launch_template) which we can use at create_instances call.

TheRealFalcon:
> If you have a use-case for why we want to set this up after initial instance creation, then I think we can keep this method as defined.

I don't. I just didn't see another way of way of doing it, but now I see it can also be passed to run_instances, so agreed, that would be preferred.

blackboxsw:
> > If you have a use-case for why we want to set this up after initial instance creation, then I think we can keep this method as defined.
> 
> I don't. I just didn't see another way of way of doing it, but now I see it can also be passed to run_instances, so agreed, that would be preferred.

OK, let's adapt this PR then to minimally update default instance_type to t3.micro, add a comment about support of IPv6 IMDS (as it's also required for IPv6-only Nitro instance support too).

TheRealFalcon:
Sounds good. I'm going to leave this branch up as-is until we can have the cloud-init integration test work without it.

holmanb (approved):
LGTM

PR #189, opened 2022-04-15
config: expand environment variables in paths (SC-941)

paride:
config: expand environment variables in paths

This is particularly helpful in Jenkins where file credentials are put
in a random location which exposed via an environment variable, which
we can now specify in the toml config.

### DIFF
--- a/pycloudlib/cloud.py
+++ b/pycloudlib/cloud.py
@@ -35,11 +35,15 @@ class BaseCloud(ABC):
 
         user = getpass.getuser()
         self.key_pair = KeyPair(
-            public_key_path=os.path.expanduser(
-                self.config.get("public_key_path", f"~{user}/.ssh/id_rsa.pub")
+            public_key_path=os.path.expandvars(
+                os.path.expanduser(
+                    self.config.get(
+                        "public_key_path", f"~{user}/.ssh/id_rsa.pub"
+                    )
+                )
             ),
-            private_key_path=os.path.expanduser(
-                self.config.get("private_key_path", "")
+            private_key_path=os.path.expandvars(
+                os.path.expanduser(self.config.get("private_key_path", ""))
             ),
             name=self.config.get("key_name", user),
         )
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -64,8 +64,8 @@ class GCE(BaseCloud):
                 "GOOGLE_APPLICATION_CREDENTIALS"
             ]
         elif "credentials_path" in self.config:
-            self.credentials_path = os.path.expanduser(
-                self.config["credentials_path"]
+            self.credentials_path = os.path.expandvars(
+                os.path.expanduser(self.config["credentials_path"])
             )
 
         credentials = get_credentials(self.credentials_path)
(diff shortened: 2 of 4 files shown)


paride:
Note that I first expand `~` and then variables, as we don't want to expand any `~` which was *in* the variables: we're keeping those literal.

pycloudlib/cloud.py
```diff
@@ -35,11 +35,15 @@ def __init__(
 
         user = getpass.getuser()
         self.key_pair = KeyPair(
-            public_key_path=os.path.expanduser(
-                self.config.get("public_key_path", f"~{user}/.ssh/id_rsa.pub")
+            public_key_path=os.path.expandvars(
+                os.path.expanduser(
+                    self.config.get(
+                        "public_key_path", f"~{user}/.ssh/id_rsa.pub"
```
blackboxsw, line 41: Could this default just be ${USER} instead of ${user} to get it from  default environ variables? Or are you looking to actually provide a separate "user" env var in CI that will be different than "jenkins"?

Running this on our jenkins node as the "jenkins" user gives me the following:

```bash
$ python3 -c 'import os; print(os.path.expandvars("test/${USER}"))'
test/jenkins
```

```suggestion
                        "public_key_path", f"~{USER}/.ssh/id_rsa.pub"
```

blackboxsw (approved):
Changeset looks reasonable to me either with $USER or $user. Just wanted to be sure of your use-case.

pycloudlib/cloud.py
TheRealFalcon: This isn't ${user}, it's an f-string referencing a `user` variable defined about 5 lines up. I don't think the expansion is useful in the default value here...I think it's intended for `public_key_path` possibly containing an env var.

TheRealFalcon (approved):

pycloudlib/cloud.py
paride: We could use `$HOME` (expanded by expandvars) instead of `~{user}` (f-string then expanded by expanduser) now, but we still need `user` (Python variable) a few lines below, in:
```
name=self.config.get("key_name", user),
```
so we can't fully drop it.

blackboxsw: Right, sorry for the miss on f-string here.  +1.

blackboxsw (approved):
LGTM thanks for clarification on my misread.

blackboxsw:
We probably don't have an immediate need to use environment variables in runners of ec2.(public_key_path|private_key_path) and openstack(public_key_path|private_key_path) to separate test runner paths?

paride:
@blackboxsw right we don't have immediate need for those, but I thought to expand all the paths in the conf file in the same way, doing `expandvars()` + `expanduser()`. We *may* have use for those if one day we decide to store the SSH keys as Jenkins credentials. But if you see them as possibly problematic we can expand only what we actually need.

blackboxsw:
>  @blackboxsw right we don't have immediate need for those, but I thought to expand all the paths in the conf file in the same way, doing expandvars() + expanduser(). We may have use for those if one day we decide to store the SSH keys as Jenkins credentials. But if you see them as possibly problematic we can expand only what we actually need.

+1 on this approach as written. Sorry, I realize my last question left the impression that I was disagreeing with OpenStack/Ec2 support for the same keys that you added via the base cloud.py changes.   I was only wondering if we had an immediate need at the moment for Ec2/OpenStack in the way we are setting up Jenkins runners, or whether this need was currently limited to how we currently setup and run GCE.

All good to land as is.

PR #190, opened 2022-04-16
Include Azure released images (SC-943)

### LINKED ISSUE #96: Cannot launch release images on Azure
This is documented in the code, but is not at all obvious without examining the codebase: https://github.com/canonical/pycloudlib/blob/1c4f3cd6a0eff165e830a43edb6527735efc895d/pycloudlib/azure/cloud.py#L440-L454

We should, at the very least, log something to explain to people what's happening, but I suspect if we just maintained an `UBUNTU_RELEASE`-like dict which had the release SKUs instead of the daily SKUs then we'd be set.

TheRealFalcon:
Previously, released_image() pointed to daily_image() on Azure. This
commit allows us to launch released images as well.

Fixes #96

### DIFF
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -25,6 +25,7 @@ UBUNTU_RELEASE_IMAGES = {
     "bionic": "Canonical:UbuntuServer:18.04-LTS",
     "focal": "Canonical:0001-com-ubuntu-server-focal:20_04-lts-gen2",
     "impish": "Canonical:0001-com-ubuntu-server-impish:21_10-gen2",
+    "jammy": "Canonical:0001-com-ubuntu-server-jammy:22_04-lts-gen2",
 }
 
 
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -460,14 +460,12 @@ class Azure(BaseCloud):
             )
 
     def _get_image(self, release, image_map):
-        release = image_map.get(release)
-        if release is None:
+        image_id = image_map.get(release)
+        if image_id is None:
             msg = "No Ubuntu image found for {}. Expected one of: {}"
-            raise ValueError(
-                msg.format(release, " ".join(UBUNTU_DAILY_IMAGES.keys()))
-            )
+            raise ValueError(msg.format(release, " ".join(image_map.keys())))
 
-        return release
+        return image_id
 
     def released_image(self, release):
         """Get the released image.
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -12,19 +12,26 @@ from pycloudlib.cloud import BaseCloud
 from pycloudlib.config import ConfigFile
 from pycloudlib.util import get_timestamped_tag, update_nested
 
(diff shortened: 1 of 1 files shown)


pycloudlib/azure/cloud.py
```diff
@@ -452,21 +459,27 @@ def delete_image(self, image_id):
                 "Error deleting %s. Request returned %d", image_id, resp_code
             )
 
+    def _get_image(self, release, image_map):
+        release = image_map.get(release)
+        if release is None:
+            msg = "No Ubuntu image found for {}. Expected one of: {}"
+            raise ValueError(
+                msg.format(release, " ".join(UBUNTU_DAILY_IMAGES.keys()))
```
blackboxsw, line 467: We should use image_map.keys() here because UBUNTU_DAILY_IMAGES and UBUNTU_RELEASED_IMAGES will frequently differ in keys.
```suggestion
                msg.format(release, " ".join(image_map.keys()))
```

pycloudlib/azure/cloud.py
```diff
@@ -452,21 +459,27 @@ def delete_image(self, image_id):
                 "Error deleting %s. Request returned %d", image_id, resp_code
             )
 
+    def _get_image(self, release, image_map):
+        release = image_map.get(release)
+        if release is None:
```
blackboxsw, line 464: In failure paths, `release` is getting reset to None in the case that we don't find it. So the error message ends up being unhelpful.

*** ValueError: No Ubuntu image found for None. Expected one of: xenial bionic focal impish jammy.


```suggestion
        image_id = image_map.get(release)
        if image_id  is None:
```

pycloudlib/azure/cloud.py
```diff
@@ -452,21 +459,27 @@ def delete_image(self, image_id):
...
+            msg = "No Ubuntu image found for {}. Expected one of: {}"
+            raise ValueError(
+                msg.format(release, " ".join(UBUNTU_DAILY_IMAGES.keys()))
+            )
+
+        return release
```
blackboxsw, line 470: Per above suggestions
```suggestion
        return image_id
```

blackboxsw (commented):
Thanks for this @TheRealFalcon ideally, I'd like unittest coverage for this.... but, our current Azure.__init__  has so many side-effects that need mocking I'd rather push a refactor of our intialization to a separate branch rather than fight/mock all the setup that __init__ is doing to add a simple unittest.

Minor changes requested to have proper error messages.

blackboxsw (changes requested):
marking changes requested to resolve the previous comment.

blackboxsw (approved):
Confirmed launch of instances and proper error message from daily and release images on Azure Jammy 
```bash
(Pdb) image_id
'Canonical:0001-com-ubuntu-server-jammy-daily:22_04-daily-lts'
(Pdb) release_image_id
'Canonical:0001-com-ubuntu-server-jammy:22_04-lts-gen2'
(Pdb) image2 = client.released_image(release="bogus")
DEBUG:pycloudlib.cloud:finding release Ubuntu image for bogus
*** ValueError: No Ubuntu image found for bogus. Expected one of: xenial bionic focal impish jammy
```

PR #194, opened 2022-05-06
Raise exception when no results from streams query (SC-996)

TheRealFalcon:
Currently if a streams query finds no results, we'll get an
'IndexError: list index out of range' exception at the call site.
This commit adds a more meaningful exception if no images are found from
the streams query.

E.g., from a kinetic run (which hasn't yet hit dailies): https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-kinetic-lxd_container/3/consoleFull

Now we should get a message like:
```
ValueError: No images found matching filters: ['combined_squashfs_sha256=kinetic']
```

### DIFF
--- a/pycloudlib/cloud.py
+++ b/pycloudlib/cloud.py
@@ -205,4 +205,8 @@ class BaseCloud(ABC):
             keyring_path="/usr/share/keyrings/ubuntu-cloudimage-keyring.gpg",
         )
 
-        return stream.query(filters)
+        result = stream.query(filters)
+        if not result:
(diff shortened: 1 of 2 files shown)


paride:
Hmm, the change looks sane but CI is unhappy.

TheRealFalcon:
I missed that `_image_info` gets overrided twice for LXD and hadn't seen the VM implementation.

blackboxsw (approved):
Yes please. LGTM

PR #195, opened 2022-05-10
Create default LXD VM profile

TheRealFalcon:
For every new Ubuntu series, we have been creating a new LXD VM profile.
Since the profiles for the past few releases have been the same, we
should be able to a default profile from now on with exceptions for
Xenial and Bionic.

### DIFF
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -8,7 +8,7 @@ import yaml
 
 from pycloudlib.cloud import BaseCloud
 from pycloudlib.constants import LOCAL_UBUNTU_ARCH
-from pycloudlib.lxd.defaults import LXC_PROFILE_VERSION, base_vm_profiles
+from pycloudlib.lxd.defaults import base_vm_profiles
 from pycloudlib.lxd.instance import LXDInstance, LXDVirtualMachineInstance
 from pycloudlib.util import subp
 
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -609,9 +609,9 @@ class LXDVirtualMachine(_BaseLXD):
         """
         image_id = self._normalize_image_id(image_id)
         base_release = self._extract_release_from_image_id(image_id)
-        profile_name = "pycloudlib-vm-{}-{}".format(
-            base_release, LXC_PROFILE_VERSION
-        )
+        if base_release not in ["xenial", "bionic"]:
+            base_release = "default"
+        profile_name = f"pycloudlib-vm-{base_release}-{LXC_PROFILE_VERSION}"
 
         self.create_profile(
             profile_name=profile_name,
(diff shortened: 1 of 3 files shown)


pycloudlib/lxd/cloud.py
```diff
@@ -609,9 +609,9 @@ def build_necessary_profiles(self, image_id):
         """
         image_id = self._normalize_image_id(image_id)
         base_release = self._extract_release_from_image_id(image_id)
-        profile_name = "pycloudlib-vm-{}-{}".format(
-            base_release, LXC_PROFILE_VERSION
-        )
+        if base_release not in ["xenial", "bionic"]:
+            base_release = "default"
```
paride, line 613: I like this change, just one question: how are we going to handle a future release requiring a profile change (that is, a introducing a new default)? At that point we'll want to get rid of the `default` profile defined for the older releases. Is your idea to do this via a profile version bump?

There's another possibility: make `base_release` the *first* release whose profile is compatible with the release of `image_id`. In practical term this means doing:
```
if base_release not in ["xenial", "bionic"]:
    base_release = "focal"
```
but this leaves room for expanding the condition if needed, e.g. with `base_release = 'focal'` for `bionic <= image_id release <= jammy` and `base_release = 'kinetic'` for `image_id release >= kinetic`, assuming that kinetic breaks compatibility.

lucasmoura: I agree with @paride. I think using the base profile scheme will be better if we need to change the profile for a release in the future

lucasmoura: We could rename it to `pycloudlib-lxd-default` or am I missing a hidden characteristic of using the `default` name ?

TheRealFalcon: That's what this PR does.

I thought the comment from Paride that you agreed with is saying changing the default at some future time would cause problems, so we should name it something other than default. His comment specifies 'focal'

lucasmoura: Okay, as far as I understand here, the name change of the profile will be `pycloudlib-vm-default-v3`. The problem with that approach is the bump issue that @paride  mentioned that we would need to invalidate all of the profiles if one release needs a change.

The idea to avoid that was the one that @paride mentioned to use the first release to use the default profile. Now, with your question I thought that instead we could use just: `pycloudlib-vm-default` without even the string version.

If we need to change the whole default profile, we can just add a string version to it, like `pycloudlib-vm-default-v1`
and if we need to specialize it just for this release.

TL;DR I think we can keep this PR as is, just dropping the `LXC_PROFILE_VERSION` from it. I think we just need to agree about that migration of the profile names in the future

TheRealFalcon: Gotcha, thanks. I like that idea. I'll do that. 

Moving forward, I don't expect our VM profiles to change much given that we're no longer working around LXD VM since Bionic. Technically, we don't even need custom profiles anymore. The default profile should work as-is.

paride: @lucasmoura definitely improved on my initial idea, thanks! I like this solution.

paride:
I'm going to test this with some bootspeed jobs.

paride (approved):
LGTM!

PR #196, opened 2022-05-11
Consistent boot

TheRealFalcon:
```
Improve boot/reboot behavior

* Catch EOFError when trying to ssh
* Keep trying if SSHException raised during wait for cloud-init
* Ignore by default GCE timing out when terminating instance
* Add boot id tracking for reboots
```

I don't love the launch/_launch indirection I added, but that seems to be the only (non [overly clever](https://stackoverflow.com/a/71206442)) way to add common base class behavior. We'll need a similar mechanism if we want a CLI that stores/reads state over time.


pycloudlib/instance.py
```diff
@@ -398,28 +408,27 @@ def _tmpfile(self):
...
-        self._log.debug("_wait_for_execute to complete")
-        test_instance_command = "whoami"
+    def _wait_for_boot_id(self):
+        """Wait until we obtain a new boot id from the instance.
+
+        This also ensures we can execute commands on the instance.
+        """
+        self._log.debug("_wait_for_boot_id to complete")
 
         # Wait 40 minutes before failing
```
paride, line 418: Good occasion to add a comment on why 40 minutes? :)

pycloudlib/instance.py
```diff
@@ -99,7 +109,7 @@ def _wait_for_instance_start(self):
     def wait(self):
         """Wait for instance to be up and cloud-init to be complete."""
```
paride, line 110: I may be wrong here, but wait() now also always waits for the boot_id to change, via `_wait_for_boot_id()`. This means that if one calls `wait()` without also doing a reboot/restart of the instance, the `wait()` call will timeout with an error. Maybe we need a separate `wait_new_boot()` function, or `wait()` could grow a `require_new_boot_id` argument.

pycloudlib/instance.py
```diff
@@ -398,28 +408,27 @@ def _tmpfile(self):
         self._tmp_count += 1
         return path
 
-    def _wait_for_execute(self):
-        """Wait until we can execute a command in the instance."""
-        self._log.debug("_wait_for_execute to complete")
-        test_instance_command = "whoami"
+    def _wait_for_boot_id(self):
```
paride, line 411: As it is implemented now I think this should be called `_wait_for_new_boot_id()`, but see my other comment on issues with always checking for new boot_id on `wait()`.

TheRealFalcon:
Thanks for the feedback @paride . I pushed a newer version that I think doesn't have the same boot id issues. I haven't updated tests or done any integration testing yet, but just wanted to throw the new idea out there.

TheRealFalcon:
This should be ready for re-review

paride (approved):
LGTM and this is working smoothly in the bootspeed jobs.

PR #199, opened 2022-05-26
No log

TheRealFalcon:
While we're waiting for a reboot we get a lot of log spam about the boot id or waiting for cloud-init.


paride (approved):
LGTM!

PR #201, opened 2022-06-02
oci: select proper subnet matching availability_domain

blackboxsw:
There can be multiple subnets in a virtual cloud network (VCN).
Typically we see 3 subnets in an availability_domain like ua-ashburn-1:
 - US-ASHBURN-AD-1
 - US-ASHBURN-AD-2
 - US-ASHBURN-AD-3

pycloudlib must select the appropriate subnet matching the specific
availability_domain and not just source the first subnet at index 0.

Wrong subnet ID results in 400 erros from python SDK or UI dashboard
errors such as:
A problem occurred while preparing the instance's VNIC.
    ((400, InvalidParameter, false)
Parameter 'availabilityDomain' does not match.
VNIC has 'iad-ad-3' while the subnet has 'iad-ad-1'
    (opc-request-id: dummyRequestId))

### DIFF
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -221,6 +221,9 @@ class OCI(BaseCloud):
         for subnet in subnets:
             if subnet.availability_domain == self.availability_domain:
                 subnet_id = subnet.id
+                break
+        else:
+            subnet_id = subnets[0].id
         if not subnet_id:
             raise Exception(
                 f"Unabled to determine subnet id for domain: "
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -214,11 +214,18 @@ class OCI(BaseCloud):
 
         """
         vcn_id = self.network_client.list_vcns(self.compartment_id).data[0].id
-        subnet = self.network_client.list_subnets(
+        subnets = self.network_client.list_subnets(
             self.compartment_id, vcn_id=vcn_id
-        ).data[0]
-        subnet_id = subnet.id
-
+        ).data
+        subnet_id = None
+        for subnet in subnets:
+            if subnet.availability_domain == self.availability_domain:
+                subnet_id = subnet.id
+        if not subnet_id:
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
Subnets can be linked to a specific availability domain, but they don't have to be. The 3 subnets in the nonadmin default VCN correspond to a specific AD. However, this won't work on the newer server compartment as we currently have a single subnet not tied to a particular availability domain. 

What about
```python
diff --git a/pycloudlib/oci/cloud.py b/pycloudlib/oci/cloud.py
index e02ccec..560d7c6 100644
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -221,6 +221,9 @@ class OCI(BaseCloud):
         for subnet in subnets:
             if subnet.availability_domain == self.availability_domain:
                 subnet_id = subnet.id
+                break
+        else:
+            subnet_id = subnets[0].id
         if not subnet_id:
             raise Exception(
                 f"Unabled to determine subnet id for domain: "
```

blackboxsw:
> Subnets can be linked to a specific availability domain, but they don't have to be. The 3 subnets in the nonadmin default VCN correspond to a specific AD. However, this won't work on the newer server compartment as we currently have a single subnet not tied to a particular availability domain.
> 
> What about
> 
> ```python
> diff --git a/pycloudlib/oci/cloud.py b/pycloudlib/oci/cloud.py
> index e02ccec..560d7c6 100644
> --- a/pycloudlib/oci/cloud.py
> +++ b/pycloudlib/oci/cloud.py
> @@ -221,6 +221,9 @@ class OCI(BaseCloud):
>          for subnet in subnets:
>              if subnet.availability_domain == self.availability_domain:
>                  subnet_id = subnet.id
> +                break
> +        else:
> +            subnet_id = subnets[0].id
>          if not subnet_id:
>              raise Exception(
>                  f"Unabled to determine subnet id for domain: "
> ```

Thanks for the review. added your correction.

PR #203, opened 2022-06-07
OciInstance: Implement network interface methods

aciba90:
OciInstance: Implement network interface methods

- Implement `add_network_interface` and `remove_network_interface`.
- Fix example

### DIFF
--- a/examples/oci.py
+++ b/examples/oci.py
@@ -21,7 +21,7 @@ def demo(availability_domain, compartment_id):
     through a number of examples.
     """
     client = pycloudlib.OCI(
-        "Oracle test",
+        "oracle-test",
         availability_domain=availability_domain,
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
Looks good! This will be nice to have in pycloudlib.

Note that it might be better to select the subnet id in the same way #201 is doing so. I think after that one merges, we might want to factor out a common "get_subnet_id" method and use that for both.

aciba90:
> Looks good! This will be nice to have in pycloudlib.
> 
> Note that it might be better to select the subnet id in the same way #201 is doing so. I think after that one merges, we might want to factor out a common "get_subnet_id" method and use that for both.

Thanks, I created a ticket for this: [SC-1100](https://warthogs.atlassian.net/browse/SC-1100).

PR #205, opened 2022-06-16
lxd: do not restart instances via stop+start

paride:
Ephemeral instanced do not survive being stopped, while they do survive
being restarted.


paride:
Doing stop/start is causing issues in the bootspeed measurements, which use ephemeral instances.

PR #210, opened 2022-08-08
azure: perform smoke test to validate CLI config

blackboxsw:
Azure recent identity changes no longer raise an error
related to CLI misconfiguration/absence on initial client creation.
We need to minimally perform an operation against the CLI to confirm
presence or absence of necessary CLI config or fallback to using the
ClientSecretCredential.

Add smoke test to attempt to list operations of the resource client
if credentials are unavailable catch the ClientAuthenticationError
and fallback to ClientSecretCredential.

### DIFF
--- a/pycloudlib/azure/util.py
+++ b/pycloudlib/azure/util.py
@@ -3,6 +3,7 @@
 import logging
 import re
 
+from azure.core.exceptions import ClientAuthenticationError
 from azure.identity import AzureCliCredential, ClientSecretCredential
 from knack.util import CLIError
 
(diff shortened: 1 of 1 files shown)


pycloudlib/azure/util.py
```diff
@@ -34,11 +35,19 @@ def get_client(resource, config_dict: dict):
     try:
         credential = AzureCliCredential()
         subscription_id = config_dict.get("subscriptionId")
-        return resource(credential, subscription_id=subscription_id)
+        client = resource(credential, subscription_id=subscription_id)
+        # smoke test credentials based on Azure CLI config if present
+        [_ for _ in client.operations.list()]
```
TheRealFalcon, line 40: Can we comment here that if this line fails we'll raise the ClientAuthenticationError so it doesn't look like a meaningless statement when we come back 3 months later?

TheRealFalcon: Also, do we actually need to build a list here or does calling `client.operations.list()` work?

blackboxsw: Yes unfortunately iterating on that list is what triggers. But we could limit that to just a single next()
```
(Pdb) client.operations.list()
<iterator object azure.core.paging.ItemPaged at 0x7fc76b9ff640>

(Pdb) _ = next(client.operations.list())
*** azure.identity._exceptions.CredentialUnavailableError: Azure CLI not found on path
```

alizing 
```suggestion
        _ = next(client.operations.list())
```

aciba90 (approved):
LGTM. Thanks for researching and putting this together.

blackboxsw:
VAlidated on a jenkins node without the azure-CLI installed 
```
 CLOUD_INIT_PLATFORM=azure CLOUD_INIT_OS_IMAGE=focal CLOUD_INIT_CLOUD_INIT_SOURCE=ppa:cloud-init-dev/daily .tox/integration-tests-jenkins/bin/pytest  tests/integration_tests/modules/test_apt.py::TestApt::test_sources_list
....
=============================================================== 1 passed ...
```

TheRealFalcon (approved):

PR #214, opened 2022-08-30
Dynamic config support

mjd-x:
Removes the requirement for pycloudlib and OCI config files to be in a
fixed path so pycloudlib can operate as a library. This allows specifying
parameters for the clouds independently of the config files, facilitating
integration with automated pipelines.

### DIFF
--- a/pycloudlib/cloud.py
+++ b/pycloudlib/cloud.py
@@ -32,7 +32,11 @@ class BaseCloud(ABC):
     _type = "base"
 
     def __init__(
-        self, tag, timestamp_suffix=True, config_file: ConfigFile = None
+        self,
+        tag,
+        timestamp_suffix=True,
+        config_file: ConfigFile = None,
+        required_values=None,
     ):
         """Initialize base cloud class.
 
@@ -44,7 +48,7 @@ class BaseCloud(ABC):
         self._log = logging.getLogger(
             "{}.{}".format(__name__, self.__class__.__name__)
         )
-        self.config = parse_config(config_file)[self._type]
+        self.check_and_set_config(config_file, required_values)
 
         user = getpass.getuser()
         self.key_pair = KeyPair(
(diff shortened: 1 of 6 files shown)


pycloudlib/cloud.py
```diff
@@ -223,3 +223,14 @@ def _streams_query(filters, daily=True):
...
+    def check_and_set_config(self, config_file):
+        """Set pycloudlib configuration.
+
+        Args:
+            config_file: path to pycloudlib configuration file
+        """
+        if config_file:
+            self.config = parse_config(config_file)[self._type]
+        else:
+            self.config = {}
```
aciba90, line 236: If `config_file == None` in `BaseCloud.__init__`, we are not trying to read the config from the default locations [CONFIG_PATHS](https://github.com/canonical/pycloudlib/blob/main/pycloudlib/config.py#L11). So, if we try to instantiate an OCI class with or without the new `config_dict` and we have a valid pycloudlib `config` in a default location, we get the following exception:

```python
>>> from pycloudlib import oci
>>> oci.cloud.OCI("test")
  File "<stdin>", line 1, in <module>
  File "/home/aciba/canonical/pycloudlib/pycloudlib/oci/cloud.py", line 56, in __init__
    availability_domain or self.config["availability_domain"]
KeyError: 'availability_domain'
```

This is also true for other clouds as, this happens in the base class initialization of every cloud.

I think we have to try to read the config in the default locations for other clouds than `OCI` and for `OCI` if `availability_domain`, `compartment_id` or `config_dict` are not given at initialization time.

aciba90 (changes requested):
Thanks, @mjd-x, for this improvement.

I have left you a comment explaining why I think we must still try to read the config from the default locations.

aciba90 (approved):
LGTM. Thanks, @mjd-x!

blackboxsw (approved):
LGTM as is. I think we have the machinery and the copy of the proposed cloud-init 22.3.3 in a separate PPA to keep iterating on making sure Azure's PPS environment better raises errors, and this existing functionality fixes the immediate upgrade issue and closed gaps in integration and unittesting to ensure we don't get down this path without a failure signal in the future on Azure.

mjd-x:
Had a quick chat with @TheRealFalcon where we decided to keep these changes over modifying `parse_config` to log instead of raise, since the config file exception makes the situation clearer for the user than getting a `KeyError` or something similar later on.

TheRealFalcon (approved):
Thanks for the update and the explanation!

PR #215, opened 2022-09-02
gce: filter daily_image by arch

orndorffgrant:
For GCE, before these changes, you might end up with an arm image when you wanted an amd image, or vice versa. This commit adds a new `arch` param to the gce `daily_image` method using the same signature as the ec2 `daily_image` which also takes an `arch` kwarg.

Test it with the following script:
```python
import logging

import pycloudlib
from pycloudlib.cloud import ImageType

gce = pycloudlib.GCE(tag="test")
for a in ["x86_64", "arm64"]:
    for r in ["kinetic", "jammy", "focal", "bionic", "xenial"]:
        gce.daily_image(r, image_type=ImageType.GENERIC, arch=a)
```

Notice that xenial doesn't have an arm image, and there are no arm images for pro.

### DIFF
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -155,22 +155,47 @@ class GCE(BaseCloud):
 
         raise ValueError("Invalid image_type: {}".format(image_type.value))
 
-    def _find_latest_image(self, release: str, image_type: ImageType):
-        project = self._get_project(image_type=image_type)
-        name_filter = self._get_name_filter(
-            release=release, image_type=image_type
+    def _query_image_list(
+        self, release: str, project: str, name_filter: str, arch: str
+    ):
+        """Query full list of images.
+
+        image list API docs:
+        https://googleapis.github.io/google-api-python-client/docs/dyn/compute_v1.images.html#list
+
+        The image list API doesn't allow filtering and sorting in one request
+        so we need to do one of those locally.
+        Filtering via the API results in fewer requests on average than
+        sorting via the API.
+        So we filter via the API and loop through all pages to get the full
+        image list matching that filter.
+        500 is the maximum allowed page size
+        Then we can sort locally and grab the latest image.
+
+        Args:
+            release: string, Ubuntu release to look for
+            project: GCE project
+            name_filter: name to filter with
+            arch: images' architecture
+
+        Returns:
+            list of images matching the given filters
+        """
+        filter_string = "(name={}) AND (architecture={})".format(
+            name_filter, arch.upper()
         )
 
-        # image list API docs:
-        # https://googleapis.github.io/google-api-python-client/docs/dyn/compute_v1.images.html#list
-        # The image list API doesn't allow filtering and sorting in one request
-        # so we need to do one of those locally.
-        # Filtering via the API results in fewer requests on average than
-        # sorting via the API.
-        # So we filter via the API and loop through all pages to get the full
-        # image list matching that filter.
-        # 500 is the maximum allowed page size
-        # Then we sort locally and grab the latest image.
+        # SPECIAL CASE
+        # Google didn't start including architecture in image descriptions
+        # until after xenial stopped getting published
+        # All xenial images are x86_64, so:
+        #   1. we can return early for non-x86_64 xenial queries
+        #   2. for xenial + x86_64 we don't include the architecture in the
+        #      filter
+        if release == "xenial":
+            if arch != "x86_64":
+                return []
+            filter_string = "name={}".format(name_filter)
 
         image_list = []
         page_token = ""
@@ -180,35 +205,31 @@ class GCE(BaseCloud):
                 self.compute.images()
                 .list(
                     project=project,
-                    filter="name={}".format(name_filter),
+                    filter=filter_string,
                     maxResults=500,
                     pageToken=page_token,
                 )
                 .execute()
             )
             reqs += 1
-            image_list = image_list + image_list_result.get("items", [])
+            image_list += image_list_result.get("items", [])
             page_token = image_list_result.get("nextPageToken", None)
 
         self._log.debug(
-            'Fetched entire image list matching "%s" in %i requests',
-            name_filter,
+            (
+                'Fetched entire image list (%i results) matching "%s" in %i'
+                " requests"
+            ),
+            len(image_list),
+            filter_string,
             reqs,
         )
 
-        if not image_list:
-            raise Exception(
-                "Could not find {} image for {} release".format(
-                    image_type.value,
-                    release,
-                )
-            )
+        return image_list
 
-        image = sorted(image_list, key=lambda x: x["creationTimestamp"])[-1]
-        self._log.debug("Found image name: %s", image["name"])
-        return "projects/{}/global/images/{}".format(project, image["id"])
-
-    def daily_image(self, release, image_type: ImageType = ImageType.GENERIC):
+    def daily_image(
+        self, release, arch="x86_64", image_type: ImageType = ImageType.GENERIC
+    ):
         """Find the id of the latest image for a particular release.
 
         Args:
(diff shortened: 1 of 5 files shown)


pycloudlib/gce/cloud.py
```diff
@@ -199,10 +203,16 @@ def _find_latest_image(self, release: str, image_type: ImageType):
             )
 
         image = sorted(image_list, key=lambda x: x["creationTimestamp"])[-1]
-        self._log.debug("Found image name: %s", image["name"])
+        self._log.debug(
```
blackboxsw, line 206: I think we should also log the searched arch when we can't find an image in the ["Could not find generic image on arch: {arch} for {release} release"](https://github.com/canonical/pycloudlib/pull/215/files#diff-6dc804f6ffbd945a53ff2ee7e65d3882f64528486f2542b3727b24050ba0a825R199)

blackboxsw (commented):
Other than the additional log architecture entry when the architecture isn't found it might be nice to:
1. add the arch= param example to [examples/gce.py daily_image call](https://github.com/canonical/pycloudlib/blob/main/examples/gce.py#L36)
2. start a unit test at pycloudlib/gce/tests/test_cloud to exercise the daily_image method just so we can build on it in the future for other cases as pycloudlib is going to start getting other contributors from Canonical's CPC-land soon.

it would 


While not strictly necessary the following might

blackboxsw (changes requested):
needs fixing for minimally the logging change when no image is found. your call on the other requests.

orndorffgrant:
@blackboxsw great suggestions! And thanks for taking a look even when I forgot to tag anyone for review haha

orndorffgrant:
@blackboxsw while addressing your comments I noticed a bug which requires a special case. I also added some unit tests, so this probably deserves a full re-review :)

aciba90 (approved):
+1. I think this PR supersedes #216, shouldn't we merge this one? 

Here is a diff fixing tox formatting issues:

<details>
  <summary>format patch</summary>
  
```diff
diff --git a/pycloudlib/gce/cloud.py b/pycloudlib/gce/cloud.py
index afbc0a9..9fbf422 100644
--- a/pycloudlib/gce/cloud.py
+++ b/pycloudlib/gce/cloud.py
@@ -158,8 +158,7 @@ class GCE(BaseCloud):
     def _query_image_list(
         self, release: str, project: str, name_filter: str, arch: str
     ):
-        """
-        Return full list of images in the given project that match the given filters
+        """Query full list of images.
 
         image list API docs:
         https://googleapis.github.io/google-api-python-client/docs/dyn/compute_v1.images.html#list
@@ -172,23 +171,31 @@ class GCE(BaseCloud):
         image list matching that filter.
         500 is the maximum allowed page size
         Then we can sort locally and grab the latest image.
-        """
 
+        Args:
+            release: string, Ubuntu release to look for
+            project: GCE project
+            name_filter: name to filter with
+            arch: images' architecture
+
+        Returns:
+            list of images matching the given filters
+        """
         filter_string = "(name={}) AND (architecture={})".format(
             name_filter, arch.upper()
         )
 
         # SPECIAL CASE
-        # Google didn't start including architecture in image descriptions until after
-        # xenial stopped getting published
+        # Google didn't start including architecture in image descriptions
+        # until after xenial stopped getting published.
         # All xenial images are x86_64, so:
         #   1. we can return early for non-x86_64 xenial queries
-        #   2. for xenial + x86_64 we don't include the architecture in the filter
+        #   2. for xenial + x86_64 we don't include the architecture in the
+        #      filter
         if release == "xenial":
             if arch != "x86_64":
                 return []
-            else:
-                filter_string = "name={}".format(name_filter)
+            filter_string = "name={}".format(name_filter)
 
         image_list = []
         page_token = ""
@@ -209,7 +216,10 @@ class GCE(BaseCloud):
             page_token = image_list_result.get("nextPageToken", None)
 
         self._log.debug(
-            'Fetched entire image list (%i results) matching "%s" in %i requests',
+            (
+                'Fetched entire image list (%i results) matching "%s" in %i'
+                " requests"
+            ),
             len(image_list),
             filter_string,
             reqs,
diff --git a/pycloudlib/gce/tests/__init__.py b/pycloudlib/gce/tests/__init__.py
index e69de29..7adcc13 100644
--- a/pycloudlib/gce/tests/__init__.py
+++ b/pycloudlib/gce/tests/__init__.py
@@ -0,0 +1,2 @@
+# This file is part of pycloudlib. See LICENSE file for license information.
+"""Main GCE tests module __init__."""
diff --git a/pycloudlib/gce/tests/test_cloud.py b/pycloudlib/gce/tests/test_cloud.py
index b2acfc3..bf6989d 100644
--- a/pycloudlib/gce/tests/test_cloud.py
+++ b/pycloudlib/gce/tests/test_cloud.py
@@ -1,7 +1,4 @@
 """Tests related to pycloudlib.gce.cloud module."""
-from io import StringIO
-from textwrap import dedent
-
 import mock
 import pytest
 
@@ -12,9 +9,18 @@ from pycloudlib.gce.cloud import GCE
 MPATH = "pycloudlib.gce.cloud."
 
 
+# pylint: disable=protected-access,missing-function-docstring
 class TestGCE:
+    """General GCE testing."""
+
     @pytest.mark.parametrize(
-        "release, arch, api_side_effects, expected_filter_calls, expected_image_list",
+        [
+            "release",
+            "arch",
+            "api_side_effects",
+            "expected_filter_calls",
+            "expected_image_list",
+        ],
         [
             ("xenial", "arm64", Exception(), [], []),
             (
@@ -57,7 +63,7 @@ class TestGCE:
             ),
         ],
     )
-    def test_query_image_list(
+    def test_query_image_list(  # noqa: D102
         self,
         release,
         arch,
@@ -110,7 +116,7 @@ class TestGCE:
     )
     @mock.patch(MPATH + "GCE._get_name_filter", return_value="name-filter")
     @mock.patch(MPATH + "GCE._get_project", return_value="project-name")
-    def test_daily_image_returns_latest_from_query(
+    def test_daily_image_returns_latest_from_query(  # noqa: D102
         self,
         m_get_project,
         m_get_name_filter,
```
</details>

orndorffgrant:
Thanks @aciba90! I think I also need to fix the unit tests as they won't pass on CI because they're trying to actually authenticate with GCE.
I can probably get around this by mocking a few things.

orndorffgrant:
@blackboxsw and @aciba90 I think my latest push fixed CI. LMK what you think. I ended up creating a `FakeGCE` class to avoid loading credentials in the unittests, but I can change that if you have a better way

pycloudlib/gce/tests/test_cloud.py
```diff
@@ -0,0 +1,148 @@
...
+import pytest
+
+from pycloudlib.cloud import ImageType
+from pycloudlib.gce.cloud import GCE
+
+# mock module path
+MPATH = "pycloudlib.gce.cloud."
+
+
+class FakeGCE(GCE):
```
blackboxsw, line 12: +1 on instrumenting this Fake. We don't want unittests reaching out to a credentialed GCP endpoint if we can avoid it as we'll get rate-limited or throttled over time because of CI load.

pycloudlib/gce/tests/test_cloud.py
```diff
@@ -0,0 +1,148 @@
...
+    @pytest.mark.parametrize(
+        [
+            "release",
+            "arch",
+            "api_side_effects",
+            "expected_filter_calls",
+            "expected_image_list",
+        ],
+        [
+            ("xenial", "arm64", Exception(), [], []),
```
blackboxsw, line 35: This is a really good start to instrumenting coverage for GCE. Thanks for all the parametrized tests.
One thing I think we generally want to do for test failure readability in the future is using `pytest.param(...., id=unique_descriptive_name)` around large sets of parameters to explain the intent of the test. With a descriptive id, we see better failure messages instead of an abbreviated dict representation of the parametrized values.

blackboxsw: Here's a consolidated diff suggestion that I'll apply here. Given it's non-functional and trivial, I'll approve and merge this
```diff
diff --git a/pycloudlib/gce/tests/test_cloud.py b/pycloudlib/gce/tests/test_cloud.py
index 990a640..3669843 100644
--- a/pycloudlib/gce/tests/test_cloud.py
+++ b/pycloudlib/gce/tests/test_cloud.py
@@ -32,8 +32,15 @@ class TestGCE:
             "expected_image_list",
         ],
         [
-            ("xenial", "arm64", Exception(), [], []),
-            (
+            pytest.param(
+                "xenial",
+                "arm64",
+                Exception(),
+                [],
+                [],
+                id="xenial_no_arm64_support_zero_sdk_list_calls",
+            ),
+            pytest.param(
                 "xenial",
                 "x86_64",
                 [{"items": [1, 2, 3]}, Exception()],
@@ -46,8 +53,9 @@ class TestGCE:
                     )
                 ],
                 [1, 2, 3],
+                id="xenial_x86_64_suppport_one_sdk_list_call_empty_pagetoken",
             ),
-            (
+            pytest.param(
                 "kinetic",
                 "arm64",
                 [
@@ -70,6 +78,7 @@ class TestGCE:
                     ),
                 ],
                 [1, 2, 3, 4, 5, 6],
+                id="non_xenial_arm64_suppport_one_sdk_list_call_per_page",
             ),
         ],
     )
```

blackboxsw (approved):

pycloudlib/gce/tests/test_cloud.py
orndorffgrant: TIL about `pytest.param` - I'll start using that everywhere I pytest now, thanks!!

PR #217, opened 2022-09-23
launch: assert image_id non None on all clouds

blackboxsw:
Let's be a bit more noisy about expected input values.

### DIFF
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -592,10 +592,14 @@ class Azure(BaseCloud):
 
         Returns:
             Azure Instance object
-
+        Raises: ValueError on invalid image_id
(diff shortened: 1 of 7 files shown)


aciba90 (commented):
Thanks, @blackboxsw, for the improvements!

Is there an explicit reason to use assertions? I know, in our use case, we use this library as a testing helper, therefore the assertions are going to be always visible.

But, other users of this library could potentially use it in a non-testing context and if they enable python optimization flags then the assertions are going to be shadowed [1].

I would suggest raising an explicit exception to be more generic.

[1] https://docs.python.org/3/using/cmdline.html#cmdoption-O

aciba90 (approved):
+1. Thanks!

PR #218, opened 2022-09-27
lxd: bionic VMs to recreate nocloud metadata templates

blackboxsw:
In Bionic, lxd-agent.service is not enabled by default.

pycloudlib used a side-channel mechanism to active the lxd-agent on VMs via profiles and a LXD config drive:

devices:
  config:
    source: cloud-init:config
    type: disk

This configuration is unnecessary for Bionic container images.

This config drive forces cloud-init to detect only the NoCloud datasource because it has higher discovery priority than DataSourceLXD.

Per LP: #1988401, LXD dropped NoCloud metadata templates from image creation in order to have common supported configuration keys and unify behavior on Bionic++ to cloud-init's DataSourceLXD instead of DataSourceNoCloud.

In order to retain NoCloud detection on Bionic VMs in integration testing, pycloudlib now recreates missing metadata templates to make sure we can properly activate the lxd-agent as well as provide meta-data, vendor-data, user-data and network-config to the Bionic VM at launch time.

To support setting LXD instance metadata config, add the following:
 - _BaseLXD.get_instance_metadata_config: to read current instance metadata
 - _BaseLXD.set_instance_metadata_config: to update nistance metadata
 - Inside LXD.launch call set_instance_metadata_config if needed on Bionic VMs to ensure cloud-init NoCloud datasource can continue to be used

## Additional context
We know a few things related to this PR:
 - Prior to any cloud image LXD metadata changes any Bionic(LXD 3.0) or Focal(LXD 4.0) host could not launch Jammy images containing `%` in the `#cloud-config`
 - https://bugs.launchpad.net/cloud-images/+bug/1988401 is today being reverted and LXD metadata no-cloud templates are being added back into bionic and focal to avoid breaking existing customers(Juju,OSM,Anbox) with complex #cloud-config user-data which contain `%` symbols
 - LXD 4.0 has a backport they'll release to LXD 4.0 stable stream in a month or so which will fix Focal hosts
https://github.com/lxc/lxd/issues/10951
 - Bionic LXD 3.0-based won't get this fix
 - cloud image creation process for Focal may drop the LXD nocloud metadata templates at some point after LXD 4.0 supports rendering user-data to `/dev/lxd/sock` which contains a `%`  character.

Given that Focal will eventually receive changes in behavior in cloud image stream metadata, it's best if pycloudlib plans for the eventuality that Nocloud templates could drop from VM images on Bionic. In that case we should re-constitute our expected nocloud metadata templates to allow us to setup a testable customized environment with lxd-agent running on Bionic VM images. If bionic VM images already have the right nocloud metadata templates, NOOP. If they get dropped in the future, reset them so pycloudlib doesn't need a hotfix to react to whenever that those metadata templates change.

### DIFF
--- a/pycloudlib/lxd/cloud.py
+++ b/pycloudlib/lxd/cloud.py
@@ -3,12 +3,17 @@
 import warnings
 from abc import abstractmethod
 from contextlib import suppress
+from typing import List
 
 import yaml
 
 from pycloudlib.cloud import BaseCloud
 from pycloudlib.constants import LOCAL_UBUNTU_ARCH
-from pycloudlib.lxd.defaults import base_vm_profiles
+from pycloudlib.lxd.defaults import (
+    BIONIC_VM_METADATA_CONFIG_NOCLOUD,
+    BIONIC_VM_TEMPLATES_NOCLOUD,
+    base_vm_profiles,
+)
 from pycloudlib.lxd.instance import LXDInstance, LXDVirtualMachineInstance
 from pycloudlib.util import subp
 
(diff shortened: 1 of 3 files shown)


blackboxsw:
We'll close this PR for now since it should not be strictly needed by end of week if all daily cloud-image updates for Bionic and Focal continue to deliver LXD nocloud metadata seed templates. If this changes in the future on Bionic, we can apply this changeset prior to such image metadata changes landing in the future.

Issue #220, opened 2022-10-17
LXD's `ip` method fails if instance has multiple NICs


TheRealFalcon:
My `lxc list` output has something like:
```
+-------------------------------+---------+---------------------+-----------------------------------------------+-----------------+-----------+
| cloudinit-1017-040803ieevd0hh | RUNNING | 10.20.30.99 (eth2)  | fd42:640c:ab08:2325:216:3eff:fe9f:385d (eth2) | CONTAINER       | 0         |
|                               |         | 10.20.30.218 (eth0) | fd42:640c:ab08:2325:216:3eff:fe3c:2dc9 (eth0) |                 |           |
+-------------------------------+---------+---------------------+-----------------------------------------------+-----------------+-----------+
```

[The ip method](https://github.com/canonical/pycloudlib/blob/main/pycloudlib/lxd/instance.py#L98) doesn't understand this and raises exception after timing out.

holmanb:
Is the exception happening at the assignment?

Testing locally I see:
```bash
arc~ lxc list me -c4 --format csv
"10.161.80.42 (eth0)
10.161.80.127 (eth0)"
```
and
```python
>>> ips = """10.161.80.42 (eth0)
... 10.161.80.127 (eth0)"""
>>> ips.split()
['10.161.80.42', '(eth0)', '10.161.80.127', '(eth0)']
>>> ip_address, _dev = ips.split()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: too many values to unpack (expected 2)
```

If that's the case, this cheeky one-character fix could work:
```python
  diff --git a/pycloudlib/lxd/instance.py b/pycloudlib/lxd/instance.py
  index 41f18b8..8fcc30a 100644
  --- a/pycloudlib/lxd/instance.py
  +++ b/pycloudlib/lxd/instance.py
  @@ -120,7 +120,7 @@ class LXDInstance(BaseInstance):
                   ip_address = None
                   try:
                       # Expect "<ip> (<interface>)" when network fully configured
  -                    ip_address, _dev = result.stdout.split()
  +                    ip_address, *_dev = result.stdout.split()
                   except ValueError:
                       self._log.debug(
                           "Unable to parse output of cmd: %s. Expected"

```

TheRealFalcon:
Probably...I didn't really look into it all. I just wanted an issue to reference in my cloud-init workaround :smile: .

[...]

PR #221, opened 2022-10-19
lxd: support multiple interfaces for lxd bridges

holmanb:
This (straw man) approach selects the non-lxd bridge between two interfaces. It solves the problem where the second interface is created via nested instances. I'm fine if we'd rather reject this solution in favor of something less quick and dirty.

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -70,6 +70,10 @@ class LXDInstance(BaseInstance):
         return subp(base_cmd + list(command), rcs=None)
 
     def parse_ip(self, query: dict):
+        """Return ip address from lxd query.
+
+        Returns None if no address found
+        """
         network = query.get("state", {}).get("network", {})
         for _, nic_cfg in network.items():
             if not nic_cfg.get("host_name"):
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -1,5 +1,6 @@
 # This file is part of pycloudlib. See LICENSE file for license information.
 """LXD instance."""
+import json
 import re
 import time
 
@@ -109,28 +110,36 @@ class LXDInstance(BaseInstance):
         while retries != 0:
             command = [
                 "lxc",
-                "list",
-                "^{}$".format(self.name),
-                "-c4",
-                "--format",
-                "csv",
+                "query",
+                f"/1.0/instances/{self.name}?recursion=1",
             ]
             result = subp(command)
             if result.ok and result.stdout:
-                ip_address = None
                 try:
-                    # Expect "<ip> (<interface>)" when network fully configured
-                    ip_address, _dev = result.stdout.split()
+                    info = json.loads(result.stdout)
                 except ValueError:
                     self._log.debug(
-                        "Unable to parse output of cmd: %s. Expected"
-                        " <ip> (<interface>), got: %s. Retrying %d time(s)...",
+                        "Unable to parse output of cmd: %s. Expected YAML,"
+                        " got: %s. Retrying %d time(s)...",
                         command,
                         result.stdout,
                         retries,
                     )
-                if ip_address:
-                    return ip_address
+                    continue
+                network = info.get("state", {}).get("network", {})
+                for _nic, nic_cfg in network.items():
+                    if not nic_cfg.get("host_name"):
+                        continue
+                    for addr in nic_cfg["addresses"]:
+                        if addr.get("family") == "inet":
+                            return addr.get("address")
+                self._log.debug(
+                    "Unable to find valid IP from cmd: %s. Found network:"
+                    " %s. Retrying %d time(s)",
+                    command,
+                    network,
+                    retries,
+                )
             retries -= 1
             time.sleep(1)
         raise TimeoutError(
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -119,7 +119,7 @@ class LXDInstance(BaseInstance):
                     info = json.loads(result.stdout)
                 except ValueError:
                     self._log.debug(
-                        "Unable to parse output of cmd: %s. Expected YAML,"
+                        "Unable to parse output of cmd: %s. Expected JSON,"
                         " got: %s. Retrying %d time(s)...",
                         command,
                         result.stdout,
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -85,6 +85,7 @@ class LXDInstance(BaseInstance):
             "Unable to find valid IP. Found network: %s",
             network,
         )
+        return None
 
     @property
     def is_vm(self):
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -69,6 +69,19 @@ class LXDInstance(BaseInstance):
         ]
         return subp(base_cmd + list(command), rcs=None)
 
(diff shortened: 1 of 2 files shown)


holmanb:
Addresses the nested-lxd case for https://github.com/canonical/pycloudlib/issues/220

blackboxsw (commented):
It felt originally that our relying on lxc ls output was fragile. Now may be the motivation to change what we are talking to for more complex instances.

Alternatives:
*  Use machine-readable format `lxc ls <name> --format=json` which doesn't give us concrete information on which device's IP to surface
* Use `lxc info <name` which gives is YAML with ["Resources"]["Network usage"] which publishes a  "Host interface" key for the primary outbound device. We could grab 'inet' value from that if we wanted and it'd allow us to rule out internal bridge devices

a diff of what I was playing with
```diff
diff --git a/pycloudlib/lxd/instance.py b/pycloudlib/lxd/instance.py
index 41f18b8..7d11424 100644
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -107,30 +107,32 @@ class LXDInstance(BaseInstance):
         retries = 150
 
         while retries != 0:
-            command = [
-                "lxc",
-                "list",
-                "^{}$".format(self.name),
-                "-c4",
-                "--format",
-                "csv",
-            ]
+            command  ["lxc", "info", self.name])
             result = subp(command)
             if result.ok and result.stdout:
                 ip_address = None
                 try:
-                    # Expect "<ip> (<interface>)" when network fully configured
-                    ip_address, _dev = result.stdout.split()
+                    info = yaml.safe_load(result.stdout)
                 except ValueError:
                     self._log.debug(
-                        "Unable to parse output of cmd: %s. Expected"
-                        " <ip> (<interface>), got: %s. Retrying %d time(s)...",
+                        "Unable to parse output of cmd: %s. Expected YAML,"
+                        " got: %s. Retrying %d time(s)...",
                         command,
                         result.stdout,
                         retries,
                     )
-                if ip_address:
-                    return ip_address
+                    continue
+                for nic, nic_cfg in info["Resources"]["Network usage"].items():
+                    if "Host interface"in nic_cfg:
+                        if nic_cfg.get("IP addresses", {}).get("inet"):
+                            return nic_cfg.get("IP addresses", {}).get("inet")
+                self._log.debug(
+                    "Unable to find valid IP from cmd: %s. Found Network usage:"
+                    " %s. Retrying %d time(s)",
+                    command,
+                    info["Resources"]["Network usage"],
+                    retries,
+                )
             retries -= 1
             time.sleep(1)
         raise TimeoutError(

```

holmanb:
Update: Talking with lxd folks, 

```
lxd query /1.0/instances/v1?recursion=1
```

with the returned json, `query ["state"]["query"]` -> returns a list, iterate until finding a dict that contains the key "host_name" which will be the host primary interface

blackboxsw:
> Update: Talking with lxd folks,
> 
> ```
> lxd query /1.0/instances/v1?recursion=1
> ```
> 
> with the returned json, `query ["state"]["query"]` -> returns a list, iterate until finding a dict that contains the key "host_name" which will be the host primary interface

Looks like lxc query  works great for this.... Here's a diff that should resolve things for us
```diff
diff --git a/pycloudlib/lxd/instance.py b/pycloudlib/lxd/instance.py
index 41f18b8..41ad477 100644
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -1,5 +1,6 @@
 # This file is part of pycloudlib. See LICENSE file for license information.
 """LXD instance."""
+import json
 import re
 import time
 
@@ -109,28 +110,36 @@ class LXDInstance(BaseInstance):
         while retries != 0:
             command = [
                 "lxc",
-                "list",
-                "^{}$".format(self.name),
-                "-c4",
-                "--format",
-                "csv",
+                "query",
+                f"/1.0/instances/{self.name}?recursion=1",
             ]
             result = subp(command)
             if result.ok and result.stdout:
-                ip_address = None
                 try:
-                    # Expect "<ip> (<interface>)" when network fully configured
-                    ip_address, _dev = result.stdout.split()
+                    info = json.loads(result.stdout)
                 except ValueError:
                     self._log.debug(
-                        "Unable to parse output of cmd: %s. Expected"
-                        " <ip> (<interface>), got: %s. Retrying %d time(s)...",
+                        "Unable to parse output of cmd: %s. Expected YAML,"
+                        " got: %s. Retrying %d time(s)...",
                         command,
                         result.stdout,
                         retries,
                     )
-                if ip_address:
-                    return ip_address
+                    continue
+                network = info.get("state", {}).get("network", {})
+                for _nic, nic_cfg in network.items():
+                    if not nic_cfg.get("host_name"):
+                        continue
+                    for addr in nic_cfg["addresses"]:
+                        if addr.get("family") == "inet":
+                            return addr.get("address")
+                self._log.debug(
+                    "Unable to find valid IP from cmd: %s. Found network:"
+                    " %s. Retrying %d time(s)",
+                    command,
+                    network,
+                    retries,
+                )
             retries -= 1
             time.sleep(1)
         raise TimeoutError(
```

holmanb:
@blackboxsw Thanks! That patch looks good to me, I've updated the PR with it.

pycloudlib/lxd/instance.py
```diff
@@ -109,28 +110,36 @@ def ip(self):
...
-                ip_address = None
                 try:
-                    # Expect "<ip> (<interface>)" when network fully configured
-                    ip_address, _dev = result.stdout.split()
+                    info = json.loads(result.stdout)
                 except ValueError:
                     self._log.debug(
-                        "Unable to parse output of cmd: %s. Expected"
-                        " <ip> (<interface>), got: %s. Retrying %d time(s)...",
+                        "Unable to parse output of cmd: %s. Expected YAML,"
```
blackboxsw, line 122: Sorry, should have been JSON
```suggestion
                        "Unable to parse output of cmd: %s. Expected JSON,"
```

pycloudlib/lxd/instance.py
```diff
@@ -109,28 +110,36 @@ def ip(self):
...
-                        " <ip> (<interface>), got: %s. Retrying %d time(s)...",
+                        "Unable to parse output of cmd: %s. Expected JSON,"
+                        " got: %s. Retrying %d time(s)...",
                         command,
                         result.stdout,
                         retries,
                     )
-                if ip_address:
-                    return ip_address
+                    continue
```
holmanb, line 128: @blackboxsw - In case you have a question when reviewing, the continue addition here causes an infinite loop (`retries` is never decremented), so I'm going to change the logic to do what I think this code was supposed to do.

holmanb: The fix (https://github.com/canonical/pycloudlib/pull/221/commits/2cef127657fe5cf9fb7003ff2492f825a837f33a) also refactors parsing logic away from retry logic into `parse_ip()` - this was to make it easier to test parsing logic in isolation easier (and also to make the code a little more readable). Note the `s/continue/else/` in this changeset to fix the exception case. I should have made that into two separate commits to make the review easier.

holmanb:
Simplestreams/launchpad was sad when last the CI tests ran. Rerunning now.

holmanb:
CI is now green, should be ready for review.

blackboxsw (approved):
Thank you for the separation of parse_ip and the test coverage here. This should solve our issues with filtering out bridge info vs host interfaces.

blackboxsw:
@holmanb sorry for delay here. Will let you squash merge with an appropriate commit message representing the diff.

PR #222, opened 2022-11-18
feat(oracle): add kwargs for oci operations

mjd-x:
Adds kwargs support for all oci operations.

Example usage: Oracle's SDK has retry strategies that can be used for most operations. Even though some of them (like launch_instance) do not retry by default, a retry strategy can be specified.

### DIFF
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -108,13 +108,13 @@ class OCI(BaseCloud):
         self.compute_client = oci.core.ComputeClient(self.oci_config)
         self.network_client = oci.core.VirtualNetworkClient(self.oci_config)
 
-    def delete_image(self, image_id):
+    def delete_image(self, image_id, **kwargs):
         """Delete an image.
 
         Args:
             image_id: string, id of the image to delete
         """
-        self.compute_client.delete_image(image_id)
+        self.compute_client.delete_image(image_id, **kwargs)
 
     def released_image(self, release, operating_system="Canonical Ubuntu"):
         """Get the released image.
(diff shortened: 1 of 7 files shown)


pycloudlib/oci/cloud.py
```diff
@@ -217,6 +223,7 @@ def launch(
         instance_type="VM.Standard2.1",
         user_data=None,
         wait=True,
+        retry_strategy=None,
```
aciba90, line 226: Being this an actual keyword-only argument (because it is not defined as an explicit arg in the base class, therefore it belongs to the **kwargs interface and it is captured here), I wonder if we should enforce it:

```suggestion
        *,
        retry_strategy=None,
```

By doing so, we would prevent client-code consuming this function like:

```python
oci.launch(image_id, instance_type, user_data, wait, "my_cool_retry_strategy")
```

which kind of break the Liskov Substitution Principle, as instances of other child classes do not have to have a fifth positional argument.

More info: [PEP 3102 – Keyword-Only Arguments](https://peps.python.org/pep-3102/)

pycloudlib/oci/cloud.py
```diff
@@ -108,13 +108,14 @@ def __init__(
         self.compute_client = oci.core.ComputeClient(self.oci_config)
         self.network_client = oci.core.VirtualNetworkClient(self.oci_config)
 
-    def delete_image(self, image_id):
+    def delete_image(self, image_id, **kwargs):
```
aciba90, line 111: I wonder if we should propagate the `**kwargs` definition to the base class and every child class.

By doing so, we would allow other instances of other child classes to ignore extra `kwargs` without breaking. For example, the following call would raise if `ec2` is not an instance of `oci.cloud.Cloud`:

```python
ec2.delete_image("image-id-asdf", other_param="qwer")
```

aciba90 (commented):
Thanks, @mjd-x. I left a couple of comments underlying some concerns about the interface modifications.

aciba90 (approved):
LGTM. Thanks, @mjd-x!

PR #226, opened 2022-12-12
azure: if az command absent, use explicit config params

blackboxsw:
## please rebase merge when approved. I'd like a separate commit for which -> command -v in instance.py

When no 'az' command is installed, avoid relying on a default client created by:
client = resource(credential, subscription_id=subscription_id)

On systems without az command line client, the default client will later fail with messages like to following:

azure.identity._exceptions.CredentialUnavailableError:
   Azure CLI not found

### DIFF
--- a/pycloudlib/azure/util.py
+++ b/pycloudlib/azure/util.py
@@ -7,6 +7,8 @@ from azure.core.exceptions import ClientAuthenticationError
 from azure.identity import AzureCliCredential, ClientSecretCredential
 from knack.util import CLIError
 
+from pycloudlib import util
+
 logger = logging.getLogger(__name__)
 
 RE_AZURE_IMAGE_ID = (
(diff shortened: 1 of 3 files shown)


pycloudlib/azure/util.py
```diff
@@ -32,20 +32,26 @@ def get_client(resource, config_dict: dict):
         The client for the resource passed as parameter.
 
     """
-    try:
-        credential = AzureCliCredential()
-        subscription_id = config_dict.get("subscriptionId")
-        client = resource(credential, subscription_id=subscription_id)
-        return client
-    except CLIError:
+    if not util.subp(["which", "az"], rcs=[0, 1]).ok:
```
paride, line 35: At some point `which` was going to be deprecated and printed a deprecation warning to stderr. This has been reconsidered *for now* (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=993582), so I think it's better to use `command -v` instead. Note that it requires a shell (`subp(..., shell=True`)` I think).

PR #233, opened 2023-01-13
lxd: delete ephemeral instances using lxc delete

paride:
There are cases where ephemeral instances are not running and not
automatically deleted, e.g. when they are in ERROR state. Let's
just call `lxc delete` as we do for non-ephemeral images, and trust
lxd to do the right thing.

This is a partial revert of 0869d07b7894d985e183be957ae8318aae667c00.

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -231,12 +231,7 @@ class LXDInstance(BaseInstance):
         """
         self._log.debug("deleting %s", self.name)
 
-        if self.ephemeral:
-            # We don't need to wait here, since the
-            # instance will be deleted once it is stopped
-            self.shutdown(wait=False)
-        else:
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
Only impact I could see here is time costs to test runs if `lxc delete <ephermal_instance> --force` blocks for a long time when resources are tight/over-subscribed.

Running tests in general on ambipom with and without this patch didn't result in significant time costs in test instance teardown in happy resource cases.

# comparison tip of main vs current branch for the following tests:
# tests_set_hostname.py tests_keys_to_console.py tests_ntp_servers.py 

# main
9m 55s
# current branch 
10m 0s

We can watch jobs in cloud-init to see if we see job timeout errors or extended delays due to failure to delete --force ephemeral instances..

Now that we've upgraded lxd to 5.0/stable channel I presume a number of these issues we saw in the past are now resolved.

PR #236, opened 2023-01-24
CI: add dependabot.yml to auto-update dependencies

aciba90:
GH actions and lint requirements will be try to be updated in autogenerated PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/dev"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 5
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
--- a/dev/requirements.txt
+++ b/dev/requirements.txt
@@ -1,0 +1,9 @@
+mypy==0.950
+types-mock==5.0.0
+types-pyyaml==6.0.12
+types-setuptools==65.6.0
+types-toml==0.10.8
+flake8==4.0.1
+flake8-docstrings==1.6.0
+pylint==2.12.2
+black==22.3.0
--- a/tox.ini
+++ b/tox.ini
@@ -6,15 +6,7 @@ skipsdist = true
 [common]
 envdir = {toxworkdir}/.testenv
 deps =
-    mypy==0.950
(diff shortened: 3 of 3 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
```
aciba90, line 3: This part will basically mimic the tip targets in Jenkins by creating a PR with updated dependencies and the added benefit of:

2. If the build passes, we can straight merge those PRs.
3. If it does not, we see that some style needs to be fixed.

If we try and like this approach, then we could remove the tip jobs in Jenkins.

aciba90: This is false, because tip targets check against the last version of python.

paride: I like this approach, but I can't find documentation on which requirement files dependabot will update, at least not in [1]. Will it update `dev/requirements.txt` and `test-requirements.txt`? Do you happen to have a reference for this?

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

aciba90: AFAICT, at the moment one only can specify the directory to search the manifest files and `dependabot` will discover them there. That's why I put the file under `dev/` and configure it in line 4.

I found some discussions explaining that behavior as for example: https://github.com/dependabot/dependabot-core/issues/3940#issuecomment-1227005267

The source code is here: https://github.com/dependabot/dependabot-core/blob/main/python/helpers/lib/parser.py

paride: Found it, apparently it probes `**/.txt`:
```
    requirement_files = glob.glob(os.path.join(directory, '*.txt')) \
        + glob.glob(os.path.join(directory, '**', '*.txt'))
```
Thanks for digging it up!

paride (approved):
LGTM, thanks!

paride:
@aciba90 is this actually still a draft?

aciba90:
> @aciba90 is this actually still a draft?

Ready! Thanks, @paride for the review. I wanted to make sure that the default configuration does not auto-merge PRs or anything weird. Checked and it looks safe: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/managing-pull-requests-for-dependency-updates

I am going to ping other as reviewers.

TheRealFalcon (approved):
LGTM!

PR #237, opened 2023-01-24
(oracle) reuse retry strategy

mjd-x:
An [OSError](https://docs.python.org/3/library/exceptions.html#OSError) is not the correct exception to raise when connection to an instance times out, so a new custom exception for this was added. If a timeout occurs, the orphaned instance will now be deleted from oracle.
Also added type hinting to some methods that manage instances since sometimes these returns would be treated as BaseInstance instead of OciInstance.

### DIFF
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -280,7 +280,9 @@ class OCI(BaseCloud):
         instance_data = self.compute_client.launch_instance(
             instance_details, retry_strategy=retry_strategy
         ).data
-        instance = self.get_instance(instance_data.id)
+        instance = self.get_instance(
+            instance_data.id, retry_strategy=retry_strategy
+        )
         if wait:
             wait_till_ready(
                 func=self.compute_client.get_instance,
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -6,7 +6,7 @@ import base64
 import json
 import os
 import re
-from typing import Optional
+from typing import Optional, cast
 
 import oci
 
@@ -79,7 +79,7 @@ class OCI(BaseCloud):
                     result.stdout, result.stderr
                 )
                 raise Exception(exception_text)
-            compartment_id = json.loads(result.stdout)["data"]["id"]
+            compartment_id = cast(str, json.loads(result.stdout)["data"]["id"])
         self.compartment_id = compartment_id
 
         if config_dict:
@@ -196,7 +196,7 @@ class OCI(BaseCloud):
         """
         raise NotImplementedError
 
-    def get_instance(self, instance_id, **kwargs):
+    def get_instance(self, instance_id, **kwargs) -> OciInstance:
         """Get an instance by id.
 
         Args:
@@ -205,9 +205,10 @@ class OCI(BaseCloud):
         Returns:
             An instance object to use to manipulate the instance further.
         """
+        # verifies that instance id exists in oracle
         try:
             self.compute_client.get_instance(instance_id, **kwargs)
-        except oci.exceptions.ServiceError as e:
+        except oci.exceptions.ServiceError as e:  # type: ignore
             raise Exception(
                 "Unable to retrieve instance with id: {} . "
                 "Is it a valid instance id?".format(instance_id)
@@ -230,7 +231,7 @@ class OCI(BaseCloud):
         *,
         retry_strategy=None,
         **kwargs,
-    ):
+    ) -> OciInstance:
         """Launch an instance.
 
         Args:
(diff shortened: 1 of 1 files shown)


TheRealFalcon (commented):
I don't think pycloudlib should be auto-deleting instances. If I can't connect to an instance after a timeout, there's likely a deeper issue that I want to be able to dig in and debug. If the instance has been deleted, I can no longer do that.. This is something that can be just as easily added to consuming projects.

Also, changing the exception thrown like this could also cause regressions in that we have consuming code expecting one exception, but now receiving another. It's certainly a change we can make, but I think until we have semantic versioning, we need to get buy-in from the consuming teams to make sure they know they may need to make changes before updating.

mjd-x:
I've proposed these changes to delete the instance after a timeout precisely because it can't be done from the consuming project. The `launch` method will not return the instance data if any exception is caught, even if the instance does exist in Oracle. The only way to delete the instance is to manually log in to Oracle and delete it.

Some more context as to why I proposed this: we've been seeing this timeout error quite frequently in the past few months. I've been investigating it, and whenever I try to connect to the instance that timed out with the SSH key in my config, I get an access denied error. I don't have means to debug much since I can't access it at all. The instance is left orphaned.
This tends to happen a few times a week, so at the end of the week I have a couple of orphaned instances running on Oracle that don't serve any purpose.

Bottom line is that I can't delete the instance from the consuming project because pycloudlib doesn't return the instance data, so I need to check into Oracle every couple days and manually delete all the orphaned instances left.

Maybe this could also be solved by either adding the instance id to the exception or having some kind of flag such as `no_clean_up=False`. I proposed this solution in particular just because I thought it was the best one myself, as I don't think anyone wants to suddenly get a raise in their cloud bill because of these instances running.

TheRealFalcon:
Hmmm, I'm worried this could be a cloud-init problem. I've had to resort to the serial console in the past for issues like this. If you give me your launch details I could try to reproduce the problem. I'd rather at least start with that route.

mjd-x:
We launch the instances like this in our project:

 ```
instance = client.launch(
         image_id=image_ocid.image_model.id,
         instance_type=self.cloud.instance_type,
         shape_config=shape_config,
         retry_strategy=oci.retry.DEFAULT_RETRY_STRATEGY,
)
 ```
Where `client` is a `pycloudlib.OCI` instance, and `shape_config` is either `None` for non-flex instance types, or a `oci.core.models.LaunchInstanceShapeConfigDetails` instance.

Even if we can get to the bottom of the timeout problem, I still think it's necessary to at least return the instance information in some kind of way, maybe by using one of the solutions I suggested in my last comment. If any error comes up during the launch process, the consuming project has no way of knowing the OCID of the instance that was just launched, so the only way to delete it is by logging into oracle and doing it manually.

TheRealFalcon:
> I still think it's necessary to at least return the instance information in some kind of way

Does something like this work?
```python
instance = client.launch(<your_args>, wait=False)
try:
    instance.wait()
except:
    instance.delete()
```

TheRealFalcon:
Hey @mjd-x , I see your new `clean_up` parameter, but my preference is still to keep things as they are. A few reasons:

1. I don't see what we gain by adding a new parameter other than saving 4-ish lines of code from the caller (i.e., my previous comment). Is there a stronger use case that I'm not aware of? If you think using the API this way isn't obvious enough, it probably wouldn't be too hard to get buy-in to set `wait=False` as default. Similarly, if the `OSError` is a problem, we can get the buy-in to change it.
2. We should aim to keep the API as consistent as possible across the clouds. Obviously this can't apply to any cloud-specific options, but in this case, this is a general parameter that could apply to any cloud. If we were to add the `clean_up` parameter here, it'd be better to apply it consistently across the API.
3. I prefer one obvious way to accomplish something. Having both `wait` and `clean_up` muddy that somewhat. Furthermore, `wait=False` is more powerful than `clean_up` in that you can get access to the instance so that you can inspect the instance before deletion.

Is there another use case or a broader need here?

mjd-x:
Dropped the main commit that added the clean up, leaving the other ones as those are still useful additions
We decided to commit to add a workaround to clean up the orphaned instances in our project

hammerstefan:
I would still like to revisit the earlier discussion around pycloudlib cleanup on failure, exception types & exception handling, and how to handle API changes.  @TheRealFalcon do you want me to create an issue on GH to discuss, start an email chain, or set up a meet on your calendar?

TheRealFalcon:
@hammerstefan , GH might be best just because it'll be the most public and able to have the most visibility, but I'm open to whatever works best for you.

TheRealFalcon (approved):
Thanks!

TheRealFalcon:
@mjd-x Do you have merge permissions? If so I'll leave it up to you to decide how to you want commit (e.g., squash vs rebase) and the commit message(s). If not, let me know if you want me to merge a certain way.

PR #238, opened 2023-01-25
lxd: fix ip parsing

aciba90:
Extend LXD instance ip parsing to detect
nics with global address an no hostname.

Previously, nics without a hostname were ignored.
This config has been found on instances with bonds
and bridges.

### DIFF
--- a/pycloudlib/lxd/instance.py
+++ b/pycloudlib/lxd/instance.py
@@ -77,10 +77,14 @@ class LXDInstance(BaseInstance):
         network = query.get("state", {}).get("network")
         if network is None:
             network = {}
-        for _, nic_cfg in network.items():
-            if not nic_cfg.get("host_name"):
-                continue
+        for nic_cfg in sorted(
+            network.values(),
+            # prefer nics with host_name defined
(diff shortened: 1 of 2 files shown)


pycloudlib/lxd/instance.py
```diff
@@ -77,10 +77,14 @@ def parse_ip(self, query: dict):
...
-        for _, nic_cfg in network.items():
-            if not nic_cfg.get("host_name"):
-                continue
+        for nic_cfg in sorted(
+            network.values(),
+            # prefer nics with host_name defined
+            key=lambda cfg: not cfg.get("host_name"),
+        ):
             for addr in nic_cfg["addresses"]:
+                if addr.get("scope") != "global":
```
aciba90, line 86: Do you think, @holmanb, there is a better way to discover IPs in this (see the new test) situations?

holmanb: I don't have anything better at the moment. That looks good to me.

TheRealFalcon (approved):
LGTM!

PR #256, opened 2023-02-03
Add continuous deployment with semantic versioning

TheRealFalcon:
Squash merge this one. The latest commit is to make it easier to see the untested PyPI parts.
I have already added my pycloudlib key to this repo (any committer can replace it (but not view it) by going into settings->secrets) so upload should work as it did to the test PyPI.

# Proposed Commit Message
```
Add continuous deployment with semantic versioning

Add new VERSION file to hold version. Enforce with Github Actions that
it must be updated on every PR. Also include Github Action to tag
and deploy to PyPI after PR is merged. Docs also updated accordingly.
```

# Test Procedure
I tested on my fork ensuring version checking, tagging, and uploading to the TEST PyPI works correctly.

ON A FORK...
checkout main
echo "0.0.1" > VERSION
commit and push

checkout this PR
git reset HEAD~ --hard (to checkout the commit that uses test.pypi.org rather than real PyPI)
change VERSION to "0.0.1"
commit and push to fork and PR against fork's main
`Check Semantic Versioning` check should fail with
`VERSION file must be updated`

change version to "1.2.3.4.5"
commit and push
`Check Semantic Versioning` check should fail with
`VERSION must be in 1!<major>.<minor>.<patch> format, not 1.2.3.4.5`

Remove latest commit so version goes back to 1.0.0 (git reset HEAD~ --hard)
commit and push
`Check Semantic Versioning` check should pass

Checkout main and git reset upstream/main --hard
commit and push to main
Retrigger PR checks (e.g., open and close)
`Check Semantic Versioning` check should pass

(Attempting to reproduce past this step won't work unless you're a maintainer of the pycloudlib project on test.pypi.org and you bump the version to something other than what already exists there. See https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)
Merge PR to fork's main
Ensure repo now has a 1.0.0 tag
Ensure https://test.pypi.org/project/pycloudlib/#history has 1.0.0


.github/workflows/tag_and_publish.yaml
```diff
@@ -0,0 +1,29 @@
+name: Tag and Publish Release
+on:
+  push:
+    branches:
+      - main
+
```
aciba90, line 6: We could limit concurrent executions of this job by adding:
```suggestion

`concurrency: ${{ github.ref }}`

```

aciba90 (commented):
Thanks, @TheRealFalcon.

It looks very good. I left a couple of comments / questions.

By including this as-is, we are going to publish a new version to PyPI for every push to main, including changes that do not affect the library code, as for example changes under `docs`, etc. I guess, we could limit this somehow in a future PR.

.github/workflows/tag_and_publish.yaml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  tag-and-publish:
+    runs-on: ubuntu-18.04
+    steps:
```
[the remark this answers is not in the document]
TheRealFalcon: Good point. I added another commit (but rebased below HEAD) to deal with this. I split the job into two jobs where the tag has an output and the publish consumes the version number from the tagging. We should be able to re-run the deploy job if it fails then. Let me know if you think that makes sense.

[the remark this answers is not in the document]
aciba90: I think this approach is better than removing tags and messing up with the repo history. Thanks!

aciba90 (approved):
LGTM. Thanks!

.github/workflows/tag_and_publish.yaml
```diff
@@ -0,0 +1,43 @@
...
+    outputs:
+      version: ${{ steps.tag-release.outputs.version}}
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v3
+      - name: Tag release
+        id: tag-release
+        run: |
+          version=$(cat VERSION)
+          git tag $version
```
blackboxsw, line 21: Do we need to check for major version bumps and drive annotated tags so `git describe` better represents major or major.minor version bumps?

blackboxsw (commented):
Looks really good and nice github workflows. Minor changes suggested and a general question about where we want to store the version.

At some point do we think we'll use our existing pyproject.toml to better represent `[project]` and the necessary `version` config there? If so does this separate VERSION file complicate that shift to pyproject.toml/setup.cfg representation of project metadata?

It seems [pypi is documenting that as a preferred approach for managing project metadata](https://packaging.python.org/en/latest/tutorials/packaging-projects/). Not something we need to solve now, but something to be wary of when building our versioning tooling. I suppose it really only slightly complicates out `cat VERSION` with a `grep 'version =' pyproject.toml`   and a sed to update that version.

blackboxsw (changes requested):
Needs fixing on ubuntu-18.04 references and probably a bit of the  version_check discussion

.github/workflows/version_check.yaml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check version
+        id: check-version
+        run: |
+          if [[ $(git show main:VERSION &> /dev/null) -eq 0 ]] &&  # Ignore if file doesn't exist yet
```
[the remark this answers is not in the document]
TheRealFalcon: Yep, this check is worthless as the next line actually covers the case here too. I think we can remove it entirely.

.github/workflows/version_check.yaml
```diff
@@ -0,0 +1,27 @@
...
+             [[ $(git diff origin/main origin/${GITHUB_HEAD_REF} -- VERSION | wc -c) -eq 0 ]]
```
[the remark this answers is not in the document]
TheRealFalcon: This one works as expected because we're piping to `wc -c`. If `wc -c` is 0, that means there are no changes between the files. If this check wasn't working correctly, this action would be failing on this PR with 'VERSION file must be updated'.

.github/workflows/tag_and_publish.yaml
TheRealFalcon: I'm not really sure the use case for `git describe` on pycloudlib. What's the value in seeing how many commits there have been since the last major or minor version bump?

TheRealFalcon:
> At some point do we think we'll use our existing pyproject.toml to better represent [project] and the necessary version config there? If so does this separate VERSION file complicate that shift to pyproject.toml/setup.cfg representation of project metadata?
It seems [pypi is documenting that as a preferred approach for managing project metadata](https://packaging.python.org/en/latest/tutorials/packaging-projects/). Not something we need to solve now, but something to be wary of when building our versioning tooling. I suppose it really only slightly complicates out cat VERSION with a grep 'version =' pyproject.toml and a sed to update that version.

Yeah, I was looking into that, but I'd like to wait until we drop bionic/3.6 to do so. I'm thinking the transition wouldn't be too hard:
- Add `__version__` to `pycloudlib/__init__.py` that sources the VERSION file
- Update `pyproject.toml/setup.cfg` to reference `attr: package.__version__`

I'd like to wait because building with setup.cfg or pyproject.toml takes a greater setuptools version than we use in Bionic. Additionally, [there are ways](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#accessing-data-files-at-runtime) of accessing package data at runtime that requires python 3.7+

.github/workflows/tag_and_publish.yaml
blackboxsw: Good thought. I suppose in this CD world we are bumping major, minor or subrevision with each merged commit anyway.... So, that commit count offset  is encoded in the subrevision basically.

.github/workflows/tag_and_publish.yaml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  push:
+    branches:
+      - main
+
+concurrency: ${{ github.ref }}
+
+jobs:
+  tag-release:
+    runs-on: ubuntu-18.04
```
blackboxsw, line 11: @TheRealFalcon Two other cases where we probably want to bump off of ubuntu-18.04 -> ubuntu-latest in our  workflows.

blackboxsw (approved):
LGTM

PR #275, opened 2023-02-10
ci: remove comments in .python-version

aciba90:
mimic how other projects set up this undocumented dependabot's feature

### DIFF
--- a/.python-version
+++ b/.python-version
@@ -1,5 +1,1 @@
 3.6.5
-# Specifies the python version that dependabot should use
-# Note: The version must specify major.minor.patch components
-# and be supported on:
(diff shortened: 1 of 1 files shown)


aciba90:
I discovered it is possible to run dependabot on a forked repo, so I tested this in my forked repo. The problem was that the comments in `.python-version` break dependabot version detection.

One can see the difference here:
https://github.com/aciba90/pycloudlib/network/updates/601673476
https://github.com/canonical/pycloudlib/network/updates/601668528

For example, `mypy`'s newest possible version (compatible with python 3.6) is 0.971.

I have commented it here to help other devs: https://github.com/dependabot/dependabot-core/issues/1455#issuecomment-1425925176

paride (approved):

Issue #263, opened 2023-02-08
Pycloudlib exception types are too generic, occassionally wrong


hammerstefan:
Pycloudlib is frequently throwing exceptions that are too generic for a use to be able to handle them intelligently.

Too Generic:
There are 25 cases in pycloudlib of `raise Exception(...)`.  Additionally there are 9 cases of `raise RuntimeError()`. When calling code tries to handle that error, there is no clear way to know what or where an error occurred. Syntax Error? Out of local resources? Pycloud misconfigured? Problem with cloud? Out of cloud resources? Network went down? Cloud API library outdated? NotImplementedError? Timeout? Etc, etc. Most cases of the `raise Exception` includes an error description that might be useful to a human, but is near useless for calling python code to handle.

Wrong exception types:
There are 2 cases of `raise OSError()` for cloud interaction errors. That is not a correct exception type. https://docs.python.org/3/library/exceptions.html#OSError. `OSError` is for system-related error raised from a system-related call, usually mapping to a system error code https://docs.python.org/3/library/errno.html#module-errno.

The are 7 cases of `raise TimeoutError()` for cloud interaction errors. This is an `OSError` and should only pertain to local system and maps to ETIMEDOUT system errors. It's best to avoid `OSError` exceptions for remote interactions.

** Rough Proposal **
Create a hierarchy of pycloudlib exceptions with increasing specificity

An example:
```
class PycloudlibError(RuntimeError)
class PycloudlibTimeoutError(PycloudlibError)
class PycloudlibOutOfResourcesError(PycloudlibError)
class OracleError(PycloudlibError)
class OracleTimeoutError(PycloudlibTimeoutError)
class GceError(PycloudlibError)
...
```

Making the "too generic" exceptions more generic should be safe to do, since any existing users that are doing `except Exception` will continue to catch those.  However, this would allow new users to use specific exception handlers for better error handling. Even just switching from `except Exception` to `except PycloudLibError` is a huge improvement, since you eliminate entire categories of errors that didn't want to catch (like syntax errors).

Changing the wrong exception types to more correct types could break existing users.  Ideally we do this with a symver major version release.

Furthermore, whatever exception schema we chose, it needs to be clearly documented since the user has to know how to handle errors.

blackboxsw:
https://warthogs.atlassian.net/browse/SC-1436

blackboxsw:
[...]

PR #282, opened 2023-02-14
exceptions: specialize base exceptions (SC-1436)

aciba90:
## Proposed commit msg:

```
exceptions: specialize exceptions

Add custom exceptions to improve consumer code handling abilities.

Fixes: #263
```

## Additional context:

Addresses #263


TheRealFalcon (commented):
Do you think we can get rid of `PycloudlibException`? I don't see why we need both that and `PycloudlibError`. Do any of these errors _need_ to inherit from RuntimeError? Looking at the error file, it's not immediately obvious which one I would choose if I needed to throw a new pycloudlib related error. 

Additionally, I think it might be worth having a `PycloudlibTimeoutError` for operations where we expect to wait for something to happen but it never gets there.

Otherwise, PR looks good!

pycloudlib/azure/cloud.py
```diff
@@ -873,21 +874,19 @@ def get_instance(self, instance_id, search_all=False):
...
-                raise Exception(
-                    "The image {} was already deleted".format(instance_id)
+                raise PycloudlibError(
+                    f"The image {instance_id} was already deleted"
                 )
 
             return instance
 
-        raise Exception("Could not find {}".format(instance_id))
+        raise InstanceNotFoundError("Could not find {}".format(instance_id))
```
aciba90, line 889:
```suggestion
        raise InstanceNotFoundError(resource_id=instance_id)
```

aciba90:
Thanks, @TheRealFalcon.

> Do you think we can get rid of `PycloudlibException`? I don't see why we need both that and `PycloudlibError`. Do any of these errors _need_ to inherit from RuntimeError? Looking at the error file, it's not immediately obvious which one I would choose if I needed to throw a new pycloudlib related error.

`PycloudlibException` mimics `Exception`. All custom exceptions must inherit from it, allowing consumer code to catch any exception. Pycloudlib should not raise it.

`PycloudlibError` mimics `RuntimeError`.

Having both exceptions allow clients to catch any exception or any uncategorized exception, offering more rich semantics.

I have reflected that in the documentation.

> Additionally, I think it might be worth having a `PycloudlibTimeoutError` for operations where we expect to wait for something to happen but it never gets there.

I have added that one and went forward, specializing the rest of `RuntimeError` exceptions and improved the documentation related to exceptions. I believe this solves #282.

TheRealFalcon:
Sorry, one more:
https://github.com/canonical/pycloudlib/blob/main/pycloudlib/instance.py#L481

I think that needs to be changed to a `PycloudlibTimeoutError`. 

#263 also has an issue with https://github.com/canonical/pycloudlib/blob/main/pycloudlib/instance.py/#L64 , but I don't think it makes sense to change that one. If we get to line 64, the execute command has worked, so we don't have a remote connection issue. Rather, the `cat` command itself has failed meaning `boot_id` does not exist or we had some trouble reading it. Given

> This exception is raised when a system function returns a system-related error, including I/O failures such as “file not found” or “disk full” (not for illegal argument types or other incidental errors).

I think `OSError` is reasonable in that case.

aciba90:
Agreed, @TheRealFalcon.

(I missed this) About `PycloudlibError` inheriting from `RuntimeError` in:

https://github.com/canonical/pycloudlib/blob/393da249ce7ee4479a74cfd1183236f573b13025/pycloudlib/errors.py#L19

The reason was to keep backwards compatibility, because I wanted to split the changes in two PRs because I did not have the versioning / publishing tooling.

Now that we have it and we are going to do the change at once, we could drop that inheritance.

Do you think it is worth removing it?

TheRealFalcon (approved):
> Do you think it is worth removing it?

Yes, since we're bumping major version, I don't think it's too much to ask people to update their exceptions appropriately.

+1 from me either way though. Feel free to make the change (or not) and merge.

aciba90:
> > Do you think it is worth removing it?
> 
> Yes, since we're bumping major version, I don't think it's too much to ask people to update their exceptions appropriately.
> 
> +1 from me either way though. Feel free to make the change (or not) and merge.

Thanks, I will do the change.

PR #288, opened 2023-02-16
Ignore .python-version

TheRealFalcon:
.python-version is a pyenv artifact that can also wind up taking
precedence over other activated virtualenvs.


aciba90 (commented):
Is it not sufficient to ignore `.python-version` but keep the file? 

In that way, developers could modify/remove the file and dependabot could successfully detect the minimum supported version.

aciba90:
As agreed by the team, I created https://github.com/canonical/pycloudlib/pull/289 to drop dependabot.

TheRealFalcon:
Closing as this is covered in #289 .

Issue #267, opened 2023-02-09
python-simplestreams is pinned to a commit from 2018


j5awry:
there are a few bugs listed on `simplestreams` usage. in digging more into `pycloudlib's` sstream requirements, i found that the pinned commit is from 2018

`"python-simplestreams @ git+https://git.launchpad.net/simplestreams@21c5bba2a5413c51e6b9131fc450e96f6b46090d",  # noqa`

```
commit 21c5bba2a5413c51e6b9131fc450e96f6b46090d
Author: Daniel Watkins <daniel.watkins@canonical.com>
Date:   Thu Dec 6 15:31:01 2018 +0000

    Add 'ubuntu' alias to CURRENT_LTS.
    
    This indicates the best "ubuntu" option available to users if they have
    no other preference
```

1. Looking at usage, there are 2 direct uses of streams now, `lxd` and `ec2`
2. #266 is opened to remove the `ec2` case, as SSM parameter store is how Canonical is advertising publicly for finding images.
3. I'm unsure the `lxd` case. it looks like it's to find an image _before_ launching, and then launching with a specific hash, instead of using `lxd` built-in aliases (`ubuntu:$SUITE` for release and `ubuntu-daily:$SUITE` for dailies). 
4. the query is hardcoded to the same URLs that `lxd` is already using, without an override (which would be helpful for `lxd` minimal, so providing a new stream via `lxc remote add`)

First there should be investigation if it is possible to take upstream `simplestreams` default branch and tip rather than pin. Then there should be some discussion on the `lxd` cases. It's possible that `simplestreams` may not be a required package, but rather something that can be reference if someone wants to use something other than an alias.  or removing simplestreams and creating your own query, as it's "just json", thus removing an unpackaged upstream dependency

Issue #257, opened 2023-02-06
Pycloudlib breaks user logging upon import


hammerstefan:
Pycloudlib is resetting all logging configuration upon import, which is potentially problem for users.

In `pycloudlib/streams.py`, there is code to reset all `logging` configuration.
```py
# Simplestreams import's above grab the root logger and set the
# log level up how it thinks they should be. This is bad and breaks
# end-user logging. This reloads the logging module to ignore the
# settings set by simplestreams.
logging.shutdown()
importlib.reload(logging)
```

I understand this was to work around a `simplestreams` bug, but this can actually make the situation worse, not better for the end-user. Now not just the root logger is getting unexpectedly configured, but all existing configuration is getting dropped.

This is particularly a problem in applications that are dynamically loading libraries (through a plugin architecture), because that could defer the pycloudlib module load (and logging config reset) well past the normal application initialization where logging is configured.  This is what's happening for our use, breaking all of our logging.

I propose we get a fix in place for `simplestreams`, so that it doesn't configure the root logger, and remove the breaking code from pycloudlib.

PR #290, opened 2023-02-20
lxd: drop simplestreams dependency (SC-1437)

aciba90:
## Proposed Commit Message

```
lxd: drop simplestreams dependency

Refactor lxd images' location to use lxc's cli instead of
simplestreams.

This change makes pycloudlib easier to distribute and
fixing a logging config problem associated with simplestreams.
```

## Additional Context

https://github.com/canonical/pycloudlib/issues/267
https://github.com/canonical/pycloudlib/issues/257


pycloudlib/lxd/cloud.py
```diff
@@ -21,6 +20,10 @@ class _BaseLXD(BaseCloud):
     _releases_remote = "ubuntu"
     _lxd_instance_cls = LXDInstance
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._is_container = True
```
TheRealFalcon, line 25: Does it make more sense to make this a class variable with just an annotation but no initialization? Then the child classes can override the value without needing a default here.

pycloudlib/lxd/_images.py
```diff
@@ -0,0 +1,153 @@
...
+            remote = cls._REMOTE_DAILY
+        if ":" not in remote:
+            remote += ":"
+        return remote
+
+    @classmethod
+    def _find_images(
+        cls, remote: str, filters: Optional[Sequence[Tuple[str, str]]] = None
+    ) -> List[dict]:
+        """Find the latest image for a given release.
```
TheRealFalcon, line 133: docstring out of date

pycloudlib/lxd/_images.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @classmethod
+    def find_last_fingerprint(
+        cls,
+        daily: bool,
+        release: str,
+        is_container: bool,
+        arch: str,
+    ) -> Optional[str]:
+        """Find the LXD fingerprint in a given remote.
```
TheRealFalcon, line 24: docstring is out of date

TheRealFalcon (changes requested):
Overall looks really good and nice refactor. I left a few comments, but nothing major.

pycloudlib/lxd/cloud.py
aciba90: I think it makes more sense to declare `_BaseLXD` as an abstract class and type-annotate `_is_container` within it. I have done that, let me know if you disagree.

aciba90:
Thanks, @TheRealFalcon, for the review and suggestions. I believe I have addressed all of them.

TheRealFalcon (approved):
+1 but Travis tests are failing now. Once those pass we should be good to go.

aciba90:
Close to re-trigger CI.

PR #291, opened 2023-02-21
Stop requiring branches to be upstream in version check

TheRealFalcon:
Also update CI to use `ubuntu-20.04` as 18.04 is deprecated on Github Actions.


blackboxsw (approved):
+1

PR #293, opened 2023-02-23
WIP: Add integration tests

TheRealFalcon:
Add integration tests and remove the current CI running example and cloud-init tests. Currently only running LXD in CI (as before), but it'll be easy to add additional clouds if we want to.


TheRealFalcon:
Closing in favor of #294 so I can test github secrets

PR #294, opened 2023-02-27
Add Integration tests

TheRealFalcon:
Probably easiest to review the commits in order on this one.

Add integration tests and remove the current CI running example and cloud-init tests. Since running on real clouds require credentials, but GH Actions doesn't allow the use of secrets in forks, the cloud tests have been changed to run post-merge. Late is better than never.

Since it's hard to check post-merge behavior on a PR, I temporary had the `main_check.yaml` workflow run on PR. Got a successful run of the post-merge tests at https://github.com/canonical/pycloudlib/actions/runs/4296379520/jobs/7488021663 , so it has now been changed to run post-merge.


tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+
+        instance.execute("echo 'pulled' > /var/tmp/pulled")
+        pull_path = Path(tmpdir).joinpath("pulled")
+        instance.pull_file("/var/tmp/pulled", str(pull_path))
+        assert pull_path.read_text(encoding="utf-8").strip() == "pulled"
+
+
+def exercise_instance(instance: BaseInstance):
+    with suppress(NotImplementedError):
+        assert instance.name is not None
```
blackboxsw, line 74: We shouldn't supress this NotImplementedError should we? name is declared as `@property @abstractmethod` in base class. I think it has to be there right?

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+
+    with suppress(NotImplementedError):
```
blackboxsw, line 76: Same here, I don't think we want supress for `ip` method do we?

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+        assert ipaddress.ip_address(instance.ip)
+
+    with suppress(NotImplementedError):
```
blackboxsw, line 79: Good here on supress as we have multiple clouds which don't implement add_network_interface

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+        ip_address = instance.add_network_interface()
+        try:
+            assert ipaddress.ip_address(ip_address)
+        finally:
+            instance.remove_network_interface(ip_address)
+
+    exercise_push_pull(instance)
+
+    assert_example_output(instance)
+    instance.execute("sync")
```
blackboxsw, line 89: Why do we need this sync?

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+
+import pycloudlib
+from pycloudlib.cloud import BaseCloud
+from pycloudlib.instance import BaseInstance
+from pycloudlib.util import LTS_RELEASES
+
+cloud_config = """\
+#cloud-config
+runcmd:
+  - echo 'hello' > /var/tmp/example.txt
```
blackboxsw, line 17: Can we actually make this append instead of write. This way we can assert cloud-init doesn't run multiple times every boot.
`  - echo 'hello' >> /var/tmp/example.txt`

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+
+    exercise_push_pull(instance)
+
+    assert_example_output(instance)
+    instance.execute("sync")
+    boot_id = instance.get_boot_id()
+    instance.restart()
+    assert boot_id != instance.get_boot_id()
+
+    assert_example_output(instance)
```
blackboxsw, line 94: What are we testing here? Was the intent to validate that cloud-init didn't run twice and append to /var/tmp/example.txt?

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+    instance.shutdown()
+
+    instance.start()
+    assert_example_output(instance)
```
blackboxsw, line 98: Same assertion I guess, cloud-init not running `runcmd:` multiple times? Only way to validate that would have been the `>> .../example.txt` in user-data I think.

blackboxsw (commented):
Looks great James. first half review, going through integration test now

pycloudlib/instance.py
```diff
@@ -200,6 +200,7 @@ def clean(self):
         """
         self.execute("sudo cloud-init clean --logs")
         self.execute("sudo rm -rf /var/log/syslog")
+        self.execute("sudo echo 'uninitialized' > /etc/machine-id")
```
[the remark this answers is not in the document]
TheRealFalcon: Yeah, I thought about that but thought this would be simpler. Didn't think about other distros though. I'll change it.

tests/integration_tests/test_public_api.py
TheRealFalcon: copypasta from an earlier example I stole from. Not needed anymore as the sync is built into the restart call.

tests/integration_tests/test_public_api.py
TheRealFalcon: In general (this also applies to previous comments) it is more just a general assertion that start/restart worked correctly and the instance is in an expected state. But agree about `>>` making it a better assertion.

tests/integration_tests/test_public_api.py
```diff
@@ -105,7 +105,9 @@ def exercise_instance(instance: BaseInstance):
         pytest.param(pycloudlib.EC2, id="ec2"),
         pytest.param(pycloudlib.GCE, id="gce"),
         pytest.param(pycloudlib.IBM, id="ibm"),
-        pytest.param(pycloudlib.LXDContainer, id="lxd_container"),
+        pytest.param(
+            pycloudlib.LXDContainer, id="lxd_container", marks=pytest.mark.ci
```
blackboxsw, line 107: TIL about passing marks into the pytest.param. thx

.github/workflows/ci.yaml
```diff
@@ -2,8 +2,12 @@ name: Pycloudlib CI
 
 on: [pull_request]
 
+concurrency:
+  group: "ci-${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
```
blackboxsw, line 7: This is 'safe' as far as instance cleanup mid a cancelation request because our integration tests already registered a
```
finally:
   instance.delete()
```
at launch time right? So we don't risk leaving unterminated instances after a cancelation event.

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+#cloud-config
+runcmd:
+  - echo 'hello' > /var/tmp/example.txt
+"""
+
+
+@pytest.fixture
+def cloud(request):
+    cloud_instance: BaseCloud = request.param(
+        tag="pycloudlib-integration",
```
blackboxsw, line 24: Let's trim down the tag for readability. I also think there are character limits on some clouds for instance name which gets derived from cloud.tag in some cases.

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,144 @@
...
+    latest_lts = LTS_RELEASES[-1]
+    try:
+        image_id = cloud.released_image(release=latest_lts)
+    except NotImplementedError:
+        image_id = cloud.daily_image(release=latest_lts)
+    with suppress(NotImplementedError):
+        # Not sure there's a great way to test this other than not raising
+        cloud.image_serial(image_id)
+
+    with launch_instance(
```
blackboxsw, line 128: Since lxd doesn't observe the cloud.tag in instance naming, can we standardize this instance naming before launch?
```diff
diff --git a/tests/integration_tests/test_public_api.py b/tests/integration_tests/test_public_api.py
index 262b16b..a4f272b 100644
--- a/tests/integration_tests/test_public_api.py
+++ b/tests/integration_tests/test_public_api.py
@@ -129,8 +129,9 @@ def test_public_api(cloud: BaseCloud):
         # Not sure there's a great way to test this other than not raising
         cloud.image_serial(image_id)
 
+    name = cloud.tag + f"-{latest_lts}"
     with launch_instance(
-        cloud, image_id=image_id, user_data=cloud_config, wait=False
+        cloud, image_id=image_id, user_data=cloud_config, wait=False, name=name
     ) as instance:
         instance.wait()
         exercise_instance(instance)
```

tox.ini
```diff
@@ -85,7 +85,7 @@ deps = {[tip]deps}
 commands = {envpython} -m flake8 pycloudlib examples setup.py
 
 [testenv:integration-tests]
-commands = {envpython} -m pytest -vv {posargs:tests/integration_tests}
+commands = {envpython} -m pytest --log-cli-level=INFO -svv {posargs:tests/integration_tests}
```
blackboxsw, line 88: Thx for tuning down log levels from DEBUG

.github/workflows/main_check.yaml
```diff
@@ -0,0 +1,29 @@
+name: Pycloudlib Post Merge Check
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  post-merge-tests:
```
blackboxsw, line 9: We may be able to only trigger this job in CI by checking the whether we are on the upstream repo with the following:
```
on: [pull_request]
jobs:
  my-upstream-repo-test:
    if: github.event.pull_request.head.repo.full_name == github.repository
```
This would allow us (upstream pycloudlib commiters) to push our branches to upstream/myname/X and allow CI to run on all clouds for PRs we want to. And for PRs we don't want to run on,.... just push to our own remotes/forks.

blackboxsw (commented):
Looks really good as is @TheRealFalcon. We can start with the main_check if you like, but we may be able to trigger that credentials-based job based on whether the repo from which we are running is our main "upstream" repo vs a fork. Untested suggestion on that part, so I'd rather work with something you have proven works and iterate to make it run on all our branches in future PRs.

.github/workflows/ci.yaml
TheRealFalcon: That's my understanding.

tests/integration_tests/test_public_api.py
TheRealFalcon: Heh, I actually already have a branch ready to fallback to tag if `name` isn't defined. Was gonna wait for this one to merge before putting it up, but I can put it up now-ish. You ok leaving this as-is  if we add that functionality?

TheRealFalcon:
I'd rather stick with the post-merge check for now. I'm not sure how I feel about tests running some times but not others.

tests/integration_tests/test_public_api.py
TheRealFalcon: #295

TheRealFalcon:
@blackboxsw I believe I addressed your comments. This PR should be ready for re-review.

tests/integration_tests/test_public_api.py
blackboxsw: Yes let's leave 295 separate from this PR thanks.

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,145 @@
...
+
+@contextmanager
+def launch_instance(
+    cloud_instance: BaseCloud, **kwargs
+) -> Generator[BaseInstance, None, None]:
+    instance = cloud_instance.launch(**kwargs)
+    try:
+        yield instance
+    finally:
+        instance.delete()
```
blackboxsw, line 49: Thoughts on observing a `PYCLOUDLIB_KEEP_INSTANCE=0` env var or `pycloudlib.toml` config param to avoid teardown after test completion to aid in development of integration tests? Or do we just think it's best for folks to drop a pdb in to prevent this sort of teardown during development.

```suggestion
        if bool(os.environ.get("PYCLOUDLIB_KEEP_INSTANCE", False)):
             instance.delete()
        else:
            LOG.info(
                f"Leaving instance up at {instance.name}: ({instance.ip})"
            )
```

blackboxsw (commented):
one more discussion pass env vars vs pytest.marks questions for specialized behavior during development or to limit tests to specific platforms or distros.

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,145 @@
...
+
+    instance.start()
+    assert_example_output(instance)
+
+
+@pytest.mark.parametrize(
+    "cloud",
+    [
+        pytest.param(
+            pycloudlib.Azure, id="azure", marks=pytest.mark.main_check
```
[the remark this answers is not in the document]
TheRealFalcon: What's preventing you from doing that? You can run `pytest "tests/integration_tests/test_public_api.py::test_public_api[lxd_container]"` or `tox -e integration-tests -- "tests/integration_tests/test_public_api.py::test_public_api[lxd_container]"` if you want to run the lxd_container-specific verison.

I'm not really expecting all that many "general purpose" tests. Beyond this, I would think any cloud-specific tests would be in their own directory.

tests/integration_tests/test_public_api.py
```diff
@@ -0,0 +1,145 @@
...
+        pytest.param(pycloudlib.OCI, id="oci"),
+        # For openstack we first need a reliable way of obtaining the
+        # image id
+        # pytest.param(pycloudlib.Openstack, id="openstack"),
+    ],
+    indirect=True,
+)
+def test_public_api(cloud: BaseCloud):
+    """Shallow test of (most) public functions in the base API."""
+    latest_lts = LTS_RELEASES[-1]
```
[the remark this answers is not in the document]
TheRealFalcon: > As we add other distro support like FreeBSD how do you envision us defining integration tests that apply only to specific distributions?

Haven't really thought about it much tbh. Pycloudlib is currently tightly coupled to Ubuntu in a few ways. There's ways we could make image ids be dynamic for testing, but I don't want it to be overly complicated either. I was thinking we'd figure it out when we get there.

tests/integration_tests/test_public_api.py
TheRealFalcon: At this point, I think I'd prefer we use pdb. I think we'd need a bit more test infrastructure for something like that, otherwise every integration test will be implementing this sort of thing on its own.

I think it made a lot more sense as an option in cloud-init because we were testing what happens between launch and teardown. Since here we're testing the launch and teardown themselves (along with other instance lifecycle), I'm not sure we'll need as much "in the middle" debugging. I could be wrong, and am certainly open to adding that if we need it, but for now I think it's better to keep it simple.

TheRealFalcon:
Branch had conflicts...pushed a rebase

tests/integration_tests/test_public_api.py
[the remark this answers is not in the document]
blackboxsw: that works for me, thanks for calling out the specific option there for platform-specialized set of unit tests. I was thinking toward having jenkins runners that would be opinionated based on platform support like we have for cloud-init so that if we would know specifically from the cloud platform "flavor" of CI/CD job  we'd know whether we have a problem that only happens on one platform... But we can just as easily click into a failed job log or failure artifacts to understand the breadth of the failure. No need to over-engineer the solution we have currently. thanks

blackboxsw (approved):
LGTM!

PR #295, opened 2023-03-03
LXD should use tag if no name provided

TheRealFalcon:
Will wait to bump version until after approval


pycloudlib/lxd/cloud.py
```diff
@@ -291,7 +293,7 @@ def launch(
                 f" Found: {image_id}"
             )
         instance = self.init(
-            name=name,
+            name=name or f"{self.tag}-{next(self._instance_counter)}",
```
blackboxsw, line 296: Good implementation choice differentiating multiple instance from each other.

blackboxsw (approved):
+1 LGTM

PR #297, opened 2023-03-31
Add VMWare cloud (SC-1383)

TheRealFalcon:
Integration test passed when run manually.

Additionally, I updated the pycloudlib.toml in lastpass to include Canonical-specific VMWare details.


docs/clouds/vmware.md
```diff
@@ -0,0 +1,81 @@
...
+
+The VMWare support in pycloudlib is specific to vSphere. In particular, vSphere 7 was tested.
+
+## Prerequisites
+
+VMWare usage in Pycloudlib requires the [govc](https://github.com/vmware/govmomi/tree/main/govc) command line tool to be available on the PATH. See [VMWare docs](https://docs.vmware.com/en/VMware-Telco-Cloud-Operations/1.4.0/deployment-guide-140/GUID-5249E662-D792-4A1A-93E6-CF331552364C.html) for installation information.
+
+## Available Images
+
+To create new instances, pycloudlib will clone an existing VM within vSphere that is designated as the image source. In order to quality, the VM must meet the following requirements:
```
blackboxsw, line 11:
```suggestion
To create new instances, pycloudlib will clone an existing VM within vSphere that is designated as the image source. In order to qualify, the VM must meet the following requirements:
```

pycloudlib/util.py
```diff
@@ -368,3 +370,19 @@ def get_query_param(uri: str, param: str) -> list:
...
+):
+    """Add a public key to the cloud-config."""
+    if not user_data:
+        user_data_yaml = {"ssh_authorized_keys": [public_key]}
+    else:
+        user_data_yaml = yaml.safe_load(user_data)
+        if "ssh_authorized_keys" not in user_data_yaml:
+            user_data_yaml["ssh_authorized_keys"] = []
+        user_data_yaml["ssh_authorized_keys"].append(public_key)
+    new_data = yaml.safe_dump(user_data_yaml, width=999999999)
```
blackboxsw, line 387: Why the width=999..? Do we deal with truncation issues or issues with line-wrap in vSphere?

pycloudlib/util.py
```diff
@@ -368,3 +370,19 @@ def get_query_param(uri: str, param: str) -> list:
...
+
+def add_key_to_cloud_config(
+    public_key: str,
+    user_data: Optional[str] = None,
+):
+    """Add a public key to the cloud-config."""
+    if not user_data:
+        user_data_yaml = {"ssh_authorized_keys": [public_key]}
+    else:
+        user_data_yaml = yaml.safe_load(user_data)
```
blackboxsw, line 383: If user_data is something other than `#cloud-config` ... such as `## template: jinja` this function will inadvertently break that userdata by truncating the jinja header line and replacing it with just `#cloud-config`.   I'm not sure how much we care as it's likely we'll be running this initially manually/locally and on a subset of tests.... But, maybe it's worth checking the header line and raise ValueError if user_data isn't `#cloud-config` or at least `log.warning`?

pycloudlib/vmware/cloud.py
```diff
@@ -0,0 +1,256 @@
...
+            )
+        self.env = {
+            "GOVC_INSECURE": "1" if insecure_transport else "0",
+            "GOVC_URL": server,
+            "GOVC_USERNAME": username,
+            "GOVC_PASSWORD": password,
+            "GOVC_DATACENTER": datacenter,
+            "GOVC_DATASTORE": datastore,
+            "GOVC_FOLDER": folder,
+            "PATH": os.environ.get("PATH"),
```
blackboxsw, line 89: Again, we don't want to source these variable from the environment because we want to provide them via pycloudlib.toml right? This way pycloudlib.toml overrides any environment variables... and provides a single file with secrets which is removed across test runs right?

pycloudlib/vmware/instance.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+
+    def start(self, wait=True):
+        """Start the instance.
+
+        Args:
+            wait: wait for the instance to start.
+        """
+        subprocess.run(
+            ["govc", "vm.power", "-on", self.vm_id], env=self.env, check=True
```
blackboxsw, line 81: Should this command be idempotent? As in, does it exit non-zero of power is already on? If so maybe we should `pass` on that exception because start was already performed? Same question for shutdown, should we succeed if power is already off or do we consider that a failure case as the VM maybe lost power due to something outside of test harness interaction.

pycloudlib/util.py
TheRealFalcon: pyyaml will "helpfully" add newlines on dump:
```
>>> import yaml; print(yaml.safe_dump(yaml.safe_load(open("/tmp/cloud-config.yaml"))))
ssh_authorized_keys:
- ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDVFCuncf2UzPNMH2qAupg0/8L3pIH6Y86lyfWa6nke/haXITT9xKsjLrau2aAGoyPBpgFah6TIqPIEaHCm8OjYzcpHQLRXbu3mCAK70E23ud8zubBiEZ3GSdvhedrNvxgo9p84UaEIPhDik30HzSiiv2oYrUmnnPFdacMXG8Q99khvSKnC15jylINFDAdzETKidDMV/p0xn5zVydwCxNzChJOb5nXzdTwXZ1YiB1vgRuGeOjZ77SwoTWkqFfDaUT4B7U88pTwOnAvQOIVd9LCQzcMUucAL4QdHK6XXtAciJUrG8I+h4xDt/JmpsQ4WbeYFukQt6pf1cpxSLWPigMhmFAygZZEft+gXkszELpMA6DBqy+VLjJ0/sNZzZvR7UwhG1n9o2OLdwxyw0Shxm0ZeFeFYNeNp6AaficHPEH+wGQuvgHuN35ZEgAw8MGLSoDdDPOxn5Py7gtz5fK4gSN+QtkaNFRIUZULoyBRDxr6SFVqnVvw8CxuCvkKfoygMXQk=
  james@cow

>>> 

```

the too many 9s ensures a single line stays a single line.

pycloudlib/vmware/cloud.py
TheRealFalcon: Yes...not sure if this is just asking for clarification or asking for a change, but the `env` being set here is only the env used by specific subp calls. The default environment is not modified, nor can the default environment be used. We need an env set here because that's how `govc` gets the values it needs to interact with vSphere. Priority for these env vars is:
1. Any argument passed to the function
2. pycloudlib.toml

If either of those two aren't provided, we should get a KeyError with `<setting> must be defined in pycloudlib.toml to make this call`.

pycloudlib/vmware/cloud.py
```diff
@@ -0,0 +1,256 @@
...
+                "-e",
+                "guestinfo.userdata.encoding=base64",
+            ],
+            env=self.env,
+            check=True,
+        )
+
+        instance = VMWareInstance(self.key_pair, instance_name, env=self.env)
+        instance.start()
+        if wait:
```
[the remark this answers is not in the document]
TheRealFalcon: If my other PR lands first, it'll all be moot anyway as the `wait` parameter will disappear as part of the rebase.

pycloudlib/vmware/instance.py
TheRealFalcon: I think idempotent makes more sense here. Thanks for the idea.

TheRealFalcon:
Thanks for the initial review. I believe I have addressed the current comments.

pycloudlib/util.py
holmanb: Fine by me, but could we document this for the next person that asks?

holmanb (commented):
@TheRealFalcon The other supported clouds have [examples](https://github.com/canonical/pycloudlib/tree/main/examples). Do you plan on adding something in a followup PR?

TheRealFalcon:
I actually wasn't planning on adding examples as there's really no additional functionality beyond the base API provided by VMWare. Not all clouds are represented in examples. Maybe we should have a "generic" examples page though for basic usage of the API? To me, that'd make more sense than essentially copying/pasting the same examples across clouds.

holmanb (approved):
There are some minor conflicts to resolve, but otherwise this looks good.

I tested this datasource and can confirm that it works as expected. This will be useful for testing from templates, for which a process to create new ones has been documented.

Feel free to merge once the conflicts are fixed.

pycloudlib/vmware/cloud.py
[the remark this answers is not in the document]
blackboxsw: sounds good. ship it

Issue #299, opened 2023-04-12
pycloudlib.azure.util.get_client credential instantiation


jrtknauer:
For every Azure SDK resource client instantiated in `pycloudlib`, the `get_client`  util function is responsible for creating and passing the credential to the client for authentication, with the following implementation:

```python
def get_client(resource, config_dict: dict):
    """Get azure client based on the give resource.

    This method will first verify if we can get the client
    by using the information provided on the login account
    of the user machine. If the user is not logged into Azure,
    we will try to get the client from the ids given by the
    user to this class.

    Args:
        resource: Azure Resource, An Azure resource that we want to get
                  a client for.
        config_dict: dict, Id parameters passed by the user to this class.

    Returns:
        The client for the resource passed as parameter.

    """
    if not util.subp("command -v az", shell=True, rcs=[0, 1, 127]).ok:
        logger.debug(
            "No azure-cli 'az' command found. Trying explicit config params"
        )
    else:
        try:
            cli_credential = AzureCliCredential()
            subscription_id = config_dict.get("subscriptionId")
            client = resource(cli_credential, subscription_id=subscription_id)
            return client
        except CLIError:
            logger.debug(
                "No valid azure-cli config found."
                " Trying explicit config params"
            )
        except ClientAuthenticationError:
            logger.debug(
                "Authentication error: No valid azure-cli config found."
                " Trying explicit config params"
            )

    required_keys = frozenset(
        {"clientId", "clientSecret", "tenantId", "subscriptionId"}
    )
    missing_keys = required_keys.difference(set(config_dict.keys()))
    if missing_keys:
        raise CloudSetupError(
            "No AZ cli config found, missing required keys: {}".format(
                ", ".join(missing_keys)
            )
        )

    parameters = {
        "active_directory": "https://login.microsoftonline.com",
        "resource_manager": "https://management.azure.com/",
        "active_directory_graph_resource_id": "https://graph.windows.net/",
        "sql_management": "https://management.core.windows.net:8443/",
        "gallery": "https://gallery.azure.com/",
        "management": "https://management.core.windows.net/",
    }
    parameters.update(config_dict)
    credential = ClientSecretCredential(
        tenant_id=parameters["tenantId"],
        client_id=parameters["clientId"],
        client_secret=parameters["clientSecret"],
        authority=parameters["active_directory"],
    )

    return resource(
        credential, subscription_id=parameters["subscriptionId"], **parameters
    )
```

There are two underlying issues with the current implementation:

[...]

Issue #232, opened 2023-01-13
Azure: when az is available the sdk expects we logged in via `az login`


paride:
Sample failure:

```
00:04:56.030              if "az login" in ex.stderr or "az account set" in ex.stderr:
00:04:56.030  >               raise CredentialUnavailableError(message=NOT_LOGGED_IN)
00:04:56.030  E               azure.identity._exceptions.CredentialUnavailableError: Please run 'az login' to set up an account
```

Pycloudlib should always rely on `pycloudlib.conf`. Maybe we can drop this `if`? https://github.com/canonical/pycloudlib/blob/e7eefbe029a7b61be643c8e0a1e8cc740f54aefb/pycloudlib/azure/util.py#L37

PR #301, opened 2023-04-14
fix(azure): remove azure cli support

jrtknauer:
Historically, pycloudlib has supported both az cli and Azure App Registration based authentication. The internal implementation of the azure get_client util function does not correctly implement support for both authentication options, specifically in the case where az cli is installed but not logged in - pycloudlib will still try to use az cli for authentication even if App Registration credentials are provided.

pycloudlib should discontinue all support for az cli based authentication, as it creates an unnecessary dependency which must be maintained.

The extra parameters which are passed to the resource client init have also been removed as they are vestigial kwargs required for the azure.common.client_factory.get_client_from_json_dict interface.


jrtknauer:
Opened this PR to start discussion related to previously reported issues:
- https://github.com/canonical/pycloudlib/issues/299
- https://github.com/canonical/pycloudlib/issues/232

aciba90 (commented):
Thanks, @jrtknauer, for bringing up this discussion.

I believe this is the direction we should follow in the future.

I have checked https://github.com/canonical/server-jenkins-jobs/ and the only uses of `az login` are in the context of stand-alone scripts, not before `pycloudlib` is used. Thus, this change does not require any modification in `server-jenkins-jobs`, right @paride ?

pycloudlib/azure/util.py
```diff
@@ -35,57 +32,25 @@ def get_client(resource, config_dict: dict):
...
     missing_keys = required_keys.difference(set(config_dict.keys()))
     if missing_keys:
         raise CloudSetupError(
-            "No AZ cli config found, missing required keys: {}".format(
+            "Missing required Azure credentials: {}".format(
                 ", ".join(missing_keys)
             )
         )
 
-    parameters = {
```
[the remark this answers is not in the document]
TheRealFalcon: Instead of reverting if there's a problem, can we check Xenial to see if removing this works? AFAIK, Pro is still using pycloudlib test things on Xenial, so as long as it works there, we should be ok.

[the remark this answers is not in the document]
aciba90: The minimal Ubuntu release used as host in pro tests is focal: https://github.com/canonical/ubuntu-pro-client/blob/b71a935df6ea8bf23d5ec2988941f2037cba0846/.github/workflows/ci-integration.yaml#L75

I have tested this on a bionic host, fixing the azure-related dependencies to the minimum ones, and it works.

pycloudlib/azure/util.py
```diff
@@ -35,57 +32,25 @@ def get_client(resource, config_dict: dict):
...
-    parameters.update(config_dict)
     credential = ClientSecretCredential(
-        tenant_id=parameters["tenantId"],
-        client_id=parameters["clientId"],
-        client_secret=parameters["clientSecret"],
-        authority=parameters["active_directory"],
+        tenant_id=config_dict["tenantId"],
+        client_id=config_dict["clientId"],
+        client_secret=config_dict["clientSecret"],
+        authority=config_dict["active_directory"],
```
aciba90, line 50: `authority` is not anymore a required argument.

Assuming there is not a strong use case to support different authorities than the default (https://login.microsoftonline.com), we could remove it and make it configurable in the future if needed:

```suggestion
```

If not, we must either provide a default value or make it a required argument in `Azure.__init__`.

aciba90 (changes requested):
I had success testing this on a bionic host.

From my point of view, this could be merged after the inline comment about `authority` is fixed. Are you, @jrtknauer, okay driving this?

pycloudlib/azure/util.py
jrtknauer: My mistake - I think that slipped through as I was copy/pasting things around.

jrtknauer:
> I had success testing this on a bionic host.
> 
> From my point of view, this could be merged after the inline comment about `authority` is fixed. Are you, @jrtknauer, okay driving this?

Final amendments have been made and pushed, just waiting for CI to run through and verify.

aciba90 (approved):
LGTM, thanks!

VERSION
```diff
@@ -1 +1 @@
-1!2.1.3
+1!2.1.4
```
aciba90, line 1: Taking into account this is technically a breaking change, could you please bump the major version instead?

```suggestion
1!3.0.0
```

PR #302, opened 2023-04-19
Properly pass through distro usernames, don't assume ubuntu.

holmanb:
Currently azure and ec2 assume username "ubuntu", which breaks other distributions using pycloudlib. Fix it.


aciba90 (approved):
+1, thanks!

Issue #262, opened 2023-02-08
Pycloudlib does not cleanup on failed launch


hammerstefan:
Currently pycloudlib makes no attempt to clean up cloud resources on failure.  This can be confusing and financially dangerous for the end-user, if they are not aware they may need to manually clean up.  

Resources that might be orphaned:
* Running instances
* VPCs
* Resource Groups

In our recent integration with the pycloudlib OCI support, we ended up with roughly 20 orphaned instances per week.

**Current workaround**: 
Make sure to pass `wait=False` into every pycloudlib command, call `instance.wait()` directly, catch all exceptions, on exception attempt to `instance.delete()`

```py
cloud: pycloudlib.oci.cloud.OCI
try:
    instance = cloud.launch(wait=False)
    instance.wait()
except Exception:
    try:
        instance.delete(wait=False)
        instance.wait()
    except Exception:
        # ORPHANED INSTANCE MAYBE
        pass
```
This has a few problems:
    * Needing to call wait() separately even though we really just want start(wait=True) - we can't proceed until instance is running
    * Needing to wrap in two try/except blocks when none are desired
    * Have no idea if delete() failed because instance didn't exist to begin with, or if it failed to communicate with running instance and we still have an orphaned instance
    * `wait=True` is the default behavior, which means pycloudlib is orphaning tens of instances per week BY DEFAULT in our CD pipeline
    * pycloudlib is throwing poorly typed exceptions, requiring user to catch `Exception` class, which is always bad and fraught with errors (since you don't actually know WHAT caused an error)

Improvement options:

**Option 1**: Update documentation only

If this current behavior is intentional, then documentation and error messages should be improved to clearly reflect that.  The user must be aware during which steps resources could be orphaned.  When an error is raised, the error should indicate if resources are left behind, and those resource IDs should be reported.  Documentation should also contain how to use `wait=False` and manual `delete()` for automation use.

Pro: 
* Documentation improvements are generally good
* Least disruptive
Con:
* Still have unexpected, potentially costly default behavior
* Needing to add boilerplate wrappers to pycloudlib interaction 

**Option 2**: Automate cleanup on error, opt-in

Provide the user with a way to automate cleanup on lifecycle interaction error (start, delete, restart, shutdown). This could be runtime configuration to indicate pycloudlib should fully manage the instance lifecycle, and not leave orphaned instances.

Pro:
* Doesn't change existing default behavior
* Acceptable solution for pipelines where we never want orphans
[...]

PR #305, opened 2023-05-19
API CHANGE: Allow for launch cleanup

TheRealFalcon:
```
API CHANGE: Allow for launch cleanup

This commit removes the `wait` parameter from launch calls and allows
all instance classes to be used as context managers.

When calling `launch(wait=True)`, callers have no mechanism to cleanup
instances that fail on launch because a handle to the created instance
hasn't been returned yet. To remedy this, always return the created
instance immediately and force the caller to call wait() manually.

Additionally, instance classes have been converted into context
managers so that instances can be launched with cleanup being
handled automatically.

Fixes: #262
```


TheRealFalcon:
This is one possible approach of fixing #262 . Any feedback is welcome!

pycloudlib/cloud.py
```diff
@@ -137,7 +137,6 @@ def launch(
         image_id: str,
         instance_type=None,
         user_data=None,
-        wait: bool = True,
```
blackboxsw, line 140: Drop wait from docstr too

pycloudlib/instance.py
```diff
@@ -41,6 +41,12 @@ def __init__(self, key_pair, username: Optional[str] = None):
         self.connect_timeout = 60
         self.banner_timeout = 60
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, _type, _value, _traceback):
+        self.delete()
```
blackboxsw, line 48: This needs try/except: pass gating as there may be integration tests that already manually delete and instance before exiting the context mgr.

blackboxsw (commented):
This looks good for handling instances specifically. What do you think about using the BaseCloud itself as a contextmanager?
We could persist/append any created resource to  BaseCloud.resources and on __exit__ from the BaseCloud instance we walk self.resources calling `delete` if available?
That'd allow us to track any VPC, resource groups that pycloudlib creates on our behalf as a side-effect of cloud.launch and allow removal of those resources when the test is over.

pycloudlib/instance.py
blackboxsw: flake will tell you you need docstrs on these

tests/integration_tests/test_public_api.py
```diff
@@ -126,21 +114,25 @@ def test_public_api(cloud: BaseCloud):
...
+                image_id=snapshot_id, user_data=cloud_config
+            ) as instance_from_snapshot:
+                instance_from_snapshot.wait()
+                exercise_instance(instance_from_snapshot)
+        finally:
+            cloud.delete_image(snapshot_id)
     finally:
-        cloud.delete_image(snapshot_id)
+        if isinstance(cloud, pycloudlib.Azure):
+            cloud.delete_resource_group()
```
blackboxsw, line 138: This and the `finally:  cloud.delete_image(snapshot_id)` above could be automatically removed if our BaseCloud were used a context manager that collected any resources created during that test run and purged those resources on `__exit__` from the context.  We'd probably want to also create a public method `BaseCloud.release_resources`  or `delete_resources` or something to allow folks not using a context manager to manually invoke this at the end of any tests they are running anyway.

TheRealFalcon:
This grew to be a lot bigger than I originally expected...
```
1. Fix unit tests
2. Add a 'clean' method/context manager to cloud API that cleans up all created resources during life of the instance.
3. Ensure exceptions aren't thrown when cloud.clean()/instance.delete() are called multiple times for the same instance.
4. Collect any exceptions raised during clean/delete and return them to caller. This is so that an exception can't interrupt other cleanup that would otherwise happen. Context managers will print these to stderr.
```

Note that this is still untested on openstack as I need to get my openstack creds back before I can test there.

pycloudlib/ec2/instance.py
```diff
@@ -355,24 +374,34 @@ def remove_network_interface(self, ip_address):
...
+            )
+            return
+
+        # Detach from the instance
         self._client.detach_network_interface(
             AttachmentId=nic.attachment["AttachmentId"]
         )
 
-        # Detach from the instance
+        # Wait for detach
```
blackboxsw, line 396:
```suggestion
        # Wait for detach of all NICs matching the provided ip_address
```

pycloudlib/ec2/instance.py
```diff
@@ -355,24 +374,34 @@ def remove_network_interface(self, ip_address):
...
+                for nic in self._instance.network_interfaces
+                if nic.private_ip_address == ip_address
+            ),
+            None,
+        )
+        if not nic:
+            self._log.debug(
+                "Not deleting NIC because no NIC with IP {} found."
+            )
+            return
```
blackboxsw, line 389: Is the reason we aren't realizing the full list of `nics_matching_ips` outside of the for loop below because we are expecting our self._instance.reload() to show the dwindling list of matching NICs over a period of loop iterations?

pycloudlib/ec2/instance.py
```diff
@@ -231,17 +246,21 @@ def _attach_network_interface(self, interface_id: str) -> str:
...
+        # It's possible the attach worked correctly but it's not yet
+        # present in the nic attributes
+        for _ in range(5):
+            self._instance.reload()
+            for nic in self._instance.network_interfaces:
+                if nic.attachment["AttachmentId"] == response["AttachmentId"]:
+                    nic.modify_attribute(
+                        Attachment={
+                            "AttachmentId": response["AttachmentId"],
+                            "DeleteOnTermination": True,
```
blackboxsw, line 258: Doesn't  this [`DeleteOnTermination`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/modify_network_interface_attribute.html#EC2.Client.modify_network_interface_attribute) mean we don't have to worry about memoizing created_interfaces on AWS?

pycloudlib/instance.py
```diff
@@ -41,6 +41,15 @@ def __init__(self, key_pair, username: Optional[str] = None):
...
         self.banner_timeout = 60
 
+    def __enter__(self):
+        """Enter context manager for this class."""
+        return self
+
+    def __exit__(self, _type, _value, _traceback):
+        """Exit context manager for this class."""
+        exceptions = self.delete()
+        print_exception_list(exceptions)
```
blackboxsw, line 51: Feels like we do not want a sys.exit(1) on each instance deletion in the cloud, but probably something we need to track when `cloud.clean` is called though.

pycloudlib/azure/cloud.py
```diff
@@ -937,11 +948,32 @@ def snapshot(
...
+        If no resource group is provided, delete self.resource_group
+        """
+        if resource_group_name is None and self.resource_group:
+            resource_group_name = self.resource_group.name
             self.resource_group = None
+        if resource_group_name:
+            with contextlib.suppress(ResourceNotFoundError):
+                self.resource_client.resource_groups.begin_delete(
+                    resource_group_name=resource_group_name
+                )
```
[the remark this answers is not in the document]
TheRealFalcon: 1. Probably. I wasn't aware of a delete poller, so I'll look into that.
2. On line 956, we only use self.resource_group if we didn't get a resource group name passed in. The clean method passes over our created resource groups and does that.

pycloudlib/azure/cloud.py
```diff
@@ -937,11 +948,32 @@ def snapshot(
...
+    # pylint: disable=broad-except
+    def clean(self) -> List[Exception]:
+        """Cleanup ALL artifacts associated with this Cloud instance.
+
+        This includes all instances, snapshots, resources, etc.
+        To ensure cleanup isn't interrupted, any exceptions raised during
+        cleanup operations will be collected and returned.
+        """
+        exceptions = super().clean()
+        for resource_group in self.created_resource_groups:
```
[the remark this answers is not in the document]
TheRealFalcon: Some of this is a bit tacked on as I didn't want to do broader class refactors at the same time. If you look at `_create_resource_group()`, if the resource group under the given tag already exists, we'll use it, otherwise we'll create it and add it to `self.created_resource_groups` while also making it `self.resource_group`, so I don't think clean will do nothing here unless the resource group already existed.

pycloudlib/cloud.py
```diff
@@ -67,8 +74,17 @@ def __init__(
...
             self.tag = validate_tag(tag)
 
+    def __enter__(self):
+        """Enter context manager for this class."""
+        return self
+
+    def __exit__(self, _type, _value, _trackback):
+        """Cleanup context manager for this class."""
+        exceptions = self.clean()
+        print_exception_list(exceptions)
```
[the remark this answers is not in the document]
TheRealFalcon: Good question. My initial thinking is I don't think clean failures need to be catastrophic, but I could also see the case where something is broken and you don't want to keep going. I think it'd be more common that a cleanup failure is something minor, like timeout before resource is gone, so I think it's better to continue rather than exit.

You have any strong opinions?

pycloudlib/ec2/cloud.py
```diff
@@ -269,7 +274,11 @@ def delete_image(self, image_id, **kwargs):
             image_id: string, id of the image to delete
         """
         image = self.resource.Image(image_id)
-        snapshot_id = image.block_device_mappings[0]["Ebs"]["SnapshotId"]
+        try:
+            snapshot_id = image.block_device_mappings[0]["Ebs"]["SnapshotId"]
+        except AttributeError:
+            self._log.debug("Not deleting image as %s not found.", image_id)
```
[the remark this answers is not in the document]
TheRealFalcon: I kept going back and forth on whether to log or not. I started out by logging and then decided to not, but missed this case. I can remove it, but can also add other logs if you think it is valuable.

pycloudlib/ec2/instance.py
TheRealFalcon: You'd think so, but I kept seeing NICs sticking around. I put a comment about it in the `delete()` method.

pycloudlib/ec2/instance.py
TheRealFalcon: We're not expecting a single IP to be attached to multiple NICs. I'm not sure if that's even possible, but the way the add_network_interface() code works, it should be 1:1. We loop through all of the nics (via list comprehension) because that's the only way to know if we have a nic that matches this particular IP, but we're only ever expecting 1 result from the loop.

Does that make sense? Do you think there's a way to make it clearer?

pycloudlib/instance.py
TheRealFalcon: Track how?

TheRealFalcon:
@blackboxsw I believe I addressed all of your comments that don't require more discussion.

pycloudlib/cloud.py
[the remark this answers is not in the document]
blackboxsw: I think we definitely should exit(1) here. As BEnjamin's comment on the original issue was:
` Regarding overall verbosity of different implementations, I care less about number of lines, and care more about the chance of accidentally leaking / orphaning cloud resources, especially with default parameters.`

pycloudlib/ec2/instance.py
blackboxsw: Nah, it's just the same boilerplate code used a few lines below. Maybe separate that to a private function `_get_nic_for_ip` and call it here and within the for loop?

pycloudlib/instance.py
blackboxsw: Sorry. 'track' is a bad word. I want to be noisy if we fail on instance or image deletion or disk deletion as those are costly things to keep around on some clouds.  The [sys.exit(1) was the intent of my comment above to "track" by raising an error via non-zero exit in pycloudlib/cloud.py on __exit__ if teardown/deletion of some resources failed at the end of a test suite](https://github.com/canonical/pycloudlib/pull/305/files#r1248110230).

pycloudlib/cloud.py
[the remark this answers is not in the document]
TheRealFalcon: Rather than exit(1), I think we raise a separate custom exception instead. Something like `CleanupException`.

exit(1) is a pretty broad hammer that gives nobody the option to recover.

Sound good?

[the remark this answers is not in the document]
blackboxsw: Much better.

TheRealFalcon:
@blackboxsw I addressed your comments and rebased the branch.

blackboxsw (approved):
Perfect, validated failure behavior injecting a zerodiv error in instance.delete on lxd. LGTM

pycloudlib/azure/instance.py
```diff
@@ -136,13 +137,27 @@ def _do_restart(self, **kwargs):
...
-        delete = self._client.virtual_machines.begin_delete(
-            resource_group_name=self._instance["rg_name"], vm_name=self.name
-        )
-
-        if wait:
-            delete.wait()
+        if self.status == "deleted":
+            return []
+        try:
+            poller = self._client.virtual_machines.begin_delete(
```
CalvoM, line 146: What if we delete the resource group instead of the vm, since it will delete every other resource allocated in the same resource group?

TheRealFalcon:
@hammerstefan or @mjd-x , since you have both been involved in discussions around this in the past, I want you to have a chance to review or raise any concerns before this gets merged. Will this work for CPC?

mjd-x (approved):
lgtm, thanks! If possible, it'd be nice to see this reflected in the documentation. It should be clear that users are responsible for doing the cleanup for any instance launched, even if errors were raised (for example, even if PycloudlibTimeoutError is raised, the instance could still be running in the cloud)

docs/resource_cleanup.md
```diff
@@ -0,0 +1,42 @@
...
+    with cloud.launch("your-ami") as instance:
+        output = instance.execute("cat /etc/lsb-release").stdout
+
+
+print(output)
+```
+
+When the context manager exits (even if due to an exception), all resources that were created during the lifetime of the `Cloud` or `Instance` object will automatically be cleaned up. Any exceptions raised during the cleanup process will be raised.
+
+Alternatively, if you don't want to use context managers, you can manually cleanup all resources using the `.clean()` method on `Cloud` objects and the `.delete()` method on `Instance` objects. For exampe, using EC2:
```
blackboxsw, line 23:
```suggestion
Alternatively, if you don't want to use context managers, you can manually cleanup all resources using the `.clean()` method on `Cloud` objects and the `.delete()` method on `Instance` objects. For example, using EC2:
```

blackboxsw (approved):
minor doc typo. example looks good +1 on this.

PR #310, opened 2023-06-01
Azure urn version fix

jrtknauer:
This will be a breaking change for users who are relying on the implicit "latest" for Marketplace URNs in hardcoded scripts, but I believe it's important that `pycloudlib` does not mask a well defined image definition pattern (the Marketplace URN).

The parsing of the image_id could also be simplified to use `str.split()` - I do believe a regex is overkill here as a Marketplace URN is guaranteed to be `<publisher>:<offer>:<sku>:<version>`, but the change itself is not necessary within the scope of this change. What is important for this PR is that users who specify Marketplace URNs with versions other than latest actually launch that version.


aciba90 (changes requested):
LGTM, thanks.

The only thing it that being this a breaking change, I think we should bump the major part of the version.

aciba90 (approved):

Issue #298, opened 2023-04-03
Azure launch VM configuration interfaces for confidential VM and trusted launch


jrtknauer:
While setting these configurations is possible through the vm_params passthrough on the launch method, it would be a beneficial enhancement for pycloudlib to directly support Confidential VM and Trusted Launch configuration as part of the Azure client interface.

As an example, for enabling trusted launch the VM configuration under the current interface, 
```
client = pycloudlib.Azure(...)
trusted_launch_config = {
    "security_profile": {
        "security_type": "TrustedLaunch"
        "uefi_settings": { "secure_boot_enabled": True, "v_tpm_enabled": True }
    }
}
client.launch(..., vm_params=trusted_launch_config)
```

Ideally, the pycloudlib Azure VM configuration interface would match that of the portal, abstracting feature toggles from direct passage of vm_params to encoded enums/flags.

aciba90:
Tracking in SC-1533.

PR #311, opened 2023-06-13
feat(azure): nic configuration overrides

jrtknauer:
Our use case requires enablement of accelerated networking on the NIC, so implementing a generic passthrough interface similar to vm_params. It would be ideal to have explicit support (as proposed by other enhancements like Issue #298).


aciba90 (commented):
LGTM. Thanks, @jrtknauer, for the improvements. Being this new feature, it feels like we should bump the minor version, right?

jrtknauer:
Version bumped accordingly. Should be good once CI passes.

aciba90 (approved):
+1, thanks!

PR #313, opened 2023-07-05
feat(azure): supports cvm and trusted launch

CalvoM:
Azure can launch a confidential vm and trusted launch instead of via vm_params


rpocase:
Anything we can do to get some test coverage on this? Full code path might not be easy, but just the new strategy generation for reducing scope

pycloudlib/azure/security_types.py
```diff
@@ -0,0 +1,136 @@
...
+            vm_params: dict, The parameters passed to Azure
+            **kwargs: dictionary of security type parameters
+        """
+        raise NotImplementedError
+
+    def is_correct_vm_size(self, vm_size: str) -> bool:
+        """Check if the vm size is supported by security type.
+
+        Args:
+            vm_size: string, The
```
aciba90, line 36: What is the format / units of `vm_size` ?

pycloudlib/azure/security_types.py
```diff
@@ -0,0 +1,136 @@
...
+            },
+        }
+        util.update_nested(vm_params, param_update)
+        return vm_params
+
+    def is_correct_vm_size(self, vm_size: str) -> bool:
+        """Check if the vm size is supported by security type.
+
+        Args:
+            vm_size: string, The
```
aciba90, line 131: Fix doc please.

pycloudlib/azure/security_types.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        size = re.search(r"[DE]C\d+a[d]*s_v5", vm_size)
```
aciba90, line 133: Could we add some example / unit test covering / showcasing this ?

pycloudlib/azure/security_types.py
```diff
@@ -0,0 +1,136 @@
...
+    VM_GUEST_STATE_ONLY = "VMGuestStateOnly"
+    DISK_WITH_VM_GUEST_STATE = "DiskWithVMGuestState"
+
+
+class SecurityType(abc.ABC):
+    """Base Azure Security Type Class."""
+
+    @abc.abstractmethod
+    def configure_vm_params(
+        self, vm_params: Dict[str, Any], **kwargs
```
aciba90, line 22: As of now, the only kwarg that is used is `os_disk_encryption` in `AzureConfidentialVMSecurityType`. What about being explicit in the interface and keep it simple, doing something in the following lines:

```suggestion
        self, vm_params: Dict[str, Any], os_disk_encryption: Optional[AzureCVMOSDiskEncryption] = None
```

pycloudlib/azure/cloud.py
```diff
@@ -685,13 +716,30 @@ def launch(
                 "Found network interface: %s. Reusing it", nic.name
             )
 
+        vm_params = kwargs.get("vm_params", {})
+        sec_type = security_type.value()
+        sec_type.configure_vm_params(
+            vm_params=vm_params,
+            **kwargs.get("security_type_params"),
```
aciba90, line 723: If `security_type_params` is not in kwargs, then we would try to expand a `**None` which is invalid python.

```suggestion
            **kwargs.get("security_type_params", {}),
```

Probably worth unit-testing this? What do you think?

pycloudlib/azure/cloud.py
```diff
@@ -685,13 +716,30 @@ def launch(
...
+        )
+        if sec_type.is_correct_vm_size(instance_type) is False:
```
aciba90, line 725: Nit:

```suggestion
        if not sec_type.is_correct_vm_size(instance_type):
```

aciba90 (changes requested):
Thanks, @CalvoM.

I have manually tested this and been able to spawn focal instances with standard, trusted_launch and cvm security types.

I have left some minor comments but in general looks great.

I think minimally we should add some level of unit-testing covering the new interfaces.

pycloudlib/azure/security_types.py
CalvoM: The `vm_size` is a string

pycloudlib/azure/security_types.py
CalvoM: Added.

pycloudlib/azure/security_types.py
rpocase: What are the valid inputs for vm_size? Its a string, but a VERY specific string that has different meanings. E.g., it could actually be a `CloudImageSize` which is a string that must be formatted as `integerSizeIdentifer` (e.g. 8gb).

CalvoM: After discussion on standup, it is better not to include this check and let the user provide a valid vm_size. Furthermore, Azure has a good error message incase of an unsupported vm size.

pycloudlib/azure/security_types.py
CalvoM: We will not need this since we do not check the vm size.

pycloudlib/azure/security_types.py
CalvoM: I agree but I think we will  need this `kwargs` later for any change in the security types configuration, considering MS api changes a lot.

aciba90 (changes requested):
Thanks, @CalvoM. I have left inline comments.

Additionally, we have to bump the [version](https://github.com/canonical/pycloudlib/blob/main/VERSION).

pycloudlib/azure/security_types.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class AzureCVMOSDiskEncryption(Enum):
+    """Represents Azure OS disk encryption types."""
+
+    VM_GUEST_STATE_ONLY = "VMGuestStateOnly"
+    DISK_WITH_VM_GUEST_STATE = "DiskWithVMGuestState"
+
+
+class SecurityType(abc.ABC):
```
aciba90, line 17: Nit: After removing `is_correct_vm_size` this class hierarchy / interface has only one method and a static attribute. Would not be simpler to encode this as a function? Like:

```py
def configure_security_vm_params(azure_security_type, vm_params, os_disk_enc):
    ...
```

CalvoM: After our discussion, and abit of research, I see this will make sense for our current condition. Incase the security types will need more work, we can switch back to classes.

VERSION
```diff
@@ -1 +1 @@
-1!4.1.1
+1!4.1.2
```
aciba90, line 1: This library follows semantic versioning, it feels like we are adding feature and we should bump the minor part:

```suggestion
1!4.2.1
```

PR #314, opened 2023-08-06
Add username everywhere

TheRealFalcon:
#302 added 'username' to Azure and EC2. Add it everywhere else.

## Additional Context
At the summit, I'd like to be able to say "pycloudlib supports other distros" unconditionally :smile:


TheRealFalcon:
There's also a number of "random" typing updates specified in a separate commit. I'm not sure why, but mypy has an annoying habit of complaining about things that seem completely unrelated after you've added new type definitions.

pycloudlib/azure/cloud.py
```diff
@@ -887,7 +888,14 @@ def _retrive_instance_ip(self, instance):
...
         return self._retrieve_ip_from_network_interface(nic=instance_nic)
 
-    def get_instance(self, instance_id, search_all=False):
+    def get_instance(
+        self,
+        instance_id,
+        search_all=False,
+        *,
+        username: Optional[str] = None,
+        **kwargs,
```
holmanb, line 897: `kwargs` is not passed anywhere. If this is needed for signature compatibility, do you think we should add a comment in the docstring to mention why?

pycloudlib/azure/cloud.py
```diff
@@ -896,6 +904,7 @@ def get_instance(self, instance_id, search_all=False):
                         for the instance in the entire reach of the
                         subsctription id. If false, we will search only
                         in the resource group created by this instance.
+            username: username to use when connecting via SSH
```
holmanb, line 907: format?

pycloudlib/ec2/cloud.py
```diff
@@ -294,26 +294,32 @@ def delete_key(self, name):
         self._log.debug("deleting SSH key %s", name)
         self.client.delete_key_pair(KeyName=name)
 
-    def get_instance(self, instance_id):
+    def get_instance(
+        self, instance_id, *, username: Optional[str] = None, **kwargs
```
holmanb, line 298: comment for unused `kwargs`?

holmanb (approved):
I left a couple of nits inline. This looks good to me.

pycloudlib/azure/cloud.py
holmanb: My bad, yes then that should be fine

PR #315, opened 2023-08-09
Oracle allow explicitly specifying vcn for subnet selection

a-dubs:
feat(oracle): ability to specify vcn by name for subnet selection

Oracle's pycloudlib implementation currently uses the most recently created VCN (Virtual Cloud Network) in the given compartment for new VM instances being created. Problems can therefore arise if a newly created VCN uses a subnet with different rules than the previously used VCN's subnet.

Now, instead of relying on just blindly choosing the most recently created VCN, pycloudlib supports specifying a vcn_name value which it will use to find a VCN that has a name that exactly matches the given vcn_name.


CalvoM (commented):
LGTM, but I need to do some testing for the functionality.

pycloudlib/oci/utils.py
```diff
@@ -47,13 +48,27 @@ def get_subnet_id(
         network_client: Instance of VirtualNetworkClient.
         compartment_id: Compartment where the subnet has to belong
         availability_domain: Domain to look for subnet id in.
+        target_subnet_name: Name of the subnet to look for.
```
holmanb, line 51: Is this line needed?

CalvoM (approved):
Tested it with CTF and it works. LGTM.

pycloudlib/oci/utils.py
```diff
@@ -37,6 +37,7 @@ def get_subnet_id(
     network_client: "oci.core.VirtualNetworkClient",  # type: ignore
     compartment_id: str,
     availability_domain: str,
+    vcn_name: str = None,
```
aciba90, line 40:
```suggestion
    vcn_name: Optional[str] = None,
```

And the appropriate import if required.

aciba90 (changes requested):
Thanks! Apart from the typing fix, LGTM.

pycloudlib/oci/utils.py
a-dubs: Fixed, tysm :D

a-dubs:
Jira ticket for context: https://warthogs.atlassian.net/browse/CPC-2908

PR #317, opened 2023-08-25
ec2: mantic image names now include disk type hvm-ssd-gp3

blackboxsw:
Daily development images for mantic and later now include the disk storage type -gp3 in the encoded image names.

Provide this disk-type to boto image filters when searching for daily Ubuntu images published by Canonical.


holmanb (approved):
Looks good, thanks for this fix!

PR #321, opened 2023-10-06
Move all unit tests into unit_tests directory

TheRealFalcon:
Somehow most of the unit tests got missed in the previous migration


blackboxsw (commented):
Do we also need to change out testenv:pytest config to specifically provide tests/unit_tests directory to default `tox -e pytest` will report the 110 tests run from the unit_tests directory instead of 3 in the root dir?

```diff
diff --git a/tox.ini b/tox.ini
index 767f304..a177cad 100644
--- a/tox.ini
+++ b/tox.ini
@@ -21,7 +21,7 @@ deps =
 [testenv:pytest]
 envdir = {[common]envdir}
 deps = {[common]deps}
-commands = {envpython} -m pytest --doctest-modules {posargs:--cov pycloudlib -- pycloudlib}
+commands = {envpython} -m pytest --doctest-modules {posargs:--cov pycloudlib -- pycloudlib tests/unit_tests}
 
 [testenv:mypy]
 envdir = {[common]envdir}
```

Without this diff, `tox -e pytest` will only run 3 tests instead of the 110 in our tests/unit_tests dir.

TheRealFalcon:
@blackboxsw , thanks! I had updated the pytest defaults previously, but forgot about tox. Since the pytest defaults already do the right thing, I switched to using that rather than providing independent args.

blackboxsw (approved):

PR #322, opened 2023-10-16
Pin openstacksdk

TheRealFalcon:
Latest versions include 'Literal' which doesn't work with 3.7


aciba90 (approved):
LGTM, thanks.

Upstream bug: https://storyboard.openstack.org/#!/story/2010903

PR #323, opened 2023-10-16
feat: Add QEMU datasource

TheRealFalcon:


tests/unit_tests/test_instance.py
```diff
@@ -56,7 +56,7 @@ def test_wait_execute_failure(
     ):
         """Test wait calls when execute command fails."""
         instance = concrete_instance_cls(key_pair=None)
-        m_time.side_effect = [1, 2, 40 * 60, 40 * 60 + 1]
+        m_time.side_effect = [1, 1, 2, 40 * 60, 40 * 60 + 1]
```
TheRealFalcon, line 59: Changes in this file are due to additional logging calls calling a time.time()

aciba90 (approved):
LGTM, thanks for the great work here!

I left some inline concerns that shouldn't probably block this PR. The most important is that the utility function, `Qemu.released_image`, does not operate with released images.

pycloudlib/qemu/cloud.py
```diff
@@ -0,0 +1,633 @@
...
+            qemu_args.extend(launch_args)
+
+        self._log.info("Launching qemu with args: %s", qemu_args)
+        handle = subprocess.Popen(  # pylint: disable=R1732
+            qemu_args,
+            stdin=subprocess.DEVNULL,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+        instance = QemuInstance(
```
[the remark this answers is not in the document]
TheRealFalcon: Isn't this true of all of our clouds? `delete()` has to be called (either manually or via context manager) or the resource is leaked. Do you think that it's a bigger problem here?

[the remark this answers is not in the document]
aciba90: I was thinking in terms of OS resources. `Popen` could leak file descriptors if `kill` / `close` is not called. So imagine we have a long-running process creating `QemuInstance`s but not calling `QemuInstance.delete`, if those instances get out of scope and garbage collected, then the qemu process created here is going to be leaked I think.

I was suggesting to clean up this process like:

```py
class QemuInstance:
    def __del__(self):
        self.handle.close()
```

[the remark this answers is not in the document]
aciba90: So, the difference with other clouds is that in this case the parent process could leak memory/fds and exhaust the host, but maybe I am overthinking here. What do you think?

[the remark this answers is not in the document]
TheRealFalcon: Yeah, I see your point. It'll definitely be a lot easier to overwhelm a host with VMs vs the cost of leaking cloud instances. I'd be most worried about RAM as using DEVNULL means we won't be allocating new fds every invocation.

On the other side of the argument though, I would like to be able to use the `KEEP_INSTANCE = True` functionality that we have in cloud-init to debug instances after the fact. In this case, it would make sense to let processes live even longer than they currently do (i.e., past the life of the pycloudlib process).

I think there are trade-offs both ways. I'll bring it up at standup.

docs/clouds/qemu.md
```diff
@@ -0,0 +1,104 @@
...
+`qemu-system-x86_64` on an x86-64 host.
+
+## Requirements
+
+In addition to requirements specified in `setup.py`, on Ubuntu systems,
+the QEMU cloud requires the following packages:
+
+* qemu-system-x86
+* qemu-utils  # for qemu-img
+* genisoimage  # to create the cloud-init NoCloud seed image
```
blackboxsw, line 17: Not for this PR: 
I like that we are building in supplemental utilities necessary to setup and launch vm, i wonder longer-term if we want to support use-cases where folks want to customize these images using `mount-image-callback` or `virt-customize` prior to initial launch of those images. Surely, these preparatory steps can be performed outside pycloudlib and fed into the `Qemu.launch(image_id="/my/customized/image")`, so probably doesn't make sense at this point. But, seeing our helper methods creating iso images, customizing base images before boot is a frequently used feature when dealing with qemu that we may want to offer in the future if possible.

pycloudlib/qemu/cloud.py
```diff
@@ -0,0 +1,634 @@
...
+                "instance_type must be in the form of c#m#, not "
+                f"{instance_type}"
+            )
+        return int(cpu_mem["cpus"]), int(cpu_mem["memory"])
+
+    def _get_kernel_path(
+        self, kernel_path, kernel_cmdline: str, image_id, base_image
+    ) -> Optional[Path]:
+        kernel = None
+        if kernel_cmdline and not kernel_path:
```
blackboxsw, line 381: We are already performing this check outside this scope at [the call-site](https://github.com/canonical/pycloudlib/pull/323/files#diff-9c031a07867513f1651aa8c522f0ca475eed8b5756bafa4ebaa92966811fdb57R511). Do you want to leave this check in here and remove the one at the call-site and we can just set or return the original `kernel_path` from this method?

pycloudlib/qemu/cloud.py
```diff
@@ -0,0 +1,634 @@
...
+            default_kernel_args += "root=/dev/sda1 "
+        if "console=" not in kernel_cmdline:
+            default_kernel_args += "console=ttyS0 "
+        return default_kernel_args + kernel_cmdline
+
+    # pylint:disable=R0914
+    def launch(
+        self,
+        image_id: str,
+        instance_type="c1m1500",
```
blackboxsw, line 402: Encoding cpu and memory size is reasonable for this first pass. But, we may end up painting ourselves into a corner as far as customization of -smp as other parameters overrides may be desired which would modify sockets/cores/threads/maxcpus and I don't know off-hand how we'd support that.


Maybe we also need to provide an additional launch param `qemu_args: List[str]` that'd allow both extending the default arguments passed to qemu-system-x86_64 and or replacing the `-smp` params if qemu_args contains `-smp` directives.

pycloudlib/qemu/cloud.py
```diff
@@ -0,0 +1,634 @@
...
+        user_data=None,
+        meta_data=None,
+        vendor_data=None,
+        kernel_cmdline="",
+        kernel_path=None,
+        no_seed_iso=False,
+        username: Optional[str] = None,
+        launch_args: Optional[List[str]] = None,
```
blackboxsw, line 410: I do think it would be helpful to provide a supplmental `qemu_args: Optional[List[str]]`  to extend or manipulate the default qemu behavior. I know there are cases for the desktop images where I'd want to deal with `-daemonize` or `--nographic` depending on my test usecases.

pycloudlib/qemu/cloud.py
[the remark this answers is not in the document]
TheRealFalcon: As discussed in standup, we'll go ahead with letting the process live until explicitly killed.

docs/clouds/qemu.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Configuration and Local Storage
+
+QEMU instances require local directories to store images and instances.
+These directories can be configured in the `pycloudlib.toml` file:
+
+`image_dir` - Directory to look for images when launching instances.
```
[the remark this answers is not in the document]
TheRealFalcon: Can that be a separate refactoring PR when needed? I think each page would need its own parsing as the url, title, filename, kernel-name (if even included...it's not for desktop) are all different.

docs/clouds/qemu.md
```diff
@@ -0,0 +1,104 @@
...
+kernel command line. When specified, `kernel_path` must also be specified.
+The `daily_image` and `released_image` calls will automatically download
+the kernel for the image, so if these calls have been used to obtain the
+image_id, the `kernel_path` is not required.
+
+By default, pycloudlib will create a seed iso for the instance in order
+to let cloud-init configure the instance. If `no_seed_iso` is set to
+`True`, pycloudlib will not create a seed iso. This may be useful when
+using the `kernel_cmdline` behavior or when using an image that already
+has keys and/or passwords configured.
```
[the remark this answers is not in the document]
TheRealFalcon: I think we should only have separate code samples if the cloud provides additional API features on top of the core API. It doesn't here. I think it would be good to have an overarching example that can work on multiple clouds though.

pycloudlib/qemu/cloud.py
```diff
@@ -0,0 +1,634 @@
...
+            self.parent_dir / Path(image_id).stem
+        )
+        instance_dir.mkdir()
+        self._log.info(
+            "Using instance dir '%s' for new instance launched from '%s'",
+            instance_dir,
+            image_id,
+        )
+
+        # We make a COW image from the base image so we can make destructive
```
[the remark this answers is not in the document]
TheRealFalcon: I actually meant COW here, but I'll change the wording to `copy-on-write`

docs/clouds/qemu.md
TheRealFalcon: Yes, thanks for mentioning these tools. I tried myself with qemu-nbd and guestmount, but ran into some brick walls.

pycloudlib/qemu/cloud.py
TheRealFalcon: Agreed. I had a hard time figuring out the best way to do this. There are tons of options we could support, and I didn't want to go about including all of them in the launch api. I figure more can be added or we can further encode the "instance_type" string as needed. I did include a `launch_args` parameter that just appends its arguments to the end of the qemu call, though I consider it kind of a hack.

pycloudlib/qemu/cloud.py
TheRealFalcon: Do you mean something more than [this](https://github.com/canonical/pycloudlib/pull/323/files#diff-9c031a07867513f1651aa8c522f0ca475eed8b5756bafa4ebaa92966811fdb57R528). I do realize that won't let you unwrite/rewrite certain parameters, but I'm not sure the best way to support that.

TheRealFalcon:
@blackboxsw , I'm not sure if I understand exactly what you're asking for wrt to `qemu_args` (see inline responses), but otherwise I believe I addressed all of your comments.

docs/clouds/qemu.md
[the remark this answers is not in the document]
blackboxsw: Yes separate PR and iteration on that should be fine and it'd be preparatory work anyway for potential support of other distros anyway so it could be folded into that type of new feature. In the desktop case, we can still curl/wget the image ourselves outside of pycloudlib and feed it into the test framework to exercise the instance lifecycle.

pycloudlib/qemu/cloud.py
```diff
@@ -0,0 +1,651 @@
...
+            f"unix:{socket_path},server,wait=off",
+            "-no-shutdown",
+        ]
+
+        if not no_seed_iso:
+            # By default, even if no user data is passed, we'll need to
+            # use NoCloud and create a seed iso, otherwise we have no
+            # way of accessing our instance
+            seed_path = self._create_seed_iso(
+                instance_dir=instance_dir,
```
blackboxsw, line 515: Do we want the opportunity to provide network_config param at launch and inject it into the seed ISO as well for NoCloud datasource? I can think of more complex cases where we may want to add mutiple nics and or setup bridges/bonds etc and validate this configuration within a qemu instance. We can see something like this historically done in certain SRU tests https://github.com/cloud-init/ubuntu-sru/blob/main/bugs/1682871/btest-launch.sh

TheRealFalcon: Yes, good idea

TheRealFalcon: Added in most recent push

pycloudlib/qemu/cloud.py
```diff
@@ -0,0 +1,661 @@
...
+            username=kwargs.get("username"),
+        )
+
+    def _create_seed_iso(
+        self,
+        instance_dir: Path,
+        user_data: Optional[str],
+        meta_data: Optional[str],
+        vendor_data: Optional[str] = None,
+        network_data: Optional[str] = None,
```
blackboxsw, line 265: Shouldn't we call this network_config because that is the name of the configuration that cloud-init is consuming?

PR #324, opened 2023-10-20
Make id a required part of BaseInstance

TheRealFalcon:
BaseCloud has a required `get_instance()` method that can return an
instance given an instance id, but BaseInstance had no consistent API
to retrieve such an instance id. This adds it everywhere.


aciba90 (approved):
LGTM, thanks!

PR #325, opened 2023-10-23
feat: Allow for disk size in ec2 instances

TheRealFalcon:
Our EC2 code allows adding and removing volumes, but not setting the default volume size. For a t3.micro, it is only 8G. This is causing some cloud-init tests to fail.

This commit allows for specifying default disk size, and sets the default to 15G.

## Context
See failures here: https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-mantic-ec2/89/#showFailuresLink

Also [pricing](https://aws.amazon.com/ebs/pricing/) for storage is cheap enough that I don't think we need to be concerned about increased costs.


aciba90 (approved):
LGTM, thanks!

PR #327, opened 2023-10-25
fix(qemu): Fix no QMP socket found

TheRealFalcon:
During QEMU cleanup, the directory containing the instance image and QMP socket are cleaned up. If a new instance is immediately launched using the same paths, the QMP socket file fails to appear.

This commit updates the QEMU cloud to always use a new directory name for the next instance, rather creating one with the same old path.


aciba90 (approved):
LGTM!

PR #328, opened 2023-11-27
releases: add noble to azure dailys and UBUNTU_RELEASE_VERSION_MAP

blackboxsw:


TheRealFalcon (approved):
LGTM, but there's a linting error to fix first.

For future reference (or discussion) I would only increment the patch number if it's a bug fix. Even though this is a very small improvement, I would consider it a change that should bump the minor number instead.

blackboxsw:
There was an ordering issue w/ globbing files in the qemu unittests. Different versions of python order those files in differently. I added an explicit sorted to the list generated to ensure we always get the first lexicographic matching file in an image directory.

PR #332, opened 2024-01-04
feat(oracle): add kwargs to utility functions

mjd-x:


VERSION
```diff
@@ -1 +1 @@
-1!5.12.1
+1!5.13.1
```
aciba90, line 1: I think we should reset the patch version to 0.

```suggestion
1!5.13.0
```

aciba90 (changes requested):
Apart from the version number, LGTM, thanks!

pycloudlib/oci/utils.py
```diff
@@ -13,7 +15,9 @@
 log = logging.getLogger(__name__)
 
 
-def wait_till_ready(func, current_data, desired_state, sleep_seconds=1000):
+def wait_till_ready(
+    func, current_data, desired_state, sleep_seconds=1000, **kwargs
```
TheRealFalcon, line 19: I think this could lead to some somewhat weird call semantics. For example, with a call like
```python
wait_till_ready(
            func=self.compute_client.get_instance,
            current_data=self.instance_data,
            desired_state="RUNNING",
            some_other_kwarg="mykwarg",
        )
```

it's not really obvious that `some_other_kwarg` is being used as an argument to the `func`. What if the function definition was instead something like
```python
def wait_till_ready(
    func, current_data, desired_state, sleep_seconds=1000, func_kwargs: Dict
)
```

The call to the passed function then looks almost exactly the same:
```python
current_data = func(current_data.id, **func_kwargs).data
```

but the callsite is a little more clear:
```python
wait_till_ready(
            func=self.compute_client.get_instance,
            current_data=self.instance_data,
            desired_state="RUNNING",
            func_kwargs={"some_other_kwarg": "mykwarg"},
        )
```

Thoughts?

TheRealFalcon (commented):
Overall LGTM, but I do have one inline comment/question.

mjd-x:
Changed the function arguments as suggested, and also added the necessary changes to the methods that use this function (and propagated the `**kwargs` definition to the base class and every child class to keep the signatures consistent)

TheRealFalcon (approved):
LGTM!

Issue #308, opened 2023-05-31
Azure: network/non-compute overrides


jrtknauer:
# Summary

The `pycloudlib.azure` interfaces and implementation set many defaults for deployment, but with no means of customizing deployments outside of of the `vm_params` kwarg for the `launch()` method (and this only applies to the compute deployment). Issue #298 already identifies a VM deployment configuration which should be explicitly enumerated into the `launch()` interface; this issue applies to the configuration of any and all deployment resources which users currently have no access to, specifically:
- NIC
- Virtual Network
- IP Address
- Network Security Group
- The underlying resource group, which is implicitly created upon client instantiation

`pycloudlib` should set sane defaults - but offer a certain degree of configurability such that users are not locked into hard-coded defaults with no means of circumventing short of forking the library.

# Motivation

In addition to the common-sense option for users to have the means of overriding defaults, there are real Azure-specific constraints which cannot easily be overcome within the current deployment implementation:

1. **Resource Quotas** - As an example, `pycloudlib` is configured to deploy "Standard" Azure IP Addresses for VM deployments, where "Standard" is actually the premium IP Address resource, and thus has a lower default quota compared to the "Basic" IP Address. Increasing quotas in Azure when dealing with larger deployments can be difficult (not guaranteed to be approved, especially with premium resources), so it is only logical that `pycloudlib` does not lock users into resources which are gated by the cloud provider.
2. **Resource Efficiency** - in a similar vein to the first point, the default deployment scheme can actually be quite wasteful (full set-up and tear-down of all requisite resources for a given deployment cycle). While I do not suggest we re-implement Terraform/ARM Templates etc. for highly customized VM deployments, offering a means similar to `vm_params` for default overrides will enable users to customize deployments to their liking while still utilizing `pycloudlib`'s framework.

PR #334, opened 2024-01-25
feat(azure): user can provide overrides for various azure non-compute options. (SC-1697)

CalvoM:
We providing a way for users to override the non-compute options.
Fixes: #308


pycloudlib/azure/util.py
```diff
@@ -14,6 +15,19 @@
 )
 
 
+class AzureParams(NamedTuple):
+    """Azure Parameters Class."""
+
+    name: str
+    parameters: Optional[Dict[str, Any]]
```
jrtknauer, line 22: Add a comment explaining the purpose of this class and its attributes - specifically how it relates to the azure sdk.

pycloudlib/azure/cloud.py
```diff
@@ -699,9 +723,13 @@ def launch(
         inbound_ports=None,
         username: Optional[str] = None,
         resource_group: Optional[util.AzureParams] = None,
-        network_security_group: Optional[util.AzureCreateParams] = None,
-        virtual_network: Optional[util.AzureCreateParams] = None,
-        subnet: Optional[util.AzureCreateParams] = None,
+        network_security_group_: Optional[util.AzureCreateParams] = None,
```
jrtknauer, line 726: Is there a style rationale for the trailing underscore?

jrtknauer (changes requested):
In addition to the inline reviews, the commit message for 19e7ff2361bdc009222b4eff7fe9558ba38d6875 does not align with the changes being made.

pycloudlib/azure/cloud.py
CalvoM: I ran into an issue with mypy, just added the _ to differentiate it with other variables in the function. It was just a hack around the issue.

jrtknauer: This is a fairly pedantic issue, but mangling variable names to make linters happy is sidestepping the issue. It would be more appropriate to rename the argument names to more closely align to what they actually are (not to embed the type of the resource within the name - but to have a more accurate definition). The arguments of `network_security_group_` and `virtual_network_` do not represent those SDK objects, but rather the configuration for them.

I would recommend renaming to either `nsg_params`/`vnet_params`/`subnet_params`. Likewise for the `ip_addresses` and `network_interface_clients` - these are all arguments which encapsulate configuration for the respective resource type - not the resource itself.

pycloudlib/azure/cloud.py
```diff
@@ -633,6 +722,14 @@ def launch(
         name=None,
         inbound_ports=None,
         username: Optional[str] = None,
+        resource_group: Optional[util.AzureParams] = None,
+        network_security_group_: Optional[util.AzureCreateParams] = None,
+        virtual_network_: Optional[util.AzureCreateParams] = None,
+        subnet_: Optional[util.AzureCreateParams] = None,
+        ip_addresses: Optional[List[Optional[util.AzureCreateParams]]] = None,
+        network_interface_clients: Optional[
```
aciba90, line 730: I think it would be clearer to name this parameter as this, as it contains parameters to create a network interfaces, not clients.

```suggestion
        network_interfaces: Optional[
```

pycloudlib/azure/cloud.py
```diff
@@ -665,15 +766,21 @@ def launch(
                 f"{self._type} launch requires image_id param."
                 f" Found: {image_id}"
             )
+        if not ip_addresses:
+            ip_addresses = list([None])
```
aciba90, line 770: Why not just?

```suggestion
            ip_addresses = [None]
```

pycloudlib/azure/cloud.py
```diff
@@ -665,15 +766,21 @@ def launch(
...
+        if not network_interface_clients:
+            network_interface_clients = list([None])
```
aciba90, line 772: Same comment.

pycloudlib/azure/cloud.py
```diff
@@ -683,46 +790,67 @@ def launch(
...
             # use it
-            nic = self._check_for_network_interfaces()
+            if not network_interface_clients:
+                nic = self._check_for_network_interfaces()
+
+        created_nics = [nic]
+        created_ip_addresses = []
 
         if nic is None:
+            created_nics.clear()
```
aciba90, line 800: Nit: Why not:

```diff
diff --git a/pycloudlib/azure/cloud.py b/pycloudlib/azure/cloud.py
index 889465b..41be201 100644
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -786,18 +786,18 @@ class Azure(BaseCloud):
         # it to enable more ports. The rationale for is that we want to reuse
         # those resources only if they are generic enough
         nic = None
+        created_nics = []
         if not inbound_ports:
             # Check if we already have an existing network interface that is
             # not attached to a virtual machine. If we have, we will just
             # use it
             if not network_interface_clients:
                 nic = self._check_for_network_interfaces()
+                created_nics.append(nic)
 
-        created_nics = [nic]
         created_ip_addresses = []
 
         if nic is None:
-            created_nics.clear()
             self._log.debug(
                 "Could not find a network interface. Creating one now"
             )
```

It feels more natural to not have invalid values on that list, so that the code is more easy to read / extend.

pycloudlib/azure/cloud.py
```diff
@@ -665,15 +766,21 @@ def launch(
                 f"{self._type} launch requires image_id param."
                 f" Found: {image_id}"
             )
+        if not ip_addresses:
+            ip_addresses = list([None])
+        if not network_interface_clients:
+            network_interface_clients = list([None])
+
+        assert len(network_interface_clients) >= len(ip_addresses)
```
aciba90, line 774: Users can provide less IP addresses than network interfaces to be used, so that pycloudlib will create the missing ones. I think:

1. We should probably document this fact in the function.
2. Raise a proper exception, allowing users to catch it.

aciba90 (changes requested):
Thanks, @CalvoM!

Overall look good, I left some inline comments.

pycloudlib/azure/cloud.py
```diff
@@ -328,27 +395,42 @@ def _create_ip_address(self):
...
         return ip_poller.result()
 
     def _create_network_interface_client(
-        self, ip_address_id, subnet_id, nsg_id, nic_params=None
+        self,
+        ip_address_id,
+        subnet_id,
+        nsg_id,
+        nic_params=None,
+        nic: Optional[util.AzureCreateParams] = None,
```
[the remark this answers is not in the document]
CalvoM: Oh yeah, I was not yet decided how to do that. I thought I should not break the api by removing `nic_params`. But I guess removing it earlier is better than later.

jrtknauer:
The commit message for d51b99f58593016dcd286dd0f46692a14107b471 contains an extraneous '!'.

VERSION
```diff
@@ -1 +1 @@
-1!5.14.0
+1!5.15.0
```
aciba90, line 1: As we are breaking the API, we should bump the major version.

```suggestion
1!6.0.0
```

tests/unit_tests/azure/test_cloud.py
```diff
@@ -119,3 +130,366 @@ def test_create_network_interface_with_inbound_ports(
...
+        )
+        type(resource_client_mock).resource_groups = mock.PropertyMock(
+            return_value=resource_group_mock
+        )
+        resource_group_mock.create_or_update.return_value = resource_mock
+        resource_group_mock.get.side_effect = ResourceNotFoundError()
+
+        new_params = {"location": "new_location", "tags": {"name": "new-tag"}}
+        new_rg = AzureParams("default-pyc-rg", new_params)
+        instance = Azure(
```
aciba90, line 156: Being Azure as subclass of BaseCloud, this is typically named as cloud because it represents that concept. We also have BaseInstance, which represent instances within a particular cloud. I think it makes sense to be consistent and name this as cloud in this and the other tests.

```suggestion
        cloud = Azure(
```

pycloudlib/azure/cloud.py
```diff
@@ -665,64 +798,96 @@ def launch(
...
                 f" Found: {image_id}"
             )
+        if not ip_addresses_params:
+            ip_addresses_params = [None]
+        if not network_interfaces:
+            network_interfaces = [None]
+
+        if len(ip_addresses_params) > len(network_interfaces):
+            raise PycloudlibError(
+                "IP addresses should not be more than the NICs provided."
```
aciba90, line 808: I think with this message, we directly refer to the input parameters and could be simpler to understand for users.

```suggestion
                "The number of `ip_addresses_params` cannot be bigger than the number of `network_interfaces`."
```

aciba90 (changes requested):
Thanks, @calvin, for addressing the comments.

Nice to see the consistent suffixes in the parameter names with types [AzureParams, AzureCreateParams].

Apart from the inline comments, PRs are always squashed in this project, so unless there is a good reason to change this, could you please summarize a commit message including all the changes in the main PR comment?

aciba90:
> The commit message for [d51b99f](https://github.com/canonical/pycloudlib/commit/d51b99f58593016dcd286dd0f46692a14107b471) contains an extraneous '!'.

The `!` represent a breaking change in [conventionalcommits](https://www.conventionalcommits.org/en/v1.0.0/).

jrtknauer:
> > The commit message for [d51b99f](https://github.com/canonical/pycloudlib/commit/d51b99f58593016dcd286dd0f46692a14107b471) contains an extraneous '!'.
> 
> The `!` represent a breaking change in [conventionalcommits](https://www.conventionalcommits.org/en/v1.0.0/).

My mistake - was not familiar with the annotation.

aciba90 (approved):
LGTM, thanks!

jrtknauer (approved):

PR #338, opened 2024-02-02
feat(azure): get boot diagnostics of instances

CalvoM:


CalvoM:
@aciba90 @jrtknauer. Could you please review this?

pycloudlib/azure/cloud.py
```diff
@@ -161,6 +170,26 @@ def __init__(
...
+        """Get azure diagnostics and then cleanup."""
+        if exc_type or self._enable_boot_diagnostics:
+            for instance in self.created_instances:
+                boot_diagnostics = self.get_boot_diagnostics(instance)
+                # min. delay for the url to be active.
+                time.sleep(60)
+                if boot_diagnostics:
+                    response = requests.get(boot_diagnostics.console_log_url)
+                    filename = f"{instance.name}.boot_diagnostics"
+                    with open(filename, "w", encoding="utf-8") as f:
```
jrtknauer, line 185: As discussed in stand up, I do not agree with writing the boot diagnostics to file by default. It diverges from how the library has been handling logs and runtime information (the only other use of `open` is within `_download_file` and for handling ssh keys).

I'm in the camp of libraries should not directly write logging artifacts. Simply download the contents of the console URIs with one of two options:

1. Dump directly to logging and defer to the application logging handlers to access the logs.
2. Store the logs as part of the `Azure` class, and implement the appropriate exception handling for `_create_virtual_machine` to store the logs as part of the Azure cloud and allow the application to use the logs as it sees fit.

Ideally you would do both. Applications can either be lazy and use a handler to access the logs, or have full introspection into the logs at runtime and decide what to do with them then (with the benefit of not having a library arbitrarily writing files).

pycloudlib/azure/cloud.py
```diff
@@ -161,6 +170,26 @@ def __init__(
 
         self.resource_group = self._create_resource_group()
         self.base_tag = tag
+        self._enable_boot_diagnostics = enable_boot_diagnostics
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        """Get azure diagnostics and then cleanup."""
+        if exc_type or self._enable_boot_diagnostics:
```
jrtknauer, line 177: `get_boot_diagnostics` already checks if boot diagnostics has been enabled. See those review comments as well.

pycloudlib/azure/cloud.py
```diff
@@ -1009,6 +1041,31 @@ def snapshot(
 
         return image_id
 
+    def get_boot_diagnostics(
```
jrtknauer, line 1044: There's a lot going on here. You've exposed boot diagnostic retrieval as a public interface, but realistically the responsibilities should be split:

1. A private method which automatically tries to get the boot diagnostics for an instance as part of creation, saving the boot diagnostics in memory
2. A public interface for users to retrieve the diagnostics ad hoc

With the current implementation the class will constantly need to re-request the diagnostics along with all of the state management which comes with it. Additionally, this implementation requires the user to specify the instance they want to have boot diagnostics for, and arbitrarily returns the boot diagnostics for the first VM in the VM list if none is specified.

We care about boot diagnostics mainly for when things go wrong, which will likely happen when calling `launch()`. Ideally this method would be bound to a cloud's instance, since it is for a given instance we are retrieving the boot diagnostics for. Unfortunately the Azure SDK requires both the VM name and resource group to resolve the diagnostics, so you've correctly injected the instance as part of the method signature. However the user will only have access to the instance if `launch()` succeeds, hence your default for grabbing the first instance if none is specified. However, if users want to retrieve the logs for failed deployments after the first created instance they're out of luck, so the interface is now inconsistent.

A bit overly detailed but I wanted to be explicit in how the current interface is hiding too many details from the user where both the usage and results are cumbersome. Revisit this interface and work backwards from what we you want to expose to the user (return the logs) and determine a more appropriate means of mapping boot diagnostics to created instances - where the set of possibilities include:

- No running instances
- Boot diagnostics is disabled
- One running instance
- Many running instances
- Requesting for an instance that has failed to launch

jrtknauer (changes requested):

pycloudlib/azure/cloud.py
```diff
@@ -615,15 +632,24 @@ def _create_virtual_machine(
...
-                params,
+        try:
+            vm_poller = (
+                self.compute_client.virtual_machines.begin_create_or_update(
+                    self.resource_group.name,
+                    name,
+                    params,
+                )
+            )
+            vm_poller.wait()
```
CalvoM, line 643: @jrtknauer Do you think we could have a timeout here, considering that it seems Azure's timeout is like 20 mins. I found out this default timeout from testing failed images.

pycloudlib/azure/cloud.py
```diff
@@ -615,15 +632,24 @@ def _create_virtual_machine(
...
-                params,
+        try:
+            vm_poller = (
+                self.compute_client.virtual_machines.begin_create_or_update(
+                    self.resource_group.name,
+                    name,
+                    params,
+                )
+            )
+            vm_poller.wait(300)
```
CalvoM, line 643: @jrtknauer Do you think we could have a timeout here, considering that it seems Azure's timeout is like 20 mins. I found out this default timeout from testing failed images. I am defaulting to 5 mins since I notice most of successful provisioning have taken like 3 mins.

jrtknauer: I would default to the Azure timeout and ideally provide an option for a user to override. Burying a timeout - especially when its explicitly truncating the Azure default could end up backfiring on users if there's an unexpectedly slow deployment.

pycloudlib/azure/instance.py
```diff
@@ -2,19 +2,42 @@
...
+BootDiagnostics = namedtuple("BootDiagnostics", ["console_log_url", "logs"])
+
+
+class VMInstanceStatus(Enum):
+    """Represents Azure VM state."""
+
+    UNPROVISIONED = 0
+    FAILED_PROVISION = 1
+    ACTIVE = 2
+    DELETED = 3
```
jrtknauer, line 24: Can be simplified with `auto()` since the underlying value of the enumeration in irrelevant.

pycloudlib/azure/instance.py
```diff
@@ -101,6 +127,17 @@ def offer(self):
         image_profile = self._instance["vm"].storage_profile.image_reference
         return getattr(image_profile, "offer", "")
 
+    def console_log(self):
```
jrtknauer, line 130: Missing type hint on the return - `Optional[str]`

jrtknauer (changes requested):
Core logic seems fine - just a few changes and if you could outline the manual testing you performed in addition to providing an example console lop output from `console_log()` for both a successful and failed (even if artificially induced) deployment.

pycloudlib/azure/cloud.py
```diff
@@ -615,15 +641,26 @@ def _create_virtual_machine(
...
+            vm_poller.wait(provisioning_timeout)
+            if not vm_poller.done():
+                raise PycloudlibTimeoutError(
+                    "Virtual machine creation timed out."
+                )
+            return vm_poller.result()
+        except HttpResponseError as e:
+            err_code = e.error.code
+            err_msg = e.error.message
+            raise PycloudlibTimeoutError(
```
aciba90, line 661: I think we should use another exception here, as we do not for sure that's the case.

aciba90 (changes requested):
@CalvoM and @jrtknauer, what do you think about the following:

> https://github.com/canonical/pycloudlib/pull/338#discussion_r1490369674

1. `Azure.launch` raise an exception which carries the boot diagnosis info if instance provisioning timeouts and requested.
2. `AzureInstance` offer a `get_diagnostics` methods which pulls on demand the info.

If we go with this, we could probably remove the instance status dance regarding `VMInstanceStatus`.

> Dump directly to logging and defer to the application logging handlers to access the logs.

I think we should probably gate this logging under an extra `log_boot_diagnosis` parameter, as those log entries are going to be very busy and not every client is going to be interested in them, probably.

CalvoM:
> @CalvoM and @jrtknauer, what do you think about the following:
> 
> > [#338 (comment)](https://github.com/canonical/pycloudlib/pull/338#discussion_r1490369674)
> 
> 1. `Azure.launch` raise an exception which carries the boot diagnosis info if instance provisioning timeouts and requested.

This is a good suggestion since it will help with providing the user with all the details concerning the failure on the spot.
> 2. `AzureInstance` offer a `get_diagnostics` methods which pulls on demand the info.
> 

We opted to store the diagnostics info in memory because the diagnostics uris have a timeout ranging from 2 hrs (default) to 24hrs(max). Storing it in memory seems like a better and safer option. However, I do see an argument for this since the vm provision fails in about 21 mins. Secondly, I am thinking that it is better to add 1 minute, minimum time for the uris to be active, during provisioning rather than when the user requests for the log diagnostics. What do you think?

pycloudlib/azure/cloud.py
CalvoM: Agreed. Thanks for the suggestion.

pycloudlib/azure/instance.py
```diff
@@ -2,19 +2,42 @@
...
+from azure.core.exceptions import ResourceExistsError
+
 from pycloudlib.errors import PycloudlibTimeoutError
 from pycloudlib.instance import BaseInstance
 
+BootDiagnostics = namedtuple("BootDiagnostics", ["console_log_url", "logs"])
+
+
+class VMInstanceStatus(Enum):
+    """Represents Azure VM state."""
```
aciba90, line 19: As I see it, this enum does represent pycloudlib's vision of the instance, not the general state/status of it. I wonder if we can improve the documentation for future us.

aciba90 (changes requested):
Thanks, @CalvoM, for this. General workflow looks good to me!

I have been able to test that `enable_boot_diagnostics` in the `Azure` cloud behaves correctly with instances that do not fail to be provisioned. I was not able to reproduce the provisioning error because I hadn't access to `/subscriptions/1b7499f2-637c-46b5-b448-c97e9267e522/resourceGroups/main_sa_north/providers/Microsoft.Compute/galleries/ubuntu/images/error-jammy-2/versions/2.0.0`.

I left inline comments / requests.

In addition to that, right now `Azure.get_instance` will always set `status` to `ACTIVE`. I wonder if there is a way to check that an instance failed to be provisioned and set it appropriately.

pycloudlib/azure/cloud.py
```diff
@@ -904,14 +945,23 @@ def launch(
...
 
-        vm = self._create_virtual_machine(
-            image_id=image_id,
-            instance_type=instance_type,
-            nic_ids=nic_ids,
-            user_data=user_data,
-            name=name,
-            vm_params=vm_params,
-        )
+        vm_state = VMInstanceStatus.UNPROVISIONED
```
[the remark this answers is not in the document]
CalvoM: I had just created that `UNPROVISIONED` state as an initializing state.

CalvoM:
@aciba90 I have updated the PR description to outline steps to reproduce the failed vms.

pycloudlib/azure/cloud.py
```diff
@@ -904,14 +945,23 @@ def launch(
...
+                instance_type=instance_type,
+                nic_ids=nic_ids,
+                user_data=user_data,
+                name=name,
+                vm_params=vm_params,
+                provisioning_timeout=provisioning_timeout,
+            )
+            vm_state = VMInstanceStatus.ACTIVE
+        except PycloudlibTimeoutError:
+            self._log.info("Getting details of VM that failed to provision.")
```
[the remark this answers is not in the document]
CalvoM: Instead of moving it into the `__init__` we can move to the `_get_boot_diagnostics` method so that it is logged when we obtain the boot diagnostics.

pycloudlib/azure/instance.py
```diff
@@ -120,6 +157,23 @@ def generalize(self):
...
 
+    def _get_boot_diagnostics(self) -> Optional[str]:
+        """Get VM boot diagnostics logs."""
+        response = None
+        try:
+            virtual_machines = self._client.virtual_machines
+            diagnostics = virtual_machines.retrieve_boot_diagnostics_data(
+                self._instance["rg_name"], self.name
+            )
+            time.sleep(60)
```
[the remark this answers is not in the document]
aciba90: Could we leave a comment explaining that, please?

pycloudlib/azure/cloud.py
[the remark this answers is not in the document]
aciba90: But can an instance end up with that state right now?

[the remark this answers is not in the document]
CalvoM: No it cannot. I have removed it.

VERSION
```diff
@@ -1 +1 @@
-1!6.0.3
+1!6.1.0
```
aciba90, line 1: Could you please rebase to fix conflicts and bump the version again?
```suggestion
1!6.2.0
```

jrtknauer (approved):

PR #342, opened 2024-02-19
ci: point to the gce creds via the GCE_CREDENTIALS_PATH env var

paride:
This allows to use a more generic pycloudlib.toml conf.


paride:
If CI passes I'll bump VERSION.

TheRealFalcon (approved):
Approved assuming version bump

paride:
@TheRealFalcon thanks, VERSION is bumped, CI is green, merging this.

PR #343, opened 2024-02-19
WIP: ci: fix credentials

paride:
Temporary hack to debug CI.


paride:
OK, I'm reconnecting the dots. This can't be tested this way because of how GitHub handles secrets.

PR #347, opened 2024-02-20
No-change VERSION bump to 6.0.3

paride:
I wasted version 1!6.0.2 while testing one previous PR, which required running jobs that normally only run when new commits land to the main branch. We can't just delete the tag and delete the release from pypi because pypi doesn't allow version reuse. Let's just bump VERSION yet another time.


paride:
I think this happened because I pushed the previous PR to the `main` branch of my fork. My fork does not have the pypi creds, but when I opened the PR I allowed creds sharing with the fork, which were needed to actually test my changes to the main-check job. However this also created the tag and pushed to pypi, pre-merge.

Once the merge happened, the real main-check jobs tried to do the same. I tried to clear the release, but that is not enough, as explained in the commit message. So here we are.

TheRealFalcon (approved):

PR #349, opened 2024-02-21
Improve oracle example script's ease of use and rename it so that it can actually be run

a-dubs:
Previously, if you try to run the oracle example script, it would throw an error because the file is named `oci.py` and this conflicts with the naming of the `oci` module such that pycloudlib would fail to import the `oci` module. Thus, it has been renamed to `oracle.py`.

Also, the flexibility and ease of use of the script has been improved so that the `compartment_id` and `availability_domain` values can be set either in variables in the script, or passed as cli args. And if neither of these two things are done, then the script will simply prompt the user for the two values.


examples/oracle.py
```diff
@@ -0,0 +1,83 @@
...
+
+        with client.launch(image_id=snapshotted_image_id) as new_instance:
+            new_instance.wait()
+            new_instance.execute("whoami")
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.DEBUG)
+    
+    # fill in the  variables here if you'd rather not
```
TheRealFalcon, line 48: Hardcoded variables in scripts that a user is expected to modify like this are generally frowned upon, especially if the file lives in version control. If you edit the file, now you can't make changes anywhere else without this being included in your source change.

Configuration is generally accomplished by:
1. CLI args
2. Environment variables
3. In some cases, a separate file that is `.gitignore`ed.

examples/oracle.py
```diff
@@ -0,0 +1,83 @@
...
+    # pass availability domain and compartment id as arguments via cli
+    AVAILABILITY_DOMAIN = ""
+    COMPARTMENT_ID = ""
+
+    # if no arguments are passed and the variables are not set
```
TheRealFalcon, line 53: I don't see the purpose of interactive input when a user can simply pass CLI args. It might be more hand-holdy the first time somebody runs a script, but beyond that there are a few downsides.

It's not repeatable. With a CLI args, I can paste the values into the CLI, then recall the same command from my history or throw it in a script. If it's interactive, I need to re-copy and re-paste the same values every single time I run it, and running it within a script becomes much harder.

It can quickly become unmaintainable. Need another arg? What about flags that only apply to specific args? 

What's the harm in seeing a usage line that tells you the correct input?

examples/oracle.py
```diff
@@ -0,0 +1,83 @@
...
+
+import pycloudlib
+
+cloud_config = """#cloud-config
+runcmd:
+  - echo 'hello' > /home/ubuntu/example.txt
+"""
+
+
+def demo(availability_domain: str, compartment_id: str):
```
TheRealFalcon, line 17: The default way of getting these values is from the `~/.config/pycloudlib.toml` values. If we default these to `None` and no arguments are passed to this script, it should "just work" if the toml values were set correctly.

TheRealFalcon (changes requested):
Thanks for the rename!

I'm actually -1 on the rest of the changes here though. I left some comments inline explaining why. Additionally, while they'd be overkill for such a simple script, Canonical does have some [UI guidelines](https://discourse.ubuntu.com/c/design/cli-guidelines/62). In particular, [interactive prompts](https://discourse.ubuntu.com/t/interactive-prompts/18881) and [incomplete commands](https://discourse.ubuntu.com/t/incomplete-commands/18882) may be useful.

examples/oracle.py
a-dubs: Agreed. I'll nuke the changes i made to the script. thanks for presenting this feedback and thoroughly explaining why. I appreciate it.

examples/oracle.py
a-dubs: Great point! I'll modify the script to do so. I thank _that_ truly makes the script more streamlined compared to the other bloated changes i added. thanks!

TheRealFalcon (commented):
LGTM, but there's also some CI failures to address and version bump.

a-dubs:
@TheRealFalcon will this be squashed? if not lmk and ill clean up my commits lmfao

TheRealFalcon:
Yeah, we can squash, but mypy and pylint are still complaining

TheRealFalcon (approved):

PR #350, opened 2024-03-14
fix(azure): add_network_interface int tests pass

CalvoM:
Azure has to deallocate the VM and resources before attaching the new network interface.


aciba90 (changes requested):
Thanks for the fix, @CalvoM.

I have a couple of concerns.

1. I think we should set the status back in `instance.start` to `ACTIVE`.
2. The flow `deallocate` -> `add_network_interface` -> `start` is specific to azure, and call sites of this method would have to know over which cloud they are acting upon, diminishing the value of this interface. I think it, unless there is some reason against it, that `azure_instance.add_network_interface` should handle the change of status internally.

Unified proposed diff, if we agree on it:

```diff
diff --git a/pycloudlib/azure/instance.py b/pycloudlib/azure/instance.py
index 63a44bb..dfd1f9a 100644
--- a/pycloudlib/azure/instance.py
+++ b/pycloudlib/azure/instance.py
@@ -206,6 +206,9 @@ class AzureInstance(BaseInstance):
         if wait:
             start.wait()
             self.wait()
+            self._status = VMInstanceStatus.ACTIVE
+        else:
+            self._status = VMInstanceStatus.ACTIVE
 
     def _do_restart(self, **kwargs):
         """Restart the instance."""
@@ -288,10 +291,11 @@ class AzureInstance(BaseInstance):
 
     def _attach_nic_to_vm(self, nics: List[Dict[str, Any]]):
         """Attach nics to instance."""
+        do_start = False
         if self._status != VMInstanceStatus.STOPPED:
-            raise PycloudlibError(
-                "VM must be deallocated before attaching NIC"
-            )
+            self._log.debug("Deallocating instance to attach nics")
+            self.deallocate()
+            do_start = True
         params = self._instance["vm"].as_dict()
         vm_attached_nics = params["network_profile"]["network_interfaces"]
         vm_attached_nics.extend(nics)
@@ -303,6 +307,8 @@ class AzureInstance(BaseInstance):
             self._instance["rg_name"], self.name, params
         )
         poll.result()
+        if do_start:
+            self.start()
 
     def _create_ip_address(self):
         us = datetime.datetime.now().strftime("%f")
diff --git a/tests/integration_tests/test_public_api.py b/tests/integration_tests/test_public_api.py
index 5d980b9..e04db28 100644
--- a/tests/integration_tests/test_public_api.py
+++ b/tests/integration_tests/test_public_api.py
@@ -58,11 +58,7 @@ def exercise_instance(instance: BaseInstance):
     assert ipaddress.ip_address(instance.ip)
 
     with suppress(NotImplementedError):
-        if isinstance(instance, pycloudlib.azure.instance.AzureInstance):
-            instance.deallocate()
         ip_address = instance.add_network_interface()
-        if isinstance(instance, pycloudlib.azure.instance.AzureInstance):
-            instance.start()
         try:
             assert ipaddress.ip_address(ip_address)
         finally:
```

Thoughts?

CalvoM:
@aciba90 As I was working on this PR I had the same idea but I was hesistant since it will mean the command will do more than just add/attach the NIC i.e. it will deallocate and start the vm.

pycloudlib/azure/instance.py
```diff
@@ -241,8 +244,14 @@ def add_network_interface(self) -> str:
...
 
         Creates NIC and adds to the VM instance.
+
+        NOTE: It will deallocate the virtual machine, add the NIC,
+        then start the virtual machine.
+
         Returns the ip address of the new NIC.
         """
+        # NOTE: Azure needs the VM to be deallocated to add NICs
+        self.deallocate()
```
aciba90, line 254: I think deallocate / start is not needed in this function, as it is already performed in _attach_nic_to_vm. Or am I missing something?

CalvoM: Oh yeah, I had not picked up on it.

aciba90 (approved):
LGTM, thanks!

PR #351, opened 2024-03-20
feat(azure): removing nics from NICs (CPC-4043)

CalvoM:
Testing script
```python
    from pycloudlib import Azure, azure

    rg = azure.util.AzureParams("main_sa_north", None)
    cloud = Azure(
        tag="calvinmwadime", region="southcentralus", resource_group_params=rg
    )
    inst = cloud.launch(cloud.daily_image("jammy"), name="JammyMultiDos2")
    print(f"Launched {inst.name} {inst.status}")
    print(f"IP: {inst.ip}")
    inst.deallocate()
    new_nic_ip = inst.add_network_interface()
    print(f"Added NIC of ip address: {new_nic_ip}")
    inst.start()
    inst.remove_network_interface(new_nic_ip)
    print(f"Launched {inst.name} {inst.status}")
    print(f"IP: {inst.ip}")

```


pycloudlib/azure/instance.py
```diff
@@ -290,7 +304,82 @@ def add_network_interface(self) -> str:
 
     def remove_network_interface(self, ip_address: str):
         """Remove nic from running instance."""
-        raise NotImplementedError
+        # get ip address id
+        all_ips = list(self._network_client.public_ip_addresses.list_all())
```
aciba90, line 308: In all other implementations of this method, `ip_address` is a private IP, this is because a NIC does not necessarily have to have a public IP.

Apart from that, to be consistent with the interface, could we implement it in that way?

pycloudlib/azure/instance.py
```diff
@@ -290,7 +304,82 @@ def add_network_interface(self) -> str:
...
+            else:
+                nic_params.append({"id": vm_nic.id, "primary": vm_nic.primary})
+        if not nic_to_remove:
+            raise PycloudlibError(
+                f"Did not find NIC with ip address: {ip_address}"
+            )
+        # if primary nic is removed, then make the next NIC as primary
+        if nic_to_remove.primary:
+            nic_params[0]["primary"] = True
+            primary_nic = [
```
aciba90, line 345: I get a `IndexError: list index out of range` when removing the last NIC, should that be handled?

This pattern, [list comprehension filtering][0], might fail in other instances added by this PR.

pycloudlib/azure/instance.py
```diff
@@ -290,7 +304,82 @@ def add_network_interface(self) -> str:
...
+                nic for nic in all_nics if nic.id == nic_params[0]["id"]
+            ][0]
+        self._remove_nic_from_vm(nic_params, primary_nic, all_ips)
```
aciba90, line 348: In all other implementations of this method, the NIC is also deleted, as it is right now, we do only effectively detach the interface and leak it.
Could we also implement a `self._remove_nic` or something in those lines?

pycloudlib/azure/instance.py
```diff
@@ -208,6 +211,17 @@ def start(self, wait=True):
             self.wait()
         self._status = VMInstanceStatus.ACTIVE
 
+    def _wait_for_instance_start(self, **kwargs):
```
aciba90, line 214: Why is this method needed now?

aciba90 (changes requested):
Thanks, @CalvoM. In general looks good. I left some inline concerns/questions.

pycloudlib/azure/instance.py
aciba90: Both private.

pycloudlib/azure/instance.py
CalvoM: 1. The default one in the abstract class does nothing.
2.  (more important) From my testing I noticed that sometimes during `start` we try accessing/using the vm before it is up. This is due to the fact that the azure `start` method is async i.e. [`begin_start` ](https://github.com/canonical/pycloudlib/blob/fec76cc7e52b4fd5397c8528727eade5ac159935/pycloudlib/azure/instance.py#L202) hence we do not confirm that it is up before running the various cloudinit commands. With this we can confirm that the vm is up and running.

pycloudlib/azure/instance.py
CalvoM: wrt deleting the nic, could we have a method for that, oorrr, we could pass an extra parameter to `remove_network_interface` to specify if we should delete it something like
```python
def remove_network_interface(self, ip_address, should_delete=False): ...
```
I am not sure about always deleting the NIC when detaching it from the VM.

pycloudlib/azure/instance.py
CalvoM: Ah, then I am putting up a PR to correct the implementation for the `add_network_interface` since I returned the public one.

aciba90: Mmm, I think we can fix it as part of this PR.

pycloudlib/azure/instance.py
aciba90: Gotcha, thanks for adding this!

pycloudlib/azure/instance.py
aciba90: > I am not sure about always deleting the NIC when detaching it from the VM.
Why not?

That might be a good compromise, would you like that default to be `False` in azure? If yes, other clouds default to `True`, and would be a bit strange to have different defaults per clouds, or maybe not. Thoughts?

pycloudlib/azure/instance.py
CalvoM: Yes, thought it would be a big change, but it is not. I am adding to this.

pycloudlib/azure/instance.py
CalvoM: Updated the logic here.

pycloudlib/azure/instance.py
CalvoM: Still thinking about this.

CalvoM: Had a meeting with @aciba90 and decided to delete in this method and then later create methods for detaching without deleting.

PR #353, opened 2024-03-22
feat(ibm): add ability to get image id from name

a-dubs:
This change makes it much easier to select an image to launch an instance from.


VERSION
```diff
@@ -1 +1 @@
-1!6.2.1
+1!6.2.3
```
uhryniuk, line 1: Should patch jump from x.x.1 -> x.x.3?

a-dubs: uhhhh nope. you saw nothing. this never happened.

uhryniuk (approved):
Looks good, sire

pycloudlib/ibm/cloud.py
```diff
@@ -403,6 +403,17 @@ def _get_or_create_key(self) -> str:
         self.created_keys.append(key_id)
         return key_id
 
+    def get_image_id_by_name(self, name: str) -> str:
```
holmanb, line 406: Nice! What sorts of images are you testing? Think that this is something we might want for other clouds? I can imagine this being useful, for example, to test cloud-init for non-ubuntu clouds.

Maybe we should put an abstract method in the base class as well so we can have a standardized interface for other clouds to match?

a-dubs: let me take a look at that! would definitely be something that would be both easy to add to all clouds and also just generally useful.

holmanb (approved):
LGMT!

Once the flake8 failures are fixed, I'll merge.

holmanb (approved):

PR #355, opened 2024-03-26
feat(ec2): ipv6 support for new nics

aciba90:
## do not squash

https://warthogs.atlassian.net/browse/SC-1754


pycloudlib/ec2/instance.py
```diff
@@ -81,9 +105,67 @@ def add_network_interface(self) -> str:
...
-        interface_id = self._create_network_interface()
+        interface_id = self._create_network_interface(ipv6_address_count)
+        if ipv4_public_ip_count is not None:
+            if (
+                ipv4_address_count is None
+                or ipv4_address_count < ipv4_public_ip_count
+            ):
+                raise PycloudlibError(
+                    f"Invalid combination of"
+                    f" `ipv4_address_count={ipv4_public_ip_count}` and"
```
CalvoM, line 126: I think you meant `ipv4_public_ip_count`.

pycloudlib/ec2/instance.py
```diff
@@ -129,7 +211,7 @@ def delete(self, wait=True) -> List[Exception]:
         """Delete instance."""
         exceptions = []
         # Even with DeleteOnTermination set True, nics can outlive instances
-        for ip in self.created_interfaces:
+        for ip in self.created_interfaces[:]:
```
CalvoM, line 214: Why have you opted for this syntax?

CalvoM (commented):
This is good. Could you please provide a sample script to test the functionality added?

pycloudlib/ec2/instance.py
aciba90: I am creating a copy of that list to iterate over because I want to remove the created interfaces from the list while being removed from this loop or manually.

See:

https://github.com/canonical/pycloudlib/pull/355/files#diff-0c9a2ff75eae3b564b5813cf4f5d5ea8cb15fbbf3b680b780fcb102503dc7e5cR550

pycloudlib/ec2/instance.py
aciba90: Good catch, thanks. Updated.

aciba90:
> This is good. Could you please provide a sample script to test the functionality added?

Thanks for the review. As for testing, one can execute the tests modified in https://github.com/canonical/cloud-init/pull/5102/. If not, to test this, one can:

1. Create an ec2 instance.
2. Add new interface with different values for the new parameters.
3. Remove the interface / instance.

pycloudlib/ec2/instance.py
CalvoM: Okay I had not seen that.

pycloudlib/ec2/instance.py
```diff
@@ -379,12 +466,61 @@ def _get_nic_matching_ip(self, ip_address):
...
+                    "Values": [private_ip],
+                },
+            ]
+        )
+        for association in self._find_nic_associations(nics_resp):
+            try:
+                self._client.disassociate_address(
+                    AssociationId=association["AssociationId"]
+                )
+                self._client.release_address(
```
blackboxsw, line 501: By placing these two operations in the same try block do we risk failing to attempt to release_address any time disassociate_address errors for each association? Is it worth separating these cases so we at least attempt both operations even if `disassociate_address` fails?

pycloudlib/ec2/cloud.py
```diff
@@ -360,12 +361,15 @@ def launch(
             vpc: optional vpc object to create instance under
             disk_size_gb: size of instance disk in GB
             username: username to use when connecting via SSH
+            enable_ipv6: indicates if ipv6 support must be added to the
```
blackboxsw, line 364:
```suggestion
            enable_ipv6: indicates if ipv6 IMDS support must be added to the
```

pycloudlib/ec2/instance.py
```diff
@@ -379,12 +466,61 @@ def _get_nic_matching_ip(self, ip_address):
...
+            Filters=[
+                {
+                    "Name": "private-ip-address",
+                    "Values": [private_ip],
+                },
+            ]
+        )
+        for association in self._find_nic_associations(nics_resp):
+            try:
+                self._client.disassociate_address(
```
blackboxsw, line 498: Do we even need to separately call disassociate_address? [Botocore seems to indicate that release_address will automatically disassociate](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/release_address.html#release-address)

blackboxsw (commented):
LGTM. Minor thought on potentially either:
- not running disassociate_address in the first place because release_address may encapsulate that operation
- or a suggestion to still attempt to run release_address even if disassociate_address fails because of the first point above

pycloudlib/ec2/instance.py
aciba90: That's only true in the default VPC:

> [Default VPC] Releasing an Elastic IP address automatically disassociates it from any instance that it’s associated with. To disassociate an Elastic IP address without releasing it, use DisassociateAddress.

> [Nondefault VPC] You must use DisassociateAddress to disassociate the Elastic IP address before you can release it. Otherwise, Amazon EC2 returns an error ( InvalidIPAddress.InUse).

pycloudlib/ec2/instance.py
aciba90: If `disassocite_address` fails, it is very likely that `release_address` is going to fail, as in non-default VPCs, address must be disassociated as a precondition before `release_address` can be performed.

pycloudlib/ec2/instance.py
```diff
@@ -81,9 +105,67 @@ def add_network_interface(self) -> str:
...
+        interface_id = self._create_network_interface(ipv6_address_count)
+        if ipv4_public_ip_count is not None:
+            if (
+                ipv4_address_count is None
+                or ipv4_address_count < ipv4_public_ip_count
+            ):
+                raise PycloudlibError(
+                    f"Invalid combination of"
+                    f" `ipv4_public_ip_count={ipv4_public_ip_count}` and"
+                    f" `ipv4_address_count={ipv4_address_count}`"
```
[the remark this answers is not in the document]
aciba90: Done.

pycloudlib/ec2/cloud.py
aciba90: Done.

pycloudlib/ec2/instance.py
aciba90: Double try-except added, thanks.

pycloudlib/ec2/instance.py
blackboxsw: Thanks for this @aciba90

pycloudlib/ec2/instance.py
```diff
@@ -81,9 +105,67 @@ def add_network_interface(self) -> str:
...
+                ipv4_address_count is None
+                or ipv4_address_count < ipv4_public_ip_count
+            ):
+                raise PycloudlibError(
+                    f"Invalid configuration:"
+                    f" `ipv4_public_ip_count={ipv4_public_ip_count}` cannot be"
+                    f" greater than `ipv4_address_count={ipv4_address_count}`"
+                )
+
+        if ipv4_address_count is not None and ipv4_address_count > 1:
```
blackboxsw, line 130: if ipv4_address_count == 1, then we aren't going to call allocate_address/associate_address below. In checking the launched instances,I'm already seeing two private ips before any additional allocation based on ipv4_address_count > 1. Do we already expect that initial IPv4 private address to be allocated and associated already without needing to call allocate_address or associate_address within the for i in range(ipv4_address_count) below? ? 

Also, if ipv4_address_count  == 2, we'll call allocate_address/associate_address twice because our range below is for i in range(ipv4_address_count). Does this actually add 2 extra private IPs to this interface even though it may already have 1 private IP associated with it? Should that range actually be `range(ipv4_address_count -1)` instead?

Also, in testing this PR with the corresponding draft PR in cloud-init/ec2-ipv6 on the integration test  tests/integration_tests/test_networking.py::test_ec2_multi_nic_reboot. I'm getting botocore errors that we are allocating too many addresses. I'm wondering if that's just on my account, or due to the approach we are taking to continue to allocate more addresses than are typically allowed for an instance 
```
E           botocore.exceptions.ClientError: An error occurred (AddressLimitExceeded) when calling the AllocateAddress operation: The maximum number of addresses has been reached.
```

pycloudlib/ec2/instance.py
```diff
@@ -81,9 +105,67 @@ def add_network_interface(self) -> str:
...
+                    nics_resp["NetworkInterfaces"][0]["PrivateIpAddress"]
+                ]
+                priv_ips.extend(
+                    (
+                        x["PrivateIpAddress"]
+                        for x in assignment_resp["AssignedPrivateIpAddresses"]
+                    )
+                )
+                for i in range(ipv4_address_count):
+                    allo = self._client.allocate_address(Domain="vpc")
```
blackboxsw, line 151: allocate_address allocates an elastic IP address. Shouldn't we only be doing this below in the `if associate_public_ip` scope? This is probably the reason I was seeing limits on public address allocation when running integration tests.

Don't we instead want `assign_private_ip_addresses` here  instead of the allocate_address/associate_address?

pycloudlib/ec2/instance.py
aciba90: I messed up, by copy-pasting, `ipv4_address_count` and `ipv4_public_count`. Expectations are that the default NIC (the one associated with the instance at launch) will have a private and public ipv4 by default so pycloudlib does not have to create them. I have fixed it, added comments and improved the documentation to reflect that.

pycloudlib/ec2/instance.py
aciba90: Yes, iterating over `range(1, ipv4_public_ip_count)` skipping the first NIC as it comes with a public ipv4 by default.

`associate_public_ip` is for a per-NIC global ip, as opposed to adding public ips attached to specific addreses.

aciba90:
Thanks, @blackboxsw, for the review and catching up the leaking resources! I have fixed it and added docs and comments about.

VERSION
```diff
@@ -1 +1 @@
-1!6.5.0
+1!6.6.0
```
blackboxsw, line 1:
```suggestion
1!6.7.0
```

pycloudlib/ec2/instance.py
```diff
@@ -81,9 +105,67 @@ def add_network_interface(self) -> str:
...
+                    _check_response(association)
+
+        if associate_public_ip:
+            allocation = self._client.allocate_address(Domain="vpc")
+            _check_response(allocation)
+            association = self._client.associate_address(
+                AllocationId=allocation["AllocationId"],
+                NetworkInterfaceId=interface_id,
+            )
+            _check_response(association)
```
blackboxsw, line 167: While this feels strange to provide a separate associate_public_ip, the use-case and the calls to the botocore API are different given we are trying to associate a public IP address to the primary private_IP on this interface_id. This works for the use-cases we have today and provides us with a simple mechanism to assign a single public IP to a given interface without having to also having to creation secondary private IPs.

blackboxsw (approved):
It would be nice to add one example to examples/ec2.py (or adapt examples/ec2.py::hot_add) to create/associate a secondary IPv4 address) with an explanatory comment.

There are no unit_tests covering pycludlib/ec2/  at the moment. But, it doesn't have to block this PR at this time. It is just something we should start thinking about as we continue to evolve cloud-specific instance APIs to increase certainty that the python logic is correct and 'documented' in unit tests.

pycloudlib/ec2/instance.py
aciba90: Thanks for the simplification.

aciba90:
Thanks, @blackboxsw, for the review and suggestions. I think I have addressed them all.

pycloudlib/azure/instance.py
```diff
@@ -264,6 +264,11 @@ def add_network_interface(self) -> str:
 
         Returns the private ip address of the new NIC.
         """
+        # pylint: disable=too-many-locals
+        us = datetime.datetime.now().strftime("%f")
+        # get ip address object
+        ip_address_obj = self._create_ip_address()
+        ip_config_name = f"{self.name}-{us}-ip-config"
```
blackboxsw, line 271: Are we doing this twice? It looks like this is peformed 20 links below as well

blackboxsw (changes requested):
I think there is a stray duplication of code in azure/instance.py. Once that is dropped, this is good to land.

pycloudlib/azure/instance.py
aciba90: Wrong merge conflicts resolution, amended.

blackboxsw (approved):
Ship it!

PR #357, opened 2024-04-09
fix(azure): make necessary checks before creating resources

CalvoM:
From testing with retries, we create a number of ips with each retry. We are avoiding this by performing checks before creating the resources.


aciba90 (approved):
LGTM, thanks! Could you, @CalvoM, bump the version again please?

PR #358, opened 2024-04-15
error: add new PycloudlibQuotaError to propagate quota errors

gjolly:
For the user to be able to retry on quota error, we need to be able to propagate this exception.

### DIFF
--- a/pycloudlib/azure/cloud.py
+++ b/pycloudlib/azure/cloud.py
@@ -20,6 +20,7 @@ from pycloudlib.errors import (
     InstanceNotFoundError,
     NetworkNotFoundError,
     PycloudlibError,
+    PycloudlibQuotaError,
     PycloudlibTimeoutError,
 )
 from pycloudlib.util import get_timestamped_tag, update_nested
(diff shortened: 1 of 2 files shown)


aciba90 (changes requested):
Thanks, @gjolly. There are some [linting issues](https://github.com/canonical/pycloudlib/actions/runs/8686020826/job/23816661973), could you please fix them?

In addition to that, I wonder if `OperationNotAllowed` really means quota errors in every case. Per https://learn.microsoft.com/en-us/troubleshoot/azure/virtual-machines/windows/error-messages, it looks like this error code could mean more things, have you researched this?

gjolly:
@aciba90 you are right, we need to check the message to check if it's a quota error or not. TBH there is no difference between checking the message in pycloudlib and mapping it to a specific exception and doing the same thing in the client. Thus I would rather do it in the client and avoid adding this logic in pycloudlib.

aciba90:
@gjolly alright, closing this PR then, thanks!

PR #362, opened 2024-04-18
fix(azure): Use correct ip when creating with multiple nics

CalvoM:
Use the primary nic ip address when creating an VM with multi nics.
Needed for https://github.com/canonical/cloud-init/pull/5180


aciba90 (approved):
LGTM, thanks!

PR #364, opened 2024-04-24
feat(ibm): add ability to choose from existing floating ips

a-dubs:
Currently, for IBM VPC, we only create floating IPs on the fly.
Adding the ability to choose an existing floating IP that has already
been reserved ahead of time, allows for launching instances using
already known IP addresses.


pycloudlib/ibm/instance.py
```diff
@@ -452,13 +452,15 @@ def __init__(
         instance: dict,
         floating_ip: Optional[dict] = None,
         username: Optional[str] = None,
+        using_existing_floating_ip: Optional[bool] = False,
```
aciba90, line 455: If the default is `False` and we do not have any meaning for `None`, why not:

```suggestion
        using_existing_floating_ip: bool = False,
```

pycloudlib/ibm/cloud.py
```diff
@@ -293,6 +323,7 @@ def launch(
             vpc: VPC to allocate the instance in. If not given, the instance
             username: username to use when connecting via SSH
             will be allocated in the zone's default VPC.
+            existing_floating_ip_name_includes: string, name to filter floating IPs by
```
aciba90, line 326: Could document here that `if using an existing floating ip, it won't be deleted`? Feel free to rephrase it please.

aciba90 (changes requested):
LGTM, thanks!

In addition to the inline comments, could you please rebase main and resolve the merge conflict?

pycloudlib/ibm/instance.py
a-dubs: Good suggestion!

pycloudlib/ibm/cloud.py
a-dubs: good idea. i modified the docstring explanation. let me know if you think this explanation is not sufficient

pycloudlib/ibm/cloud.py
```diff
@@ -261,6 +261,42 @@ def get_or_create_vpc(self, name: str) -> VPC:
...
+        floating_ips = list(
+            _iter_resources(
+                self._client.list_floating_ips,
+                resource_name="floating_ips",
+                # Select that match the desired name
+                filter_fn=lambda ip: name_includes in ip["name"],
+            )
+        )
+        if not floating_ips:
+            raise IBMException("No floating IPs found matching substring %s", name_includes)
```
aciba90, line 285: This syntax is for loggers. One way to correctly interpolate strings would be:

```suggestion
            raise IBMException(f"No floating IPs found matching substring `{name_includes}`")
```

pycloudlib/ibm/cloud.py
```diff
@@ -261,6 +261,42 @@ def get_or_create_vpc(self, name: str) -> VPC:
...
+        # filter out floating ips that are already associated with an instance
+        floating_ips = [ip for ip in floating_ips if "target" not in ip]
+        if not floating_ips:
+            raise IBMCapacityException(
```
aciba90, line 289: Same here:

```python
            raise IBMCapacityException(
                f"All floating IPs matching substring `{name_includes}` are already in use.",
            )
```

aciba90 (changes requested):
Thanks, @a-dubs, for addressing the comments.

In addition to the in-line comments, could you please perform a `git fetch upstream && git pull --rebase main` and resolve the conflicts with the `VERSION` file?

aciba90 (approved):
LGTM, thanks!

PR #365, opened 2024-04-29
Stricter-floating-ip-selection-per-zone

a-dubs:
My recent change to allow for using existing floating IPs didn't take into account the zone of the floating IP. As such, attempting to attach a floating IP from zone 3 but to an instance in zone 2 would fail. But now it will only choose floating IPs in the same zone. :D


TheRealFalcon (approved):
LGTM!

PR #366, opened 2024-04-29
feat(ibm): Allow floating IP in toml config

TheRealFalcon:
Allow behavior added in #364  to work via pycloudlib.toml.


aciba90 (approved):
LGTM, thanks!

PR #367, opened 2024-04-30
feat(ibm): check if floating IP is available before creating instance

a-dubs:
Previously, the instance would be created/launched and then the IP address would be attached. But if an existing floating IP is specified to be used, and no IP can be found, then an instance was created for nothing.


TheRealFalcon (approved):
LGTM!

Small nit because I noticed it on your previous PR too: with semantic versioning, you should only increment the last number if it's a backwards-compatible bug fix. Since this is a feature addition, I would for bumping the middle number instead. This doesn't need to be fixed here, but I'm mentioning it more as a future FYI.

PR #368, opened 2024-05-03
feat(oracle): ability to specify which fault domain to launch in

uhryniuk:
Fault domains are physical groupings of hardware within an availability domain for when a compute instance is going to be created. PR allows to specify which fault domain to use which is useful for low capacity instances. If no fault domain is provided (the default) then oci infra will automatically try to choose the best one.


pycloudlib/oci/instance.py
```diff
@@ -89,6 +90,13 @@ def instance_data(self):
         """Return JSON formatted details from OCI about this instance."""
         return self.compute_client.get_instance(self.instance_id).data
 
+    @property
+    def fault_domain(self):
+        """Obtain the fault domain the instnace resides in."""
```
a-dubs, line 95: Nit: fix typo

aciba90 (approved):
LGTM, thanks!

PR #370, opened 2024-05-07
feat(PEP561): distribute type hints

aciba90:
Allow consumers of the library to check type hints with tools like mypy.

https://peps.python.org/pep-0561/#packaging-type-information


TheRealFalcon (approved):
LGTM!

PR #371, opened 2024-05-07
feat: add IBM Softlayer to enable using IBM Classic Infrastructure

a-dubs:
IBM Classic Infrastructure actually has quite a robust SDK. Adding IBM CLassic/Softlayer to pycloudlib will provide immediate value to Canonical, and I'm sure this could be useful for other 3rd parties.


pycloudlib/ibm_softlayer/cloud.py
```diff
@@ -0,0 +1,492 @@
...
+        for key in self._ssh_key_manager.list_keys():
+            if key["label"] == name:
+                target_key = key
+                break
+        if target_key is None:
+            raise IBMSoftlayerException(f"Key with name {name} not found")
+        self._log.debug("Deleting SSH key: %s", name)
+        self._ssh_key_manager.delete_key(target_key["id"])
+
+    def create_default_security_groups(self) -> tuple[int, int]:
```
aciba90, line 350: [builtins.tuple](https://docs.python.org/3/library/stdtypes.html#tuple) is not usable as type hints in [python 3.8](https://docs.python.org/3.8/library/typing.html#annotating-tuples), which is the minimum supported version of pycloulib. Could we use typing.Tuple?

```suggestion
    def create_default_security_groups(self) -> Tuple[int, int]:
```

https://docs.python.org/3/library/typing.html#annotating-tuples

pycloudlib/ibm_softlayer/instance.py
```diff
@@ -0,0 +1,374 @@
...
+            ) from e
+        except Exception as e:
+            logger.error(
+                "Unexpected error while verifying instance configuration: %s",
+                e,
+            )
+            raise IBMSoftlayerException(
+                f"Unexpected error while verifying instance configuration: {e}"
+            ) from e
+        else:
```
aciba90, line 132: Nit: As every except clause raises, could we delete the else and deindent the last block of code?

pycloudlib/ibm_softlayer/instance.py
```diff
@@ -0,0 +1,374 @@
...
+            return full_raw_instance_info
+
+    def get_boot_id(self):
+        """Get the instance boot_id.
+
+        Returns:
+            string with the boot UUID
+        """
+        result = self.execute(
+            "cat /proc/sys/kernel/random/boot_id", no_log=True
```
aciba90, line 155: Why no_log? Is the boot_id sensitive info? If so, cloud we leave a breadcrumb comment explaining it?

pycloudlib/ibm_softlayer/instance.py
```diff
@@ -0,0 +1,374 @@
...
+            )
+        return result
+
+    def console_log(self):
+        """Return the instance console log.
+
+        Raises NotImplementedError if the cloud does not support fetching the
+        console log for this instance.
+        """
+        # IBM Classic does not support console log
```
aciba90, line 170: Nit: This comment is already documented in the exception. It feels redundant.

```suggestion
```

pycloudlib/ibm_softlayer/cloud.py
```diff
@@ -0,0 +1,492 @@
...
+from pycloudlib.ibm_softlayer.instance import IBMSoftlayerInstance
+from pycloudlib.instance import BaseInstance
+
+
+class IBMSoftlayer(BaseCloud):
+    """IBM Softlayer Class."""
+
+    _type = "ibmsoftlayer"
+
+    VALID_REGIONS = [
```
aciba90, line 20: Instead of hardcoding a set of valid regions, couldn't we try-except on the sdk calls that use the region and fail? With a hardcoded list we will need to update it every time it changes.

pycloudlib/ibm_softlayer/cloud.py
```diff
@@ -0,0 +1,492 @@
...
+        )
+        self._log.debug("Added rules to security groups.")
+        return public_security_group_id, private_security_group_id
+
+    def _create_security_group(self, name: str, description: str) -> int:
+        """
+        Create a security group.
+
+        Args:
+            name (str): The name of the security group.
```
aciba90, line 407: As the types are already documented as type hints, could we remove them from the docstrings? It feels repetitive and very likely to end up out of sync.

If we do so, please make it in all the docstrings.

pycloudlib/ibm_softlayer/cloud.py
```diff
@@ -0,0 +1,492 @@
...
+        self.created_security_groups.append(new_group["id"])
+        self._log.debug(
+            "Created new security group %s: %s", name, new_group["id"]
+        )
+        return new_group["id"]
+
+    def _add_rules_to_security_group(
+        self,
+        group_id: int,
+        directions: list[str],
```
aciba90, line 425:
```suggestion
        directions: List[str],
```

pycloudlib/ibm_softlayer/cloud.py
```diff
@@ -0,0 +1,492 @@
...
+            A single string with the latest released image ID for the
+            specified release.
+
+        """
+        public_images_gen = self._image_manager.list_public_images(
+            name=f"*{release}*"
+        )
+        public_images = list(public_images_gen)
+        if not public_images:
+            raise Exception(f"No public images found for {release}")
```
aciba90, line 106: In order to respect https://github.com/canonical/pycloudlib/issues/263, could we raise more concrete exceptions? 

```suggestion
            raise IBMSoftlayerException(f"No public images found for {release}")
```

Please, make sure this holds for the whole PR.

aciba90 (changes requested):
Thanks, @a-dubs, for adding this feature!

I have left in-line comments / requests. In addition to that, could we please add a documentation section to `docs/clouds`, explaining at least what configuration is required to use the new cloud (api keys, etc)?

pycloudlib/ibm_softlayer/cloud.py
```diff
@@ -0,0 +1,492 @@
...
+        private_images_gen = self._image_manager.list_private_images(
+            name=f"*{name}*"
+        )
+        private_images = list(private_images_gen)
+        if not private_images:
+            raise Exception(f"No private images found for {name}")
+        return private_images[0]["globalIdentifier"]
+
+    def get_instance(
+        self, instance_id, *, username: Optional[str] = None, **kwargs
```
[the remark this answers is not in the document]
a-dubs: hmm i thought mypy would get angry about the function signature changing but apparently not! thanks.

pycloudlib/ibm_softlayer/cloud.py
```diff
@@ -0,0 +1,492 @@
...
+        Args:
+            instance: Instance to snapshot
+            clean: run instance clean method before taking snapshot
+
+        Returns:
+            An image id
+
+        """
+        if clean:
+            instance.clean()
```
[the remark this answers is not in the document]
a-dubs: That's a good question. I will have to test this out. I naively just copied off the existing ibm vpc implementation
without giving it much thought.

pycloudlib/ibm_softlayer/cloud.py
a-dubs: Oh really? I was completely unaware that it's that new of a feature. Thanks for pointing this out

pycloudlib/ibm_softlayer/cloud.py
a-dubs: yeah good call. will do!

pycloudlib/ibm_softlayer/instance.py
a-dubs: haha good nit! will do

pycloudlib/ibm_softlayer/instance.py
a-dubs: hahaha agreed

pycloudlib/ibm_softlayer/instance.py
a-dubs: okay so apparently, i just copied this from the BaseInstance class thinking that i would need to modify it but never
ended up needing to. And so the no log is default behavior for all clouds apparently. So i just removed the duplicate
code from the IBMSoftlayerInstance Class.

pycloudlib/ibm_softlayer/cloud.py
a-dubs: thanks for pointing this out! i didnt realize how sloppy i was being with those.

pycloudlib/instance.py
```diff
@@ -485,19 +485,25 @@ def _tmpfile(self):
...
+    def _wait_for_execute(self, old_boot_id=None, timeout: int = 40):
+        """
+        Wait until we can execute a command in the instance.
+
+        Args:
+            old_boot_id: boot id before restart. If provided, we will wait
+            until we find a new boot id
+            timeout: time to wait for instance to be reachable in minutes
 
-        If old_boot_id is specified, we use its value to wait until we
```
aciba90, line 491: I think this sentence is still relevant, could we please keep it?

pycloudlib/ibm_softlayer/cloud.py
aciba90: This is still unfixed in some places.

pycloudlib/ibm_softlayer/cloud.py
aciba90: There are still a couple of `ValueError`s that probably would be better as an `IBMSoftlayerException` exception.

pycloudlib/ibm_softlayer/cloud.py
```diff
@@ -0,0 +1,500 @@
...
+                "Invalid disk_size given. "
+                "disk_size must be either '25G' or '100G'"
+            )
+        if user_data:
+            self._log.error(
+                "IBM Softlayer does not support user data for instance "
+                "launch. No user data will be used."
+                )
+
+        # check if image_id is a GID by checking if it contains hypens
```
aciba90, line 234:
```suggestion
        # check if image_id is a GID by checking if it contains hyphens
```

pycloudlib/ibm_softlayer/instance.py
```diff
@@ -0,0 +1,361 @@
+# This file is part of pycloudlib. See LICENSE file for license information.
+# pylint: disable=too-many-public-methods
+"""IBM Softlayer (Classic Infrastructure) instance and VPC classes."""
+
+import logging
+from typing import List, Optional
+
+# from SoftLayer import BaseClient, SoftLayerAPIError, VSManager
```
aciba90, line 8:
```suggestion
```

VERSION
```diff
@@ -1 +1 @@
-1!7.2.0
+1!8.0.0
```
aciba90, line 1: Why a breaking change?

aciba90 (changes requested):
Thanks, @a-dubs, for addressing the concerns and fixing stuff.

Apart from in-line comments, what is still missing is:

> could we please add a documentation section to docs/clouds, explaining at least what configuration is required to use the new cloud (api keys, etc)?

Documentation about IBM classic's specificities.

Another question: we refer to IBM classic / softlayer interchangeably, with the code and within log messages and exceptions. I wonder if it would be better to stick to one of the terms globally, or if not, clearly document it somewhere.

pycloudlib/instance.py
a-dubs: I kept this information, just reworded above in the new docstring. Is the rewording poor/is detail lost?

VERSION
a-dubs: ah good point, this is not breaking.

a-dubs:
> Another question: we refer to IBM classic / softlayer interchangeably, with the code and within log messages and exceptions. I wonder if it would be better to stick to one of the terms globally, or if not, clearly document it somewhere.

Good question. Unfortunately, at this point, I really think I'd rather refer to it as IBM Classic. This is mostly because despite the IBM provided sdk and api being called softlayer, this is not obvious to an end user, who sees either "IBM VPC" or "IBM Classic" when browsing IBM Cloud Console, with no obvious mention of softlayer. So if you agree, I'll go through and mass rename everything to IBM Classic. I would like the naming to be consistent throughout for sure.

Also, when I write up the docs for this new IBM ~Softlayer~ Classic cloud, I'll be sure to have  a small preamble in there that explains this naming confusion.

aciba90:
+1 to the renaming to classic, thanks!

aciba90 (changes requested):
Many thanks, @a-dubs, for the renaming and documentation efforts, this is getting in shape.

I left in-line comments and in addition to that we need to:

1. Include the new cloud config in `pycloudlib.toml.template`.
2. Remove Softlayer from `TODO.md`.

pycloudlib/ibm_classic/cloud.py
```diff
@@ -0,0 +1,500 @@
...
+        sl_username: Optional[str] = None,
+        sl_api_key: Optional[str] = None,
+    ):
+        """Initialize base cloud class.
+
+        Args:
+            tag: string used to name and tag resources with
+            timestamp_suffix: Append a timestamped suffix to the tag string.
+            config_file: path to pycloudlib configuration file
+            sl_username: IBM Softlayer username
```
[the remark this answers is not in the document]
a-dubs: I guess we could yeah. I'm slightly hesitant to because I don't want people trying to use their normal username (the softlayer username is entirely different) and also just wanted to distinguish the softlayer api key from the vpc api key. But I can agree to removing the softlayer naming, I will just want to add a section to the documentation explaining how to find these credentials (which i should really do anyway)

pycloudlib/ibm_classic/cloud.py
```diff
@@ -0,0 +1,500 @@
...
+        self._network_manager = SoftLayer.NetworkManager(client=self._client)
+
+    def delete_image(self, image_id: str, **kwargs):
+        """Delete an image.
+
+        Args:
+            image_id: string, id of the image to delete
+            **kwargs: dictionary of other arguments to pass to delete_image
+        """
+        self._image_manager.delete_image(int(image_id))
```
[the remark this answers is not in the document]
a-dubs: thank for catching this!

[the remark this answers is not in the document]
a-dubs: okay so i changed image_id to a int in the funtion signature and mypy is now throwing a fit :(

pycloudlib.toml.template
```diff
@@ -57,6 +57,14 @@ project = ""  # glcoud config get-value project
 # private_key_path = ""  # Defaults to 'public_key_path' without the '.pub'
 # key_name = ""  # Defaults to your username if not set
 
+[ibm_classic]
+sl_username = ""
+sl_api_key = ""
```
aciba90, line 62:
```suggestion
api_key = ""
```

pycloudlib/ibm_classic/cloud.py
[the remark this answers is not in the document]
aciba90: I do not see any mypy error with:

```suggestion
        self._image_manager.delete_image(image_id)
```

Could we change it to this? Otherwise, this going to error if image_id is not convertible to int.

a-dubs:
@aciba90 regarding the delete_image function, mypy only throws an error when I change the function definition to have image_id be an int. Also if `self._image_manager.delete_image(int(image_id))` raises an error on failing to cast an int, then that is good, because the softlayer image id is numeric and should never fail to be cast to an int. Should I wrap this in a try catch and raise a pycloudlib IBMClassicException  though?

tests/unit_tests/ibm_softlayer/test_cloud.py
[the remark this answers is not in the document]
a-dubs: good catch!

tests/unit_tests/ibm_classic/test_cloud.py
aciba90, line 1: Could you add an `test/unit_tests/ibm_classic/__init__.py` to avoid https://github.com/canonical/pycloudlib/actions/runs/9662468005/job/26652416577?pr=371?

a-dubs: ah good catch, thanks! done :)

aciba90:
> @aciba90 regarding the delete_image function, mypy only throws an error when I change the function definition to have image_id be an int. Also if `self._image_manager.delete_image(int(image_id))` raises an error on failing to cast an int, then that is good, because the softlayer image id is numeric and should never fail to be cast to an int. Should I wrap this in a try catch and raise a pycloudlib IBMClassicException though?

Ahh, I see your point now, then I guess it does not matter that much. We could:

1. `self._image_manager.delete_image(int(image_id))` (as-is) with or without exception handling
2. Use https://docs.python.org/3/library/typing.html#typing.cast and `self._image_manager.delete_image(image_id)`
3. `self._image_manager.delete_image(image_id)`
4. Modify the parent and children types associated to image_id to `Union[str, int]`

(1) does a superfluous type conversion at runtime
(2) is type-checking the softlayer API / SDK
(3) is probably obscure
(4) unnecessarily complex

I leave it up to you to decide.

a-dubs:
I'll go for (1) I guess then! Feels like the better solution among the rest. Thanks for your input!

a-dubs:
I'm doing one final round of testing to make sure everything functions as desired. I will report back when done.

aciba90 (approved):
LGTM, thanks for the long work here!

a-dubs:
IT FINALLY WORKED RAHHHH LETS GOOOO



                                 #@@@@@@@@@@@@@@@#                                                  
                           *@@@*.........::::::::::-%@@@*                                           
                        %@@:..........:**=:::::::::::::::=@@%                                       
                     @@+.....-@@@#:....::::::*@%::=@:::::::::+@@                                    
                  #@-....@%:..........:::::::::::-@=:#-:::::::::%@@                                 
              +@@@%...*@..............::::-+%%#=:::::#:#:-:::::::::@@#                              
           #@#.....%@@...........+@@@@@@@@#:::::::::-::::%@::::::::::-@#                            
    #@@@@%@@@*..#..%:.......:@@@@@@@*:::::::::::::-@:::::::@:::::::::::-@+                          
  #@......%.....:.@......*@@@@@@#.::::::::::::::@@%@::#@#:::::-*+-:::::-:-:-                        
  @@@:........:%.-.........@@@@...:-%@@@@@@@@@@@+*@@@-:@:-@::..+*=...++--+.#%                       
  @.@*:......*#..........:@@@@@@@#:::::::::::::::-:::-@:*::::.:::::::-%+-#---@                      
  @..@:......@*.........@@@@*...::::-:::::::+%#=---:=-:@:::+.+::::::::::------@                     
  @...@......++....%%@@@@@:....:::::::::::::::::--:::::#:::=.::::::::::--------@#                   
  @=..-@.....:@......*@@@@@@@@@@@@@@@@@*#@::::::*#:::-*::::.=:::%::::::---------@*                  
  @.@%:.#@....@@@@@@*:.......:::::::::::+@@@-::::-+=%@%::::.+::::-::::-----------@#                 
  @.......*@@@-....+@@@@@@%##########%%@@@#@@@:::::@@@:::::.+::::-+::-------------@                 
 @.....@@#..:%@@@##########################%@#@%::::@@:::::::::::::#---------------@                
#-...@..:@@%#######@..@@###@. .@%##@..@##@: .@%#@::::@:::::+.::-:::-----------------@               
@..#-.@@#@@@######@.  ..@%=    .%@@.  .%@.   .@#@@::::@::::= =-:::------------------@               
@.:+%@#%@. ..@%#%@.     .+.     .=.    :.     @. .@::::*::::.*:::----#--*-----------=@              
 @%@###@.    ..#*.       %.     ..     #.    ..   .@-:::::::*.-:------=--@-----------@              
  @@.#@=                 %.     .            ..  .*.@-::-@::-.*:------%---@----------@%             
   @. .:                 ..    .#     :.   ..   .%. :@-::=@-:=.--------@--+@---------%@             
   @. .@                 =.    :.    .=  .::  .:...%@@@-::*@--.*-------*---@=--------+@             
    @. ..             ...--. .*%@@#@@@@@@@@@@@@@@@@@@@@@-::#@**.--------@---@---------@             
    *#.        .+@@@%#@@:..%+.  .@%%@@@@@@@%@@@@@@@@@@@@@:::@=*-.-------%---@---------@             
      @.  .@#-...+.   .+   =.  .@@@@@@@@@@@@@@@@@@@@@@@@@@---@-=.=------=@---%----**+-@             
       %@.@.     .   .:. .:=..:@@@%%%%%%%%%@@@@@@@@@@%%%@@@---@-=.#------@---%--#.#=#. +            
          @@#...#=..=@@%%%%%%%@@@@@@@@@@@@@@@@@@@%%%%%%%%@@@---@-%.+-----@+---==.*---=@             
           @@@@@@@@@@@@@@@@@@@@@@@@@%%%@@%%%%%@%%%%%%%%%%%@@@---@@*.=----+@---=:.----+@             
            @-.=@%#%@@@@@@%%%%%%%@@@@@%##%%%@%%%%%%%%%%%%%@@@@---@@#.*----@----.*----@@             
             @.:#@%#%%%@%@@@@@@@@%%@%#####%@%%%%%%%%%%%%%@@@@@@--=@--..---@=--=.-----@              
             #@::@%@%%%%@@%%%%@@@@@#######@%%%%%%%%%%%%@@@@@@@@@--=---*..=%%-*.+----#@              
              @=::@#@%%%%%@@@@%%@@#######%##%%%%%%%%%%@@@@@@@@@@@-------=.....=--#--@               
               @:::@#%#%%%%@@@@@%#######%####%%%%%%%@@@@@%@@@@@@@@--#------@-----@-@                
               *@::-%%#%%%%@@@@##############%%%%%@@@=.    .#@@@@@---=-----@-----@@*                
                @=::#%@#%%%@@@###############%%%@%.--.       .%@@@+--------@----@@%                 
                 @:::%@@@@@.=@###########%@@#@..       .      .-@@---*-----@----@@                  
                @@+:@@%*.. .@##%%..+@@*.                #-... .=@----@-----@---@@                   
                @+@@    ..:                   .*.        @####%#----@------@--@%                    
               %-%@.                   .       #@@@=.  .:@##@#----@#@-----#@@@                      
               @*@                     #       :%####%@@%@@----@@*+@-----=@@#                       
              @#*.                     .+.    .@#####@@#-%=*%#*+@=#-----%@%                         
             #@*@. .+%.         .@+.   .@%@%:.@#%@@%+==+@%==++@-------+@%                           
             @:@@..@%@%  .@@.  .%%##@#*@###@@@*-%=++==+%+@@#==+%--%@@@*                             
             @::@@%###@%-@###@:%%####@@@@+---=-----+===============+*%                              
             #@:::@@%######%@@@@@@*-:-:::--+#*=-=-+%=%+===+*%#====++++*%%%%                         
               @@:::::::-:::::::+%@@@-::------------------%#+%#  #%#++++%+                          
                  *@@@@@@@@@@@@@@*@=@+@=**----------%@@@@*             ##+++%*                      
                                #%@@@@@@@@@@@@@@%#*                        +%*++%#                  
                                                                                 #%#%               
                                                                                                    
                                                                                          %%%       
                                                                                             %+*%   
                                                                                                #%*#

a-dubs:
Proof of working run (in case i gaslight myself and forget if it worked or not):
[ibm-classic-example-successful-run.log](https://github.com/user-attachments/files/15991002/ibm-classic-example-successful-run.log)

aciba90:
Sweet @a-dubs, could you please resolve conflicts in `VERSION`?

aciba90:
Latest commit: https://github.com/canonical/pycloudlib/pull/371/commits/50089ec1954b0fe21d1302c11596e986bce6de6d contains commented code and non-trivial changes. Is that all required?

a-dubs:
@aciba90 fixed version

pycloudlib/ibm_softlayer/cloud.py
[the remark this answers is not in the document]
a-dubs: cleaning before snapshotting seems to work! i have baked snapshotting into the example script

Issue #373, opened 2024-05-09
Migrate away from deprecated docstring linter


aciba90:
Pyclouldlib uses [flake8-docstrings](https://github.com/canonical/pycloudlib/blob/a866c23491847e1bad3e6082088be031a07c3cd7/tox.ini#L10), which is a plugin that integrates [pydocstyle](https://github.com/pycqa/pydocstyle) with flake8.

Pydocstyle has been archived and deprecated in favor of ruff, see https://github.com/PyCQA/pydocstyle/commit/8d0cdfc93e86e096bb5753f07dc5c7c373e63837.

> Migrate to https://docs.astral.sh/ruff/rules/#pydocstyle-d for a seamless transition.

PR #374, opened 2024-05-09
refactor: migrate to ruff

aciba90:
Refactor black, isort, flake8 to use ruff.
Apply ruff format.

Fixes: GH-373


aciba90:
> what is the benefit of using ruff vs black and isort?

Generally the advantage of ruff is speed. In the case of black and isort there is not a huge difference, but having a single tool reduces complexity and dependencies.

pyproject.toml
```diff
@@ -31,3 +27,34 @@ module = [
   "paramiko.*",
 ]
 ignore_missing_imports = true
+
```
TheRealFalcon, line 30: We should remove the `[tool.black]` section too, correct?

pyproject.toml
```diff
@@ -31,3 +27,34 @@ module = [
...
+  "I",  # isort
+  "PL",  # pylint
+]
+
+# Commented rules were not available in ruff. Enable them as they are added to ruff
+# See: https://github.com/astral-sh/ruff/issues/970
+ignore = [
+  "N815",  # Invalid name
+  "SIM300",  # misplaced-comparison-constant
+  # "PLC0209",  # Formatting a regular string which could be a f-string (consider-using-f-string)
```
TheRealFalcon, line 49: Were the commented lines carried over from flake8? Can we remove them now?

.pylintrc
```diff
@@ -1,20 +0,0 @@
-[MASTER]
```
TheRealFalcon, line 1: Pylint still does a fair number of things that ruff cannot (and probably never will). I don't think we should be removing pylint here.

TheRealFalcon (changes requested):
Thanks for this! See my inline comments.

pyproject.toml
aciba90: Removed.

pyproject.toml
aciba90: Those were not yet available pylint rules, see the header comment. This is not needed anymore as we reverted to keep pylint.

.pylintrc
aciba90: Reverted.

tox.ini
```diff
@@ -106,13 +97,8 @@ deps = {[testenv:integration-tests-ci]deps}
...
-[flake8]
-# E203: whitespace before ':' ... This goes against pep8 and black formatting
-# W503: line break before binary operator
-ignore = E203, W503
-
 [pytest]
 testpaths = tests/unit_tests
 markers =
-    ci: run test on as part of continous integration
+    ci: run test on as part of continouos integration
```
TheRealFalcon, line 103: Nit:
```suggestion
    ci: run test on as part of continuous integration
```

TheRealFalcon (approved):
One nit inline, but otherwise LGTM!

tox.ini
aciba90: Thanks, fixed.

PR #375, opened 2024-05-15
feat(ibm): improve ibm vpc baremetal behaviour

a-dubs:
## Commit Description
Previously, if baremetal instances could not be launched due to
lack of capacity on the IBM Cloud, pycloudlib would fail to detect
this and just assume that the instance booted correctly.
Now, pycloudlib will return a capacity exception once this is
detected, allowing for consumers of pycloudlib to properly
implement retries for BM instances on IBM VPC.

## Description of Testing:
This was tested both through an external program that leverages pycloudlib, and through a modified version of the ibm example script. The external program validated the use case of implementing retries for baremetal.


pycloudlib/ibm/instance.py
```diff
@@ -746,9 +758,38 @@ def start(self, wait=True):
...
+            else:
+                raise IBMException(
+                    f"Instance {self.id} failed to start: "
+                    f"{self._instance['status_reasons'][0]['message']}"
+                )
+
     def _wait_for_instance_start(self, **kwargs):
         """Wait for the cloud instance to be up."""
-        self._wait_for_status(_Status.RUNNING)
+        # if self.b
```
aciba90, line 787:
```suggestion
```

pycloudlib/ibm/instance.py
```diff
@@ -746,9 +758,38 @@ def start(self, wait=True):
...
+        Check if the instance failed to start and raise an exception if so.
+
+        Raises:
+            IBMCapacityException: If the instance failed to start due to
+                capacity issues.
+            IBMException: If the instance failed to start due to other reasons.
+        """
+        if self._instance["status"] == _Status.FAILED.value:
+            if any(
+                "capacity" in reason["code"]
```
aciba90, line 772: Do we have a link pointing for docs covering this? If so, would you mind adding it for reference?

aciba90 (changes requested):
Thanks, @a-dubs, for improving this!

LGTM, could you please fix the inline commnets, the pylint error and would you mind adding a bit of context in the commit message explaining what we are improving?

pycloudlib/ibm/instance.py
a-dubs: I don't have any docs for this. I discovered this by manually launching and inspecting instance status via the API/SDK.

pycloudlib/ibm/instance.py
```diff
@@ -746,9 +758,38 @@ def start(self, wait=True):
...
+                f"{self._instance['status_reasons'][0]['message']}"
+            )
+
     def _wait_for_instance_start(self, **kwargs):
         """Wait for the cloud instance to be up."""
-        self._wait_for_status(_Status.RUNNING)
+        # if self.b
+        self._wait_for_status(
+            _Status.RUNNING,
+            sleep_seconds=900,
```
aciba90, line 790: This is good for now. But, the `900` should be configurable from a consumer POV, as it might be not flexible enough.

aciba90 (approved):
LGTM, thanks!

pycloudlib/ibm/instance.py
a-dubs: Okay I'll look at adding this sometime soon. Thanks!!

PR #380, opened 2024-05-31
feat: Replace version map with distro-info

TheRealFalcon:
Every 6 months we add another entry to
UBUNTU_RELEASE_VERSION_MAP. Instead, call distro-info to be more future proof.


blackboxsw (approved):
Minor change, then +1

pycloudlib/util.py
```diff
@@ -20,22 +20,40 @@
...
+    >>> release_version_from_series("noble")
+    "24.04"
+    """
+    try:
+        return subprocess.run(
+            ["distro-info", "--release", f"--series={series}"],
+            check=True,
+            capture_output=True,
+            universal_newlines=True,
+        ).stdout.strip()
```
blackboxsw, line 50: This returns `24.04 LTS` on noble and other LTS releases. I think we want to trim that content too as that space will give us problems in instance names and the extra lengths of instance names gives us probs on lxc integration tests in pro I believe.
```suggestion
        return subprocess.run(
            ["distro-info", "--release", f"--series={series}"],
            check=True,
            capture_output=True,
            universal_newlines=True,
        ).stdout.split()[0]
```

TheRealFalcon: Thanks!

holmanb (changes requested):
> Every 6 months we add another entry to UBUNTU_RELEASE_VERSION_MAP. Instead, call distro-info to be more future proof.

This is an Ubuntu-only dependency. I like to run integration tests on non-Ubuntu hosts sometimes.

-1 from me

TheRealFalcon:
@holmanb 

> This is an Ubuntu-only dependency

I'm replacing one Ubuntu-only dependency with another Ubuntu-only dependency. Image lookups have always been Ubuntu-only. That didn't prevent anybody from specifying their own image id. Neither will this.

holmanb:
>> This is an Ubuntu-only dependency
>
> I'm replacing one Ubuntu-only dependency with another Ubuntu-only dependency. Image lookups have always been Ubuntu-only. That didn't prevent anybody from specifying their own image id. Neither will this.

You are describing launching non-Ubuntu images, and yes currently we only support launching Ubuntu via these lookups. That isn't what breaks by this change.

As I said before, I like to run our integration tests (which use pycloudlib) **on non-Ubuntu hosts** sometimes. Requiring an Ubuntu-only dependency will mean that now if one wants to run the integration test suite, they will not be able to launch it from a Debian / Arch / Redhat system.

TheRealFalcon:
> they will not be able to launch it from a Debian / Arch / Redhat system

What about this code is preventing that?

TheRealFalcon:
Oh, I see...specifically looking up Ubuntu from non-Ubuntu.

holmanb:
@TheRealFalcon I'm experiencing deja-vu

https://github.com/canonical/cloud-init/pull/4586

PR #382, opened 2024-06-03
fix(ec2): Define GP3 releases separately from LTS releases

TheRealFalcon:
They are completely unrelated. They just happened to overlap in the
past.
Also drop lunar from the list as it is unsupported.
The previous commit that updated the list of LTS releases uncovered
this bug.


a-dubs:
Changes look good!

pycloudlib/ec2/cloud.py
```diff
@@ -134,11 +137,7 @@ def released_image(
     def _get_name_for_image_type(
         self, release: str, image_type: ImageType, daily: bool
     ):
-        if release in LTS_RELEASES + ["lunar"]:
-            # Images before mantic don't have gp3 disk type
-            disk_type = "hvm-ssd"
-        else:  # Mantic and later
-            disk_type = "hvm-ssd-gp3"
+        disk_type = "hvm-ssd" if release in NO_GP3_RELEASES else "hvm-ssd-gp3"
```
a-dubs, line 140: beautiful ternary :fire:

blackboxsw (changes requested):
This is very ubuntu-focused here and then places any non-ubuntu release name in the camp of getting "hvm-ssd-gp3" disk_type. Should we really be ensuring that we 
are in the intersection of  known pycloudlib.util.UBUNTU_RELEASE_VERSION_MAP and not in  NO_GP3_RELEASES?

TheRealFalcon:
> places any non-ubuntu release name in the camp of getting "hvm-ssd-gp3" disk_type

I seem to have a blind spot for seeing these kinds of issues, but was this not already the case? The old code selected the disk based on ubuntu release. The new code just stores those releases in a separate list so the Ubuntu LTS list can evolve separately...but the old code still relied on Ubuntu releases. I don't think I functionally changed anything here other than dropping Lunar from the list.

Furthermore, ALL release lookup is currently ubuntu-specific. I know of no plans of changing that anytime soon.

It seems you're asking for an enhancement whereas I'm just trying to fix a bug that prevents noble from launching on ec2.

blackboxsw:
> Furthermore, ALL release lookup is currently ubuntu-specific. I know of no plans of changing that anytime soon.

That works for me I was just thinking, that since we are touching a space that we know is not going to work great on non-ubuntu distros, we can an opportunity to set a specific default for non-ubuntu releases. But the conditional you have will set hvm-ssd-gp3 by default for any non-ubuntu release name. Honestly this is probably the 'right' default anyway for newer images, so I guess there is nothing really to fix here at the moment. We can enhance this conditional for other distros when pycloudlib grows support for multi-distro testing.

blackboxsw (approved):
Leaving other-distro enhancements to when we undertake the effort broadly. The current defaults for disk_type look good for the moment for non-ubuntu distro images.

PR #385, opened 2024-06-18
typing: add no_implicit_optional lint

aciba90:
Explicit optional types are preferred and the default in mypy >= 0.980[1].

Add the config to check for this and fix any instance of the lint error.

References:
[1] https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_optional


TheRealFalcon (approved):
LGTM! I trust you can bump the version accordingly.

PR #386, opened 2024-06-18
typing: add check_untyped_defs

aciba90:
Add the lint to check bodies of untyped definitions[1] which is enabled by default in mypy v1.0.0[2].

Disable modules that do not pass this check with the intention to fix them in the future but not add new modules or make the lint not pass for already passing ones.

Rerefences:
[1] https://mypy.readthedocs.io/en/stable/config_file.html#confval-check_untyped_defs [2] https://github.com/pandas-dev/pandas/commit/20e4c186191859dcde2437edacaee43d8d34dc46


pyproject.toml
```diff
@@ -3,9 +3,10 @@ requires = ["setuptools"]
 build-backend = "setuptools.build_meta"
 
 [tool.mypy]
+check_untyped_defs = true
 follow_imports = "silent"
-warn_unused_ignores = "true"
 warn_redundant_casts = "true"
+warn_unused_ignores = "true"
```
holmanb, line 9: unnecessary diff?

aciba90: I sorted them lexicographically. I can revert if unnecessary.

holmanb (approved):
LGTM

PR #392, opened 2024-07-10
feat: Allow multiple launches from single IBM instance

TheRealFalcon:
The default naming convention for new launches from the IBM cloud class used the same name for every launch from that instance. IBM cloud doesn't allow this, so ensure names are unique.


pycloudlib/ibm/cloud.py
```diff
@@ -72,6 +73,7 @@ def __init__(
 
         api_key = api_key or self.config.get("api_key")
         authenticator = IAMAuthenticator(api_key)
+        self.instance_counter = itertools.count()
```
blackboxsw, line 76: 1 for fun
```suggestion
        self.instance_counter = itertools.count(1)
```

pycloudlib/ibm/cloud.py
```diff
@@ -350,7 +352,7 @@ def launch(
 
         vpc = vpc or self.vpc
         floating_ip_name = f"{name}-fi" if name else None
-        name = name or f"{self.tag}-vm"
+        name = name or f"{self.tag}-{next(self.instance_counter)}vm"
```
blackboxsw, line 355:
```suggestion
        name = name or f"{self.tag}-vm{next(self.instance_counter)}"
```

blackboxsw (approved):
just bikesheds, LGTM either way. I know we are pressed for instance name length on some platforms, so avoiding excess hyphens may come into play at some point if we have datetimestamps in our tags. But looks good

a-dubs (approved):
Changes look good. So this is for the use case where you keep the same cloud object and try to launch multiple instances from it?

TheRealFalcon:
> Changes look good. So this is for the use case where you keep the same cloud object and try to launch multiple instances from it?

Yep!

PR #395, opened 2024-08-07
Don't swallow exceptions  when _wait_for_execute raises

TheRealFalcon:
```
fix: Don't swallow exceptions  when _wait_for_execute raises
    
The swallowed exception makes it hard to know why we're waiting
forever. Logging at debug means this shouldn't consumer logs unless
they want more verbose logs.
```


holmanb (approved):
Looks good, thanks for this @TheRealFalcon!

a-dubs (approved):
Great change! :rocket:

Issue #396, opened 2024-08-13
Make name/tag validation cloud-specific


a-dubs:
I had an issue launching an instance due to the tag i specified, and so from this incident, have realized two things should be done:

1. Implement cloud-specific tag/instance naming validation so that if one cloud is stricter than others, it doesn't impede on other clouds. 
2. For each cloud, the list of restriction should be explicitly enforced in an itemized way such that if the name must be <64 chars long and all lowercase, if 65 chars are given, then the error message should read "Name must be <64 characters long" or something like that.

By making name/tag validation cloud specific, the user will have greater flexibility on clouds with less restrictions, and the user experience will be improved by better error messages.

PR #398, opened 2024-08-14
feat!: add cloud specific tag validation

a-dubs:
Revamped base class tag validation function and added cloud
specific tag validation functions for IBM VPC and IBM Classic.

BREAKING CHANGE: Remove pycloudlib.util.validate_tag

Fixes GH-396


aciba90 (approved):
LGTM as is, thanks for adding unit-tests!

A version bump is required prior to merging.

Some points that might or might not be address in this PR, other PR:

1. The validation functions raise `ValueErrors`, we should think specialized exceptions are required to conform with https://github.com/canonical/pycloudlib/issues/263.
2. As-is now, rules are duplicated across different clouds, we might consider to re-architect this to make them reusable.

a-dubs:
I completely agree with the desire for more code re-usability. But I decided that for now, I didn't want to worry about over-engineering a solution that just might end up adding more complexity.

a-dubs:
@aciba90 I will change the `ValueErrors` to pycloudlib exceptions to conform with #263. I appreciate you pointing this out, especially since I am repeat offender for this issue. I thought about changing it to a pycloudlib exception but just decided to keep it as `ValueError` since that's what was raised previously and I just wanted to maintain existing functionality in that sense.

a-dubs:
@aciba90 I created a new type of Pycloudlib error and am raising this instead. This helped eliminate some redundant code with the generation of the error message. I also moved the `validate_tag()` function to a "private" function `_validate_tag()`.

aciba90:
@a-dubs, thanks for the improvements!

> I also moved the validate_tag() function to a "private" function _validate_tag()

I do not see that in the PR, maybe there is something left to push here?

pycloudlib/errors.py
```diff
@@ -138,3 +138,21 @@ class CleanupError(PycloudlibException):
 
 class MissingPrerequisiteError(PycloudlibException):
     """Raised when a prerequisite is missing."""
+
+
+class InvalidTagNameError(PycloudlibException):
+    """Raised when a tag for a cloud is invalid."""
+
+    def __init__(self, tag: str, rules_failed: list):
```
aciba90, line 146:
```suggestion
    def __init__(self, tag: str, rules_failed: List[str]):
```

pycloudlib/cloud.py
```diff
@@ -285,3 +285,39 @@ def _check_and_set_config(
             self.config = {}
         else:
             self.config = parse_config(config_file)[self._type]
+
+    @staticmethod
+    def _validate_tag(tag: str):
```
aciba90, line 290: Do we need to return the tag? Why not simply it and not returning?

If we insist in returning the tag, could we fully type it, please?

```suggestion
    def _validate_tag(tag: str) -> str:
```

And the same in child methods.

aciba90 (changes requested):
Thansk, @a-dubs, for adding a specific exception. I left some in-line comments.

In addition to that, as we are breaking the API and increasing the major version, could we please add a brief explanation in the commit message, in the main PR comment, about what we are breaking? Something like:

```
feat!: add cloud specific tag validation 

Revamped base class tag validation function and added cloud specific tag validation functions for IBM VPC and IBM Classic.

BREAKING CHANGE: Remove pycloudlib.util.validate_tag

Fixes GH-396
```

pycloudlib/cloud.py
```diff
@@ -72,9 +72,9 @@ def __init__(
             name=self.config.get("key_name", user),
         )
         if timestamp_suffix:
-            self.tag = validate_tag(get_timestamped_tag(tag))
+            self.tag = self._validate_tag(get_timestamped_tag(tag))
```
aciba90, line 75: After the last modifications, we need to explicitly assign the tag here:

```diff
diff --git a/pycloudlib/cloud.py b/pycloudlib/cloud.py
index e1b7941..375f314 100644
--- a/pycloudlib/cloud.py
+++ b/pycloudlib/cloud.py
@@ -72,9 +72,9 @@ class BaseCloud(ABC):
             name=self.config.get("key_name", user),
         )
         if timestamp_suffix:
-            self.tag = self._validate_tag(get_timestamped_tag(tag))
-        else:
-            self.tag = self._validate_tag(tag)
+            tag = get_timestamped_tag(tag)
+        self._validate_tag(tag)
+        self.tag = tag
 
     def __enter__(self):
         """Enter context manager for this class."""
```

aciba90 (changes requested):
Just a minor in-line fix and ship it!

pycloudlib/cloud.py
a-dubs: oh wow great catch! thank you.

aciba90 (approved):
LGTM, but there are some [formatting issues](https://github.com/canonical/pycloudlib/actions/runs/10492050476/job/29062630747?pr=398), could you please fix them?

PR #399, opened 2024-08-14
ci: fix ruff autofixable issues in tox -e format command

a-dubs:
This will handle fixing all auto fixable ruff issues when running `tox -e format`

## Testing Changes

1. choose a random file in the `pycloudlib/` dir and move one of the imports around so that it is out of order
2. run `tox -e format` and that change should be automatically fixed


holmanb (approved):
Sure, let's do it!

a-dubs:
LETS GOO

PR #400, opened 2024-08-15
feat!(ibm): fix race condition for multiple vms trying to attach same existing floating ip

a-dubs:
Before this change, when trying to launch several instances simultaneously using existing floating IPs, the instances would all choose the same floating IP and "successfully" attach it, but in reality, only one instance would actually have the IP attached. This leaves the rest stuck without a public IP address and are therefore inaccessible. This change makes it so that an instance actually confirms that it has attached the target floating IP after it tries attaching. And to reduce contention over floating IPs, instead of choosing the first floating IP among the available existing ones, choose randomly from the set of options.

BREAKING CHANGE:
- pycloudlib.ibm.cloud.IBM.launch: arg renamed from
  use_existing_floating_ip_with_name to floating_ip_substring 

- Rename IBMInstance.with_floating_ip to
  IBMInstance.from_raw_instance and drop floating_ip arg. This
  classmethod creates an instance and automatically handles 
  floating_ip attachment.
  
---

## Testing of changes
- Launch two instances simultaneously in same region and zone until they both attempt to use the same floating IP address and if they do, ensure that one proceeds normally and the other one realizes it did not attach the IP successfully and thus retries and is then able to find and attach a floating IP successfully.


pycloudlib/ibm/cloud.py
```diff
@@ -381,16 +333,28 @@ def launch(
...
             username=username,
             using_existing_floating_ip=floating_ip_substring is not None,
         )
+
+        # append before attaching floating ip so we can clean up in case of failure while attaching floating ip
         self.created_instances.append(instance)
 
+        # floating ip stuff
+        if floating_ip_substring:
+            instance.attach_floating_ip_until_success(
```
aciba90, line 349: What if we have only one free ip but two instances get to this point? One of them will never success here, right?

What about something like the following?

```python
        need_floating_ip = not floating_ip_substring
        if floating_ip_substring:
            try:
                instance.attach_floating_ip_until_success(
                    floating_ip_name_includes=floating_ip_substring,
                    zone=self.zone,
                )
            except Exception as e:
                need_floating_ip = True
        if need_floating_ip:
            self._log.info("Creating new floating ip.")
            floating_ip = self._create_floating_ip(name=floating_ip_name)
            instance.attach_floating_ip(floating_ip=floating_ip)

```

pycloudlib/ibm/instance.py
```diff
@@ -479,35 +481,132 @@ def __init__(
...
             client=client,
             instance=instance,
-            floating_ip=floating_ip,
             username=username,
             **kwargs,
         )
 
+    def _choose_from_existing_floating_ips(
+        self,
+        zone: str,
```
aciba90, line 512: Could you please add doc about `zone`?

pycloudlib/ibm/instance.py
```diff
@@ -479,35 +481,132 @@ def __init__(
...
+        floating_ip = random.choice(floating_ips)
+
+        self._log.info(
+            "Using existing floating ip '%s' with address %s",
+            floating_ip["name"],
+            floating_ip["address"],
+        )
+        return floating_ip
+
+    def attach_floating_ip(self, floating_ip: dict) -> Optional[dict]:
```
aciba90, line 560: Is this method intended to be part of the public API? Or should we make it private?

pycloudlib/ibm/instance.py
```diff
@@ -479,35 +481,132 @@ def __init__(
...
+            self._client,
+            id=floating_ip["id"],
+            instance_id=self._instance["id"],
+            network_interface_id=nic_id,
+        ).get_result()
+        # sleep 1s to let cloud update then check to make sure that floating IP successfully attached
+        time.sleep(1)
+        return self._discover_floating_ip(self._client, self._instance)
+
+    def attach_floating_ip_until_success(
```
aciba90, line 581: Is this method intended to be part of the public API? Or should we make it private?

pycloudlib/ibm/instance.py
```diff
@@ -479,35 +481,132 @@ def __init__(
...
+        nic_id = self._instance["primary_network_interface"]["id"]
+        self._ibm_instance_type.add_instance_network_interface_floating_ip(
+            self._client,
+            id=floating_ip["id"],
+            instance_id=self._instance["id"],
+            network_interface_id=nic_id,
+        ).get_result()
+        # sleep 1s to let cloud update then check to make sure that floating IP successfully attached
+        time.sleep(1)
+        return self._discover_floating_ip(self._client, self._instance)
```
aciba90, line 579: The dict info is not used, can we simplify it and return a boolean representing success?

pycloudlib/ibm/instance.py
```diff
@@ -479,35 +481,132 @@ def __init__(
...
+
+        Args:
+            floating_ip_name_includes: substring to filter floating IPs by
+            zone: zone to filter floating IPs by
+
+        Returns:
+            A floating IP dict
+        """
+        attached_floating_ip = None
+        while attached_floating_ip is None:
```
aciba90, line 595: * This loop is a potential infinite loop. Could we top-limit with a fixed number of retries?
* This loop does not give feedback to the user, could we add some logs from within the loop?

pycloudlib/ibm/instance.py
```diff
@@ -479,35 +481,132 @@ def __init__(
...
+            target_floating_ip = self._choose_from_existing_floating_ips(
```
aciba90, line 596: tl;dr: ignore this comment.

This method internally chooses randomly from the existent free IPs. If there are the same number of ips as vms spwaning, we can think about 2 ips and 2 vms, they could in theory indefinitely fight for the free ips.

While this is possible in theory, I think it is statistically difficult to get to that case and a solution would need pycloudlib to store state about which ips it's trying to grab. And even then, the solution wouldn't be complete, as multiple pycloudlib process could be acting upon the same ibm account and/or other external users could interact with the account.

aciba90 (changes requested):
Thanks, @a-dubs, for addressing this. The approach looks good to me. I left in-line comments / requests.

It looks that the commit summary is wrong, right? Instead of

`feat(ibm): fix race condition for multiple vms trying to attach same existing floating ip`

it should be

`fix(ibm): fix race condition for multiple vms trying to attach same existing floating ip`

pycloudlib/ibm/cloud.py
a-dubs: This is handled in the downstream function _choose_from_existing_floating_ips(). This function will raise a capacity error once there are no more available floating IPs. This is important because it allows for consumers of pycloudlib to do retries (like CTF). Doing it this way is also important because if the user specifies that an existing floating IP should be used, we do NOT want to spin up a new one for them since that would be unwanted behavior.

pycloudlib/ibm/instance.py
a-dubs: good catch, thanks!

pycloudlib/ibm/instance.py
a-dubs: so the nice thing is, that if two VMs try to attach the same floating IP (which will happen often), one of them will succeed and the rest will fail. They don't all fail. Which ensures that this will never loop forever. and then if they run out floating IPs, a capacity exception will be raised so that edge case is handled as well.

pycloudlib/ibm/instance.py
a-dubs: Oof yeah I need to go add more consistent logging throughout.

> This loop is a potential infinite loop. Could we top-limit with a fixed number of retries?

See following comment for this. wouldn't be opposed to adding a limit tho just in case.

pycloudlib/ibm/instance.py
a-dubs: hmmmm i kinda like the idea of returning the dict but you are right it is not used and doesn't really add much. will convert to returning a bool :saluting_face:

pycloudlib/ibm/instance.py
a-dubs: definitely want this private

pycloudlib/ibm/instance.py
a-dubs: hmmm... this does not provide much utility as a public api method so will make this private as well.

pycloudlib/ibm/instance.py
aciba90: Thanks for the explanation. Do you think it makes sense to add that comment on top of the while loop, for future reference?

aciba90 (changes requested):
Thanks for the improvements!

Unfortunately, `_attach_floating_ip_until_success` and `_attach_floating_ip` are used from `pycloudlib.ibm.cloud` and we cannot express in python that a method is public to the package itself but private to outside callers. Could we please make them public again?

pycloudlib/ibm/instance.py
```diff
@@ -477,37 +477,206 @@ def __init__(
...
+        """
+        Attach a floating IP to this instance.
+
+        If `floating_ip_substring` is given, it will try to attach an existing
+        floating IP that contains the substring in its name. If not, it will
+        create a new floating IP and attach it.
+
+        Args:
+            floating_ip_substring: optional substring to find existing floating IPs by.
+                If not given, ta new floating IP will be created and attached.
```
aciba90, line 593: Typo:

```suggestion
                If not given, a new floating IP will be created and attached.
```

pycloudlib/ibm/instance.py
```diff
@@ -477,37 +477,206 @@ def __init__(
...
+
+        Returns:
+            A floating IP dict if successful, None otherwise
```
aciba90, line 596: Do we really want to return the floating IP dict? That dict is a dict that directly comes from the ibm sdk library and:

1. It is unspecified from a pycloudlib perspective.
2. If we return it here, then we are coupling pycloudlib's api with the ibm-sdk one. And if they break, pycloudlib breaks.

If this dict is needed to be exposed for any reason, then I would probably suggest creating a proper class to present it in our API and translate the raw dict response to the class. WDYT?

aciba90 (changes requested):
This is looking great, thanks for the improvements. I left an in-line comment.

Apart from that, I think we need to bump the major version and modify the commit message to include a footer with something like:

```
BREAKING CHANGE:
- pycloudlib.ibm.cloud.IBM.launch: arg renamed from
  use_existing_floating_ip_with_name to floating_ip_substring 

- Rename IBMInstance.with_floating_ip to
  IBMInstance.from_raw_instance and drop floating_ip arg. This
  classmethod creates an instance and automatically handles 
  floating_ip attachment.
```

pycloudlib/ibm/instance.py
a-dubs: This is honestly a good a point. Thanks for bringing that up. I'll just change it to return the ID or the name or something.

pycloudlib/ibm/instance.py
a-dubs: added a max number of tries that this can do because i do still hate the idea of a forever while loop.

a-dubs:
bumped version. so version check will pass after https://github.com/canonical/pycloudlib/pull/398 is merged

aciba90 (changes requested):
Great work here @a-dubs. Can we please fix the [lint error](https://github.com/canonical/pycloudlib/actions/runs/10516307301/job/29138251553?pr=400) and we ship it!?:

> pycloudlib/ibm/instance.py:1:0: C0302: Too many lines in module (1001/1000) (too-many-lines)

aciba90 (approved):
LGTM, awesome work here, thanks!

PR #401, opened 2024-08-15
Oracle various improvements to config and ipv6

a-dubs:
When merging, this should be left as 3 separate logical commits

## Testing of changes:
- Tested passing profile, region, both profile and region and then did the same except specified via pycloudlib.toml and ensured that my default profile settings were overriden 
- Confirmed `get_image_id_from_name()` works by using it in a script.
- Tested ipv6 changes in that same script


pycloudlib.toml.template
```diff
@@ -71,6 +71,8 @@ domain_name = ""
 config_path = "~/.oci/config"
 availability_domain = ""  # Likely in ~/.oci/oci_cli_rc
 compartment_id = ""  # Likely in ~/.oci/oci_cli_rc
+# region = "us-phoenix-1"  # will use region from oci config file if not specified
```
aciba90, line 74: Do we think it is worth adding docs for this new configs to https://pycloudlib.readthedocs.io/en/latest/clouds/oci.html ?

aciba90 (approved):
LGTM, thanks!

Before merging, we need to update the minor version and I left one inline comment / request.

pycloudlib.toml.template
a-dubs: Normally, I would say yes the docs should be updated as well, but in this case, the oracle docs seem woefully out of date and far too barebones so I think this should be done in a separate PR. I will put up an issue for this and fix it myself sometime soon. S

a-dubs:
@aciba90  thanks for your review! version has been bumped!

pycloudlib.toml.template
aciba90: Thanks! Could you please link the issue here when done?

Issue #341, opened 2024-02-14
Use priority list for pubkey names


holmanb:
[Defaulting to `$HOME/.ssh/id_rsa.pub`](https://github.com/canonical/pycloudlib/blob/6e1ee98acd88cff495e115f0c3c9dce1ce717236/pycloudlib/cloud.py#L67) made more sense when that was the default key type generated by `ssh-keygen`, but in Noble that is not the case:

```bash
$ ssh-keygen 
Generating public/private ed25519 key pair.
Enter file in which to save the key (/home/holmanb/.ssh/id_ed25519):
```

For convenience it would be nice if we could check a priority list of key names for their existence rather than assuming a single file.

a-dubs:
Fixed by #406

a-dubs:
Fixed by #406

PR #406, opened 2024-08-26
feat: add ability to check for multiple default ssh key locations

a-dubs:
Since Ubuntu 24.04 LTS and newer, the default location for ssh keys is
`~/.ssh/id_ed25519` instead of `~/.ssh/id_rsa`. Thus, when no ssh key
path is specified, this commit changes pycloudlib to check for an
existing ssh key at either  `~/.ssh/id_ed25519` or `~/.ssh/id_rsa`.

Fixes GH #341

---

### Testing id_ed25519 can be found
1. made sure no key exists at `~/.ssh/id_rsa`
2. created new key at `~/.ssh/id_ed25519`
3. removed `public_key_path` and `private_key_path` from `[oci]` config in my pycloudlib.toml config file
4. ran the oracle example script: `python3 examples/oracle.py`
5. made sure the following line showed up in the log (was one of the first few lines logged):
```
INFO:pycloudlib.cloud.OCI:using SSH pubkey: /home/a-dubs/.ssh/id_ed25519.pub
```

### Testing error is raised when neither default key is found
1. Removed keys at `~/.ssh/id_ed25519`
2. ran the oracle example script: `python3 examples/oracle.py`
3. Ensured error stating the following was thrown:
```
pycloudlib.errors.PycloudlibError: No public key path provided and no key found in default locations: '~/.ssh/id_rsa.pub' or '~/.ssh/id_ed25519.pub'
```

### Testing error is raised when key from config does not exist:
1. set `public_key_path` in `[oci]` config in my pycloudlib.toml config file to be `~/.ssh/nonexistant-key.pub`  
2. ran the oracle example script: `python3 examples/oracle.py`
3. Ensured error stating the following was thrown:
```
pycloudlib.errors.PycloudlibError: Provided public key path '/home/a-dubs/.ssh/oracle-ipv6-testing.pub1' does not exist
```


a-dubs:
putting this into draft until i fix tests.

a-dubs:
Also, should i add new unit tests that verify the functionality of the added changes? @holmanb (tag you're it)

pycloudlib/cloud.py
```diff
@@ -62,13 +66,39 @@ def __init__(
         self._check_and_set_config(config_file, required_values)
 
         user = getpass.getuser()
+        # check if id_rsa or id_ed25519 keys exist in the user's .ssh directory
```
holmanb, line 69: This turns `__init__()` mostly into a key-handling function. I would prefer a helper method `_get_ssh_keys()`[1] to contain this logic and get called from `__init__()`

[1] maybe a better name, I didn't put any thought into that one

holmanb (commented):
> Also, should i add new unit tests that verify the functionality of the added changes? @holmanb (tag you're it)

Pycloudlib's test coverage is unfortunately very light on tests, but in this case the feature being tested is something that a user would immediately notice if it isn't working.

I'm honestly not sure whether it would be worth the effort in this case to write and maintain a test given the infrequency of changes to this part of the code and the immediate feedback that we would receive.

What do you think?

pycloudlib/cloud.py
a-dubs: Haha agreed! Way too much going on there. And would allow for actually unit testing that (if so desired)

a-dubs:
> > Also, should i add new unit tests that verify the functionality of the added changes? @holmanb (tag you're it)
> 
> Pycloudlib's test coverage is unfortunately very light on tests, but in this case the feature being tested is something that a user would immediately notice if it isn't working.
> 
> I'm honestly not sure whether it would be worth the effort in this case to write and maintain a test given the infrequency of changes to this part of the code and the immediate feedback that we would receive.
> 
> What do you think?

Let me go fix the existing unit tests that this change broke, and see if this change gets inherently tested that way. If not, it should be sufficient to just test the helper function that I'll create encapsulating this functionality.

a-dubs:
@holmanb ready for re-review!

tests/unit_tests/openstack/test_cloud.py
```diff
@@ -15,7 +15,7 @@
 
 """
 
-
```
holmanb, line 18: Whitespace

holmanb (approved):
@a-dubs Nit: I see some unnecessary whitespace changes in diff, but other than that everything else looks good. Please clean up the whitespace changes, then feel free to merge.

a-dubs:
> @a-dubs Nit: I see some unnecessary whitespace changes in diff, but other than that everything else looks good. Please clean up the whitespace changes, then feel free to merge.

@holmanb all those double whitespaces are because they are between classes not functions and this is the formatting enforced by ruff. to verify, I removed the double whitespace then ran `tox -e format` and it added the second newline back in.

PR #407, opened 2024-08-28
feat(oracle): add ability to retain snapshot after cleanup

a-dubs:
This allows for marking snapshots to be kept at time of creation
instead of being cleaned up automatically. This allows for still using
context managers for easy cleanup while retaining snapshots where
desired, and also when not using context managers and the 
cloud.clean() method is called.

---

Keeping this as a draft until the following is resolved.
Question for the maintainers: Do we want to make this an optional part of the standard pycloudlib API across clouds? I could easily add this to all other clouds if so.


pycloudlib/oci/cloud.py
```diff
@@ -366,7 +366,11 @@ def snapshot(self, instance, clean=True, name=None):
...
             desired_state="AVAILABLE",
         )
-
-        self.created_images.append(image_data.id)
+        
+        # only add image to created_images list for cleanup if keep is False
+        if not keep:
+            self.created_images.append(image_data.id)
+        else:
+            self._log.info("Keeping image %s", image_data.display_name)
```
blackboxsw, line 374: I think it's worth us persisting a list of `self.kept_images` or `self.retained_mages`. Then on teardown/`cloud.clean()`  we can make sure we emit a footer log or something conspicuous that says `f"Skipping deletion of preserved  image(s): {self.retained_images}"`. Otherwise there is little hope we will see this inline log during image creation during our busy integration tests.

Also we probably want to log both name and id so that we can reference that image in the future on CLI/SDK/cloud-dashboards.
```suggestion
            self._log.info("Keeping image %s[id:%s]", image_data.display_name, image_data.id)
```

blackboxsw (commented):
> Question for the maintainers: Do we want to make this an optional part of the standard pycloudlib API across clouds? I could easily add this to all other clouds if so.

I agree with your sentiment. I think it's a valuable feature to have generally across clouds. Let's make this something available in the general API of `pycloudlib.Cloud.snapshot`.

pycloudlib/oci/cloud.py
a-dubs: Agreed! Thanks for the feedback @blackboxsw!

pycloudlib/cloud.py
```diff
@@ -224,6 +225,13 @@ def clean(self) -> List[Exception]:
                 self.delete_image(image_id)
             except Exception as e:
                 exceptions.append(e)
+        for image_info in self.preserved_images:
+            # send info log informing that the image is NOT being cleaned up
+            self._log.info(
+                "Preserved image %s [id:%s] is NOT being cleaned up.",
```
blackboxsw, line 231: LGTM. should be a conspicuous enough log message at the end of tests. thanks.
We'll need to update the `pycluldlib.Cloud.snapshot` signature and docstr to expect an optional `keep: bool = False` param too and reflect that functionality to each `pycloudlib.<could_name>.cloud.snapshot` if we could.

pycloudlib/oci/cloud.py
```diff
@@ -366,7 +366,15 @@ def snapshot(self, instance, clean=True, name=None):
...
-        self.created_images.append(image_data.id)
+        
+        # only add image to created_images list for cleanup if keep is False
+        if not keep:
+            self.created_images.append(image_data.id)
+        else:
+            self.preserved_images.append({
+                "id": image_data.id,
+                "name": image_data.display_name,
+            })
```
blackboxsw, line 377: If we want to get exotic we could use a namedtuple for this instead of a dict keyed on id and name.

pycloudlib/cloud.py
a-dubs: Sounds good! Should I throw this back into draft state and make the changes across all clouds?

blackboxsw: no need to draft state this, I'll just await a request for re-review from you when you are ready on this.

pycloudlib/cloud.py
```diff
@@ -34,6 +35,13 @@ class ImageType(enum.Enum):
     PRO_FIPS = "Pro FIPS"
 
 
+@dataclasses.dataclass
+class ImageInfo:
```
blackboxsw, line 39: Even fancier than NamedTuple. +1 py 3.7 support where dataclass was introduced is good enough for me. Makes the declaration of this  datatype easier to read/consume.

blackboxsw: Note if pycloudlib requires python 3.7 ++ we need to update our project docs which state python 3.6 https://pypi.org/project/pycloudlib/

blackboxsw: Since this could represent a breaking change due to py 3.7 features, we should bump the major version unless we can find py 3.7 features that are already in pycloudlib.

pycloudlib/azure/cloud.py
```diff
@@ -1229,7 +1232,10 @@ def snapshot(
         image_id = image.id
         image_name = image.name
 
-        self.created_images.append(image_id)
+        self._store_snapshot(
```
blackboxsw, line 1235: typo

a-dubs:
@blackboxsw  finally ready for re-review!

a-dubs:
closed in favor of #472 which cherry-picks this change and wraps it into one big v11 bundle

a-dubs:
closed in favor of #472 which cherry-picks this change and wraps it into one big v11 bundle

PR #409, opened 2024-09-06
refactor(gce): migrate discovery api to compute_v1

mstpn:
When pycloudlib was created, the compute engine library for python (compute_v1) did not exist. As CPC is looking to make more contributions to the GCE module of pycloudlib as part of an internal tooling refactor, it makes sense to move to the more modern library that has support for type hinting and a more modern development interface. The new library wraps the same underlying REST API, so there is no regression in features.


pycloudlib/gce/cloud.py
```diff
@@ -104,13 +106,14 @@ def __init__(
...
-            "v1",
-            cache_discovery=False,
-            credentials=credentials,
+        self.images = compute_v1.ImagesClient(credentials=credentials)
+        self.disks = compute_v1.DisksClient(credentials=credentials)
+        self.instances = compute_v1.InstancesClient(credentials=credentials)
+        self.zone_operations = compute_v1.ZoneOperationsClient(
+            credentials=credentials
+        )
+        self.global_operations = compute_v1.GlobalOperationsClient(
```
a-dubs, line 115: for clarity, could we rename these to all end with "_client" ?

a-dubs: also nit/question: would it make sense to rename these to start with a "_" to denote them as "private"?
For the two ibm clouds, i kept the "client" variables as private but many other clouds don't do this so this is by no means a well defined standard or rule.

setup.cfg
```diff
@@ -30,7 +30,8 @@ install_requires =
     azure-mgmt-resource >= 15
     boto3 >= 1.14.20
     botocore >= 1.17.20
-    google-api-python-client >= 1.7.7
+    google-cloud-compute <= 1.9.0
+    googleapis-common-protos == 1.63.1
```
a-dubs, line 34: given that you had issues with dependency resolution, could you add a comment to each of these entries explaining why they are pinned the way they are?

a-dubs (changes requested):
Thank you for these changes @mstpn! Amazing work fr.

Gave this a readthrough and left a coupe inline comments.

Will definitely want some other eyes/reviews on this PR given how sweeping the changes are.

setup.cfg
mstpn: Good call. I actually should test without pinning `google-cloud-compute` now that I found the issue lied with the `googleapis-common-protos` dependency.

pycloudlib/gce/cloud.py
mstpn: I can rename to append "_client" for all the applicable class attributes.

Can't see an issue with making them private as well, so this refactor is a good time to do that as well.

mstpn:
Addressed @a-dubs comments above and resolved conflicts with [417](https://github.com/canonical/pycloudlib/pull/417). PR is ready for review.

a-dubs (approved):
Change look good. Would like to see more unit tests added in the future, but you updated all existing ones which is sufficient for now!

Thanks for tackling this large issue @mstpn :partying_face:

PR #410, opened 2024-09-09
feat(lxd): support launching minimal images on lxd_vm and container (SC-1750)

blackboxsw:
Add optional image_type param to LXD*.released_image and daily_image methods to provide daily and released Ubuntu minimal images.

To support this feature, lxd._images.find_last_fingerprint also accepts an image_type param to distinguish between ubuntu-minimal and generic ubuntu image remotes.

Support for jenkins integration test runners of Ubuntu daily minimal images. Component of improved integration test coverage for SC-1750.

### test procedure
```
tox -e pytest
tox -e integration-tests -- tests/integration_tests/test_public_api.py::test_public_api[lxd_container]
.tox/.testenv/bin/python3 example/lxd.py
```


tests/integration_tests/test_public_api.py
```diff
@@ -137,3 +137,22 @@ def test_public_api(cloud: BaseCloud):
...
+            assert (
+                "status: done" == instance.execute("cloud-init status").stdout
+            )
+            assert (
+                "minimal"
+                in instance.execute(
+                    "grep build_name /etc/cloud/build.info"
+                ).stdout
+            )
+            instance.delete()
```
a-dubs, line 158: should this test be added to an lxd integration test instead? given that this will only run on lxd and the public api is unaffected by this change?

a-dubs (changes requested):
thanks for this change chad! implementation looks solid but one question about the integration test addition inline.

tests/integration_tests/test_public_api.py
blackboxsw: Good suggestion, I pulled it into a separate test and also parametrized that test to cover GCE and Ec2 which also support ImageType.MINIMAL image_type overrides in released_image/daily_image

tests/integration_tests/test_public_api.py
```diff
@@ -137,3 +137,37 @@ def test_public_api(cloud: BaseCloud):
...
+
+
+@pytest.mark.parametrize(
+    "cloud",
+    [
+        pytest.param(pycloudlib.EC2, id="ec2", marks=pytest.mark.main_check),
+        pytest.param(pycloudlib.GCE, id="gce", marks=pytest.mark.main_check),
+        pytest.param(
+            pycloudlib.LXDContainer, id="lxd_container", marks=pytest.mark.ci
+        ),
```
a-dubs, line 153: should LXD VM also be tested here? or is that unnecessary?

a-dubs (approved):
Changes look great! Thanks @blackboxsw. Left one inline question but it's minor so I'm not going to block on it.

PR #412, opened 2024-09-10
feat(ec2): Allow using ec2 profile as creds

TheRealFalcon:
I have a feeling I'm just missing something obvious, but with my SSO creds I can no longer get a secret key, but using a profile is an alternative that should work too.


a-dubs (approved):

aciba90:
> I have a feeling I'm just missing something obvious, but with my SSO creds I can no longer get a secret key, but using a profile is an alternative that should work too.

I think the idea behind SSO access is:

SSO let you access your root AWS account. From the root account, one can create a non-root user. And those non-root user's credentials can be used as previously in pycloudlib.

Does that work for you?

TheRealFalcon:
> Does that work for you?

Makes sense, but I was actually able to find the credentials. They're just listed on a different screen at login time rather than the original account page.

I think this functionality is still useful though as it obviates the need to manage secret keys.

pycloudlib.toml.template
```diff
@@ -24,10 +24,15 @@ tenant_id = ""
...
 
 [ec2]
-# Most values can be found in ~/.aws/credentials or ~/.aws/config
-access_key_id = ""  # in ~/.aws/credentials
-secret_access_key = ""  # in ~/.aws/credentials
 region = ""  # in ~/.aws/config
+# If 'aws configure sso' has been run, 'profile' should be the only credentials needed
+profile = ""  # in ~/.aws/config
+
+# With modern SSO, these should no longer be necessary.
```
aciba90, line 31: Nit: `With modern SSO` sounds a bit pretentious and likely prone to age bad. Something like the following?

```suggestion
# If profile is given, these are not necessary.
```

aciba90 (changes requested):
Changeset looks good to me, thanks.

I think we should extend https://github.com/canonical/pycloudlib/blob/main/docs/clouds/ec2.md to clarify on how to use the `profile` mode. WDYT?

aciba90 (approved):
LGTM

PR #413, opened 2024-09-10
feat: Add cloud version string

holmanb:


tests/integration_tests/test_public_api.py
```diff
@@ -112,6 +112,8 @@ def test_public_api(cloud: BaseCloud):
         # Not sure there's a great way to test this other than not raising
         cloud.image_serial(image_id)
 
+    version = cloud.version()
+    assert  isinstance(version, str) or version is None
```
a-dubs, line 116:
```suggestion
    assert isinstance(version, str) or version is None
```

a-dubs (changes requested):
could you add a super quick unit test to each lxd and qemu cloud unit test files?

pycloudlib/cloud.py
```diff
@@ -260,6 +260,10 @@ def use_key(self, public_key_path, private_key_path=None, name=None):
         self._log.debug("using SSH key from %s", public_key_path)
         self.key_pair = KeyPair(public_key_path, private_key_path, name)
 
+    def version(self) -> Optional[str]:
+        """Version string of the platform."""
+        return None
```
blackboxsw, line 265: Is there a lot of value in us returning None here instead of just an empty string ""? We can then avoid the Optional[str] handling for typing validation.

pycloudlib/qemu/cloud.py
```diff
@@ -660,3 +660,12 @@ def clean(self) -> List[Exception]:
             exceptions.append(e)
 
         return exceptions
+
+    def version(self):
```
blackboxsw, line 664: @holmanb We should probably lru_cache this as it's unlikely to change mid integration test run, but it's something we don't want to recalculate any time we call it.

pycloudlib/lxd/cloud.py
```diff
@@ -437,6 +437,10 @@ def clean(self) -> List[Exception]:
                     exceptions.append(e)
         return exceptions
 
+    def version(self):
```
blackboxsw, line 440: Same, I'd cache this value somehow as it's unlikely to change on the host mid testing.

PR #420, opened 2024-09-18
fix(ibm): actually fix attaching of floating ips

a-dubs:
A boolean was being checked against None incorrectly so the falsey value was not respected. This fixes that.

--- 

Testing of changes:
Ran `python examples/ibm.py` locally and confirmed exception was no longer raised when attaching floating ip.


blackboxsw (commented):
At risk of being a blocker can we get a unit test for this? I see no coverage at all for IBM and floating_ip. I'll spend about 20 mins to see if I can't cobble one up. Let's strive for adding skeletal test coverage where we can and built it up over time.

a-dubs:
> At risk of being a blocker can we get a unit test for this? I see no coverage at all for IBM and floating_ip. I'll spend about 20 mins to see if I can't cobble one up. Let's strive for adding skeletal test coverage where we can and built it up over time.

Honestly, good point. Given that this is my third time trying to get this working and second consecutive bug fix, I think it would be more than wise to add unit testing and integration testing for this. 

I'll see what I can throw together unit testing wise, but it will definitely be easy to get an integration test for this and that should ensure that this functionality is working as expected. 

Thanks for standing on business and raising this 🫡

blackboxsw (commented):
@a-dubs well this was a pain due to side-effects inside existing IBMInstance pushing me to mock a bunch. I've pushed a start to unittests there which admittedly have a whole lot of mocking. It's a start FWIW. But I hit time cap for this initial investment in coverage.

```diff
diff --git a/tests/unit_tests/ibm/test_instance.py b/tests/unit_tests/ibm/test_instance.py
new file mode 100644
index 0000000..65dbbee
--- /dev/null
+++ b/tests/unit_tests/ibm/test_instance.py
@@ -0,0 +1,81 @@
+"""Module for IBM instance tests."""
+
+import pytest
+from unittest import mock
+
+from pycloudlib.ibm.instance import IBMInstance, _IBMInstanceType
+
+SAMPLE_RAW_INSTANCE = {
+    "id": "ibm1",
+    "primary_network_interface": {"id": "nic1"},
+    "profile": {"name": "metal"},
+    "zone": {"name": "zone1"},
+}
+M_PATH = "pycloudlib.ibm.instance."
+
+
+class TestIBMInstance:
+    @pytest.mark.parametrize(
+        "raw_instance,inst_id,zone_id,inst_type",
+        (
+            (
+                SAMPLE_RAW_INSTANCE,
+                "ibm1",
+                "zone1",
+                _IBMInstanceType.BARE_METAL_SERVER,
+            ),
+        ),
+    )
+    @mock.patch(M_PATH + "VpcV1", autospec=True)
+    def test_type_from_raw_instance(
+        self, client, raw_instance, inst_id, zone_id, inst_type
+    ):
+        """Factory function inits the appropriate IBMInstanceType."""
+        inst = IBMInstance.from_raw_instance(
+            key_pair=None,
+            client=client,
+            instance=raw_instance,
+        )
+        assert client == inst._client
+        assert inst_id == inst.id
+        assert zone_id == inst.zone
+        assert inst_type == inst._ibm_instance_type
+
+    @mock.patch(M_PATH + "VpcV1", autospec=True)
+    def test_attach_floating_ip(self, client, caplog):
+        """
+        retry attach_floating_ip until VpcV1 lists non-empty floating_ips."""
+
+        inst = IBMInstance.from_raw_instance(
+            key_pair=None,
+            client=client,
+            instance=SAMPLE_RAW_INSTANCE,
+        )
+        metal_floating_ips = mock.Mock(
+            get_result=mock.Mock(
+                side_effect=[
+                    {"floating_ips": []},
+                    {
+                        "floating_ips": [
+                            {"id": "floatingid1", "name": "floatingname"}
+                        ]
+                    },
+                ]
+            )
+        )
+        with mock.patch.object(
+            client,
+            "list_bare_metal_server_network_interface_floating_ips",
+            return_value=metal_floating_ips,
+        ), mock.patch.object(
+            inst,
+            "_choose_from_existing_floating_ips",
+            return_value={"name": "ci-ip-match-5", "id": "floatingid1"},
+        ):
+            inst.attach_floating_ip(floating_ip_substring="ci-ip-match")
+        for expected_log in [
+            "Failed to attach floating ip: ci-ip-match-5.",
+            "Successfully attached floating ip: floatingname",
+        ]:
+            assert expected_log in caplog.text
+        assert 2 == metal_floating_ips.get_result.call_count
```

a-dubs:
@blackboxsw added integration test that validates floating IPs can actually be attached :sweat_smile:. Ready for re-review (:

tests/integration_tests/ibm/test_launch.py
```diff
@@ -0,0 +1,68 @@
...
+
+    This tests the following:
+    - The instance is launched successfully
+    - The instance is reachable via SSH
+    - The instance has a floating IP as expected
+    - The instance name was set correctly
+    """
+    time_id = time.time_ns()
+    image_id = ibm_cloud.released_image("noble")
+    floating_ip_substring = "default-floating-ip"
```
blackboxsw, line 52: Hard-coding this string in integration tests won't work in most environments unless an available prerequisite floating-ip has already been allocated for a group named with the default-floating-ip prefix (and not already used by some other CI runner like jenkins). 

This test should either, setup it's own unique floating IP with a custom prefix, or source pycloudlib.toml config.

Let's go the pycloudlib.toml config front here for the moment and to proposed diff
```diff
--- a/tests/integration_tests/ibm/test_launch.py
+++ b/tests/integration_tests/ibm/test_launch.py
@@ -49,7 +49,9 @@ def test_ibm_launch(ibm_cloud: IBM):
     """
     time_id = time.time_ns()
     image_id = ibm_cloud.released_image("noble")
-    floating_ip_substring = "default-floating-ip"
+    floating_ip_substring = ibm_cloud.config.get("floating_ip_substring", "")
+    if not floating_ip_substring:
+        floating_ip_substring = "default-floating-ip"
     unique_instance_name = f"integration-test-launch-instance-{time_id}"
     # create ssh_keys for use
     manage_ssh_key(ibm_cloud, f"integration-test-key-{time_id}")
```
```

Also, i'm not a big fan of our generic IBMException for any IBM error (not due to this PR). We should specialize errors so we know what to report to end-users.  In server team case, we happen to name our floating ip's differently than the `"default-floating-ip`, we I get the following traceback from integration tests:
```
        if not floating_ips:
>           raise IBMException(
                f"No floating IPs found matching substring {name_includes}"
            )
E           pycloudlib.ibm.errors.IBMException: No floating IPs found matching substring default-floating-ip
```

For a separate PR: what I'd rather see is a raised exception that is a subclass of pycloudlib.errors.ResourceNotFoundError  (maybe named `FloatingIPNotFoundError`). Then I can have a try/except FloatingIPNotFoundError around the launch to better inform the integration-test runner that they haven't pre-allocated the expected availble floating-ips in order to run the test


That comment about subclassing ResourceNotFoundErrors in IBM applies to many of the errors generated by pycloudlib.ibm.instance which can probably be taken to a separate issue/PR
 - "Subnet not found" 
 - No subnet associated to vpc found
 - VPC not found
 - No subnet available

blackboxsw (changes requested):
Avoid hard-coding and source pycloudlib.toml -> ibm_cloud.config["floating_ip_substring"] when set.

tests/integration_tests/ibm/test_launch.py
a-dubs: @blackboxsw 100% agree with the generic exception feedback. Will create an issue for this. Has been on the back of my mind for a while.

a-dubs:
@blackboxsw changes are up. If you're opposed to the new `_floating_ip_substring` variable, i am happy to just use `config.get(...)` instead!

blackboxsw (approved):
LGTM! ship it

PR #421, opened 2024-09-19
build: unrestrict protobuf deps

a-dubs:
Since pycloudlib is consumed by various other projects and is often
not standalone, therefore, we should not unnecessarily restrict
dependencies. This dependency was added recently in PR #409.


a-dubs:
@mstpn FYI since this relates to your GCE refactor.

mstpn:
@a-dubs Thanks for the effort in relaxing dependencies where it is possible.

aciba90 (approved):
LGTM, thanks @a-dubs.

Issue #72, opened 2020-12-09
Consistent approach to 'wait' methods


TheRealFalcon:
We have a number of wait methods including
wait, wait_for_execute, wait_for_cloudinit, wait_for_delete, wait_for_stop, ...

Most of these wait for some arbitrary amount of time and retry and different wait intervals. We should refactor all of these methods to take an optional timeout argument with a consistent default and a consistent retry interval.

Personally, I'd consider [the retry library](https://pypi.org/project/retry/) or an API that looks like it.

aciba90:
Linking duplicated: https://github.com/canonical/pycloudlib/issues/229

PR #426, opened 2024-09-24
feat: adding retry for Azure VM creation

r3tro7:
adding logic so that Quota related errors can be retried on Vm creation. Adding this to pycloudlib helps facilitate clearing of dangling resources on vm create retry.


a-dubs:
Hello @r3tro7. Thanks for contributing these changes! I noticed you are using the `backoff` library but this is currently not a requirement/dependency for pycloudlib so it will need added to `setup.cfg`.

aciba90:
Before merging this, we need to evaluate the new dependency on https://github.com/litl/backoff in terms of:

- How well maintained it is
- How many other dependencies it pulls
- Worth in general? If we use one function / feature, maybe it is worth implementing a simple retry decorator from our side

This is related to https://github.com/canonical/pycloudlib/issues/72.

danpdraper:
@r3tro7 @a-dubs @aciba90 Implementing the exponential back-off logic is straightforward. I do not see a reason to pull in the `backoff` dependency.

a-dubs:
@danpdraper appreciate the input. I'll throw up a solution for this and unit tests to insure it functions as desired.

aciba90 (changes requested):
As the consensus is that `backoff` is not needed as a new requirement to solve this issue, I am rejecting this PR for the moment.

a-dubs:
Created new PR that adds new custom "backoff" functionality and incorporates the Azure changes too in #435

Issue #428, opened 2024-09-24
Fix Failing CI Integration Tests


a-dubs:
3 tests are failing due to #409 and #410. These should be resolved ASAP because 2/3 failures seem to be symptoms of a regression being introduced. The other one just seems like an issue with how an integration tests is being setup / reusing a fixture.

```
FAILED tests/integration_tests/test_public_api.py::test_public_api_minimal_images*** - ValueError: Protocol message Instance has no "serviceAccounts" field.
FAILED tests/integration_tests/test_public_api.py::test_public_api*** - ValueError: Protocol message Instance has no "serviceAccounts" field.
ERROR tests/integration_tests/test_public_api.py::test_public_api_minimal_images*** - botocore.exceptions.ClientError: An error occurred (InvalidKeyPair.Duplicate) when calling the ImportKeyPair operation: The keypair 'pycl-test-0924-111426' already exists.
```

The first two relate to #409 and the last one relates to #410.

Issue #431, opened 2024-09-25
feature: ec2 should provide a delete_keys operation


blackboxsw:
A delete_keys operation would be useful in cloud teardown when upload_keys is performed by intrgration tests in
tests/integration_tests/test_public_api.py.

a-dubs:
after further investigation, i think i can close this ticket because:
A) ec2 already provides a delete_key() function 
B) it is already automatically called in the cleanup function

The reason the key was not being cleaned up in the error we see in #428 is because the key is created at the cloud level and the cloud gets re-used between tests. So once tests have completed, then the key gets auto cleaned up successfully.

blackboxsw:
Thank you for vigilance here. I quick fired the issue an overlooked the call.

PR #432, opened 2024-09-25
fix: avoid duplicated upload_key in ec2 integration tests

blackboxsw:
Fixes GH-428


blackboxsw:
added a followup feature request #431 so we don't leave stale/unused test keys in ec2 after integration tests

PR #435, opened 2024-10-01
feat: add new internal exponential "backoff" decorator util

a-dubs:
By implementing this as internal util, we avoid bringing in another dependency to add such simple functionality. Thorough unit tests were added to ensure this backoff decorator functions as desired.


r3tro7:
Thank you for the great work on this @a-dubs ! 
I've reviewed the changes and everything looks good and seems to work as intended.

pycloudlib/utils/__init__.py
```diff
@@ -0,0 +1 @@
+"""Utils module for pycloudlib."""
```
aciba90, line 1:
```suggestion
# This file is part of pycloudlib. See LICENSE file for license information.
"""Utils module for pycloudlib."""
```

pycloudlib/utils/backoff.py
```diff
@@ -0,0 +1,70 @@
...
+
+import random
+import time
+from functools import wraps
+from typing import Tuple, Type
+
+from pycloudlib.errors import PycloudlibTimeoutError
+
+
+def exponential_backoff(
```
aciba90, line 12: Nit suggestion:

```suggestion
def exponential_backoff(
    *,
```

This, forces client code to always give args as kwargs and, IMO, improves readability in call-sites.

pycloudlib/utils/backoff.py
```diff
@@ -0,0 +1,70 @@
...
+    retries=5,
```
aciba90, line 13: Could you please type the rest of params?

tests/unit_tests/test_backoff.py
```diff
@@ -0,0 +1,319 @@
+import pytest
+from unittest.mock import patch
+from pycloudlib.utils.backoff import exponential_backoff
+import time
+
+
+# A helper function to track call counts for testing
```
aciba90, line 7: Could please add this comment as the class docstr?

```
class CallTracket:
    """A helper class to track call counts for testing"""
```

tests/unit_tests/test_backoff.py
```diff
@@ -0,0 +1,319 @@
...
+    def __call__(self, *args, **kwargs):
+        self.call_count += 1
+        if self.fail_times > 0:
+            self.fail_times -= 1
+            if self.exception:
+                raise self.exception
+        return "Success"
+
+
+# Class for simpler backoff tests
```
aciba90, line 23: Add this as the class / function docstr.

tests/unit_tests/test_backoff.py
```diff
@@ -0,0 +1,319 @@
...
+        Tests that the function retries on any exception when no specific
+        exception list is provided. It ensures that the exponential backoff
+        retries the correct number of times and succeeds if the function
+        eventually stops raising exceptions.
+
+        Optimization: time.sleep is mocked to avoid actual delays.
+        """
+        tracker = CallTracker(exception=Exception, fail_times=2)
+
+        @exponential_backoff(retries=3, base_delay=1)
```
aciba90, line 37: This is a bit superfluous as we could directly call decorators with:

```
exponential_backoff(retries=3, base_delay=1)(tracker)()
```

tests/unit_tests/test_backoff.py
aciba90, line 1: Nice touch on these tests.

pycloudlib/errors.py
```diff
@@ -156,3 +156,16 @@ def __init__(self, tag: str, rules_failed: List[str]):
     def __str__(self) -> str:
         """Return string representation of the error."""
         return f"Tag '{self.tag}' failed the following rules: {', '.join(self.rules_failed)}"
+
```
aciba90, line 159: Are these exceptions Azure-specific? If so, I think they should live under `pycloudlib/azure/errors.py` and derive from another exception.

pycloudlib/azure/cloud.py
```diff
@@ -782,6 +793,88 @@ def _check_for_network_interfaces(self):
...
+
+        Raises:
+            ValueError: the specified VM size was not found or vCPU capability is invalid.
+        """
+        resource_sku_list = self.compute_client.resource_skus.list(
+            filter=f"location eq '{self.location}'"
+        )
+
+        if not resource_sku_list:
+            raise ValueError(
```
aciba90, line 807: In order to conform with https://github.com/canonical/pycloudlib/issues/263, this and the rest of exceptions in this file should be pycloudlib-specific, to facilitate catching them in call-sites.

pycloudlib/azure/cloud.py
```diff
@@ -782,6 +793,88 @@ def _check_for_network_interfaces(self):
...
+            vcpu_capability.value is None
+            or not vcpu_capability.value.isdigit()
+        ):
+            raise ValueError(
+                f"Invalid vCPU capability value: {vcpu_capability.value} for {instance_type}"
+            )
+
+        return resource_sku.family, int(vcpu_capability.value)
+
+    def _check_compute_quota(self, instance_type: str):
```
aciba90, line 846:
```suggestion
    def _check_compute_quota(self, instance_type: str) -> None:
```

pycloudlib/azure/cloud.py
```diff
@@ -782,6 +793,88 @@ def _check_for_network_interfaces(self):
...
+        """Check if there is sufficient Azure Compute quota for the deployment.
+
+        Raises:
+            QuotaLimitError: the vCPU requirement exceeds the region quota limit.
+            InsufficientQuotaError: the vCPU requirement exceeds the remaining quota.
+        """
+        family, vcpu = self._get_resource_sku_info(instance_type)
+
+        usage_list = self.compute_client.usage.list(self.location)
+        usage = next(
```
aciba90, line 856: This could raise a `StopIteration` exception if the generator expression is empty. Shouldn't we catch it and act in another way?

pycloudlib/azure/cloud.py
```diff
@@ -863,6 +957,17 @@ def launch(
         # resources we are creating.
         self.tag = get_timestamped_tag(self.base_tag)
 
+        try:
+            self._check_compute_quota(instance_type)
+
+        except Exception as e:
+            # Clean up resources created during the failed attempt
+            self._log.error("Exception during launch: %s", e)
+            if self.resource_group:
```
aciba90, line 966: I think this block is not needed, as the class will clean up these resources using the `clean` function if instantiated as a context manager and leak them if not.

aciba90 (changes requested):
Thanks, @r3tro7 and @a-dubs, for this work and making pycloudlib better. I left in-line comments and I have the following concerns:

## Concern 1

We have created a new python package called `utils` to host the new backoff implementation, but we have already a `utils` module.
I think this is counter-intuitive and unnecessarily complex for a newcomer. Why not hosting the backoff implementation within `utils.py`?

## Concern 2

The file structure under `tests` mimics the one under `pycloudlib` but `test_backoff.py` does not convey that rule. For better test discoverability, could we align this.

## Concern 3

As @r3tr7 proposed https://github.com/canonical/pycloudlib/pull/426 and we are using part of its work in the second commit of this PR, I think we should give them co-authorship at least in that commit with a footer with `Co-authored-by: ANOTHER-NAME <ANOTHER-NAME@EXAMPLE.COM>"`.

## Concern 4

@r3tro7: Before launching an Azure instance, we are doing calculations that try to anticipate conditions under which `Azure.launch` will fail due to vcpu quotas and the like.

I wonder if these condition might be misaligned with Azure in the future, and propose the following as an alternative:

Rather than pycloudlib statically checking some of the conditions under which azure sdks will raise exceptions, would be better to directly catch the exceptions the sdk raises and retry on them? That would eliminate the aforementioned coupling and be less error-prone, IMO. What do you, @r3tro7 and @a-dubs, think about this?

a-dubs (commented):
thank you for your detailed review @aciba90. Your diligence and attention to detail is incredibly appreciated. 

To address your 4 concerns:
1 & 2: I will move everything into existing util.py file so this will be resolved. I agree that having an entirely new utils "module" is unnecessary. 
3: I have already reached out to @r3tro7 for their contact info and will add them as co-author in my updated commits once i make all the changes necessary per your feedback.
4: Still unsure on this. Let's sync on this soon.

pycloudlib/errors.py
```diff
@@ -156,3 +156,16 @@ def __init__(self, tag: str, rules_failed: List[str]):
...
+    """Raised when the remaining vCPU quota is insufficient."""
+
+
+class QuotaLimitError(PycloudlibError):
+    """Raised when the vCPU limit will always be exceeded.
+
+    This typically occurs when the default vCPU limit for the selected SKU is
+    zero. Users should not re-attempt deployments when this Exception is raised
+    until the limit has been raised.
+    """
```
a-dubs, line 171: clarify a bit more in these docstrings.

pycloudlib/azure/cloud.py
```diff
@@ -863,6 +957,17 @@ def launch(
...
+                self.delete_resource_group(self.resource_group.name)
+                self.resource_group = None
+            raise
```
a-dubs, line 969: get rid of this cleanup for now. if cleanup errors persist, let's open a new issue and fix in a separate PR.

pycloudlib/azure/cloud.py
a-dubs: make this function public and add type annotation like alberto suggested

pycloudlib/azure/cloud.py
```diff
@@ -782,6 +793,88 @@ def _check_for_network_interfaces(self):
...
+            raise InsufficientQuotaError(
+                f"Insufficient Azure Compute quota: {vcpu} vCPU(s) required, {remaining_vcpu} available."
+            )
+
+    @exponential_backoff(
+        exceptions=(InsufficientQuotaError,),
+        retries=15,
+        max_time=10800,  # 10800 3 hours ceiling to cover random_jitter
+        jitter=True,
+    )
```
a-dubs, line 877: let's nuke the backoff stuff. this belongs in consumers of pycloudlib, not in pycloudlib itself.

pycloudlib/azure/cloud.py
```diff
@@ -665,6 +669,13 @@ def _create_virtual_machine(
         except HttpResponseError as e:
             err_code = e.error.code
             err_msg = e.error.message
+            if (
+                err_code in ("OperationNotAllowed", "QuotaExceeded")
+                and "quota" in err_msg.lower()
+            ):
+                if "limit" in err_msg.lower():
+                    raise QuotaLimitError(err_msg) from e
+                raise InsufficientQuotaError(err_msg) from e
```
a-dubs, line 678: add some inline comments explaining further. Also add "Raises:" section to docstring.

pycloudlib/azure/cloud.py
```diff
@@ -782,6 +793,88 @@ def _check_for_network_interfaces(self):
...
+                f"Insufficient Azure Compute quota: {vcpu} vCPU(s) required, {remaining_vcpu} available."
+            )
+
+    @exponential_backoff(
+        exceptions=(InsufficientQuotaError,),
+        retries=15,
+        max_time=10800,  # 10800 3 hours ceiling to cover random_jitter
+        jitter=True,
+    )
     def launch(
```
a-dubs, line 878: add "raises:" docstring section here too

a-dubs (commented):
met with @r3tro7 and discussed next steps for this PR.

pycloudlib/azure/cloud.py
a-dubs: also remove all custom backoff code. thanks!

PR #437, opened 2024-10-04
fix(gce): convert instance id to str

aciba90:
Recent versions of the google sdk return the instance id as integer rather than string as previously. Convert it to string to conform with the pycloudlib.Instance.id API typing.


blackboxsw (approved):
ship it

PR #438, opened 2024-10-04
feat: add pty support to execute()

holmanb:
Context:
https://github.com/canonical/cloud-init/pull/5784


a-dubs (changes requested):
@holmanb could you please add some context to the commit message and/or the docstrings for `_run_command()` or `_ssh()` explaining what `get_pty` will do?

Also, I am uncertain as to how this change will be used given that these are "private" commands and no changes has been made to the "public" `execute()` function. Is the expectation that the `get_pty` value will just be passed as a kwarg to `execute()`?

holmanb:
> @holmanb could you please add some context to the commit message and/or the docstrings for `_run_command()` or `_ssh()` explaining what `get_pty` will do?
> 
> Also, I am uncertain as to how this change will be used given that these are "private" commands and no changes has been made to the "public" `execute()` function. Is the expectation that the `get_pty` value will just be passed as a kwarg to `execute()`?

Yes, you've got it right. I just added a bit of both.

a-dubs (approved):
Thanks @holmanb !

PR #441, opened 2024-10-11
ci(pytest): modify tox -e pytest to gather branch coverage

a-dubs:
---

Was discovered while working through PRs with @uhryniuk 

## Testing of changes

run `tox -e pytest` and verify that 4 columns (instead of 2) are present in coverage table output at end of pytest run


TheRealFalcon (approved):
Seems a little premature since we're not even close to full statement coverage, but LGTM!

a-dubs:
@TheRealFalcon haha true, but we might as well see all the numbers so we can see exactly how poor our coverage is 🤣

Issue #445, opened 2024-10-16
build status badge is obsolete in README.md


aciba90:
The repo doesn't use Travis anymore for CI. https://github.com/canonical/pycloudlib/blob/335c4a4d9525d78120cc127905ccbd1928046e2d/README.md?plain=1#L3 is obsolete.

JP-sDEV:
Hi @aciba90, could I be assigned this task?

TheRealFalcon:
@JP-sDEV , your comment is enough to signal that you are working on it. We'll consider it yours!

PR #447, opened 2024-10-16
docs: update CI status and included release badge

JP-sDEV:
#445 Updated CI status badge

Replaced the Travis CI badge with Github CI badge. Also included a badge for the status of the `Tag and Publish Release` workflow.


README.md
```diff
@@ -1,6 +1,8 @@
 # pycloudlib
 
-[![Build Status](https://travis-ci.com/canonical/pycloudlib.svg?branch=master)](https://travis-ci.com/canonical/pycloudlib)
+[![Pycloudlib CI](https://github.com/canonical/pycloudlib/actions/workflows/ci.yaml/badge.svg)](https://github.com/canonical/pycloudlib/actions/workflows/ci.yaml)
+
```
aciba90, line 4: Nit: To display both badges on the same line:

```suggestion
```

aciba90 (approved):
LGTM, thanks!

aciba90 (changes requested):
Wait, I think we should limit the badge status to jobs execute from the `main` branch. Otherwise, failures from PRs will show up on the badges. See: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge#using-the-branch-parameter

JP-sDEV:
Understood, I will only include the CI and Tag and Versioning workflows from the `main` branch. I will also put the badges on a single line (e.g. side by side)

JP-sDEV:
Is the failed workflow a config error? or do I need to fix something on my end?

aciba90:
> all the workflows badges in the main branch to be in the README

I think that is what makes sense, thanks!

> Is the failed workflow a config error? or do I need to fix something on my end?

Yes, before merging you need to bump the version in https://github.com/canonical/pycloudlib/blob/main/VERSION.

README.md
```diff
@@ -1,6 +1,8 @@
 # pycloudlib
 
-[![Build Status](https://travis-ci.com/canonical/pycloudlib.svg?branch=master)](https://travis-ci.com/canonical/pycloudlib)
+[![Pycloudlib CI](https://github.com/canonical/pycloudlib/actions/workflows/ci.yaml/badge.svg)](https://github.com/canonical/pycloudlib/actions/workflows/ci.yaml)
```
a-dubs, line 3: Use main_check.yaml instead of ci.yaml since ci.yaml is only for PRs.
```suggestion
[![Pycloudlib Main CI](https://github.com/canonical/pycloudlib/actions/workflows/main_check.yaml/badge.svg)](https://github.com/canonical/pycloudlib/actions/workflows/main_check.yaml)
```

a-dubs (changes requested):
hello @JP-sDEV ! Sorry this has been sitting untouched for so long. Would you mind making yet another fix?

@aciba90 was on the right track suggesting that CI only be for the main branch, but this still does not fully accomplish what we are looking for. We should instead use main_check.yaml. See my inline comment for details. 

Also, would you mind making @aciba90's suggested change so that there is no empty newline between the badges in the markdown, that way the ci badges will display inline (on the same line as each other).

Then once you make these changes, rebase against main and update the VERSION file appropriately (bump the _._.X part of the version) and I will merge!

Thank you in advance.

a-dubs:
@JP-sDEV would you mind rebasing your changes against canonical:main so that there are no conflicts regarding the VERSION file, and then squashing your commits into one commit with the following commit message:
```
docs: update build CI badge and add post merge CI badge in README
```

Thanks!

a-dubs:
This has been inactive for far too long. I will have to pick this back up myself in the future so we can get this change added since it is a nice feature to have.

PR #448, opened 2024-10-17
[DRAFT] Add upload and registering of images

a-dubs:
Do not squash - keep as separate commits


a-dubs:
this is just clogging up the PR queue. will close for now and revisit later if this functionality is actually desired in the future.

PR #451, opened 2024-11-19
chore: add plucky to releases

TheRealFalcon:
Note that I completely guessed on the Azure images as I still don't see them in `az vm image list --all`.


aciba90 (approved):
LGTM, thanks!

PR #453, opened 2024-11-28
refactor: BaseCloud method returns config instead of setting config

danpdraper:
This commit renames the `_check_and_set_config` method of the BaseCloud class to `_check_and_get_config`, and modifies the method such that it returns a dict instead of setting the config instance variable. This change facilitates modification of the method's behaviour during unit testing.


a-dubs (approved):
Thanks for fixing this!

PR #454, opened 2024-12-11
feat: Add the ability to specify which Oracle VCN to use through the config file

Bryzizzle:
When there are multiple virtual cloud networks (VCNs) available on an Oracle compartment, pycloudlib will use the newest VCN available, which might not be desired. Currently, the VCN selection functionality is present within the cloud class level, however it cannot be set through the config file. This patch modifies the `pycloudlib/oci/cloud.py` file to retrieve the `vcn_name` key from config and use it to force a VCN to use. A commented entry is also included in the `pycloudlib.toml.template` for reference.


a-dubs:
fire in the hole! soooooooooooo glad to finally be able to use this config from my toml file

a-dubs (approved):
lgtm!! this is a good bandaid solution until we do the full configuration overhaul!

PR #456, opened 2025-01-15
Oracle-bm-timeout-fix

a-dubs:
do not squash merge, rebase merge to keep the two separate commits please!


uhryniuk (approved):
:vulcan_salute:

a-dubs:
no idea what is broken with the ci-integration-tests job :frowning_face: 
but ignoring that CI should be good to go

pycloudlib/oci/instance.py
```diff
@@ -168,6 +168,28 @@ def start(self, wait=True):
         if wait:
             self.wait()
 
+    def wait(self, **kwargs):
```
TheRealFalcon, line 171: I think it'd be better to modify the base class instead. Something like `def wait(self, timeout=40, **kwargs):` (with an updated docstring explaining why 40) that gets passed through to `_wait_for_execute()`. The current hardcoded-for-ec2 option isn't great.

TheRealFalcon (approved):
I'm ok with it as-is, but I have a suggestion inline.

pycloudlib/ec2/instance.py
```diff
@@ -15,6 +15,7 @@ class EC2Instance(BaseInstance):
     """EC2 backed instance."""
 
     _type = "ec2"
+    ready_timeout = 40  # To keep backwards compatibility (will lower soon)
```
TheRealFalcon, line 18:
```suggestion
    ready_timeout = 40  # To keep backwards compatibility (will lower soon™)
```
(this is a joke, don't actually apply this)

pycloudlib/ibm_classic/instance.py
```diff
@@ -285,13 +285,6 @@ def is_started():
         logger.info("Instance %s started", self.name)
         self._instance = self._vs_manager.get_instance(self.id)
 
-    def wait(self, **kwargs):
```
TheRealFalcon, line 288: we need to add `ready_timeout = 180` to this class, correct?

TheRealFalcon (commented):
One more question inline.

pycloudlib/ibm_classic/instance.py
TheRealFalcon: Nevermind...not needed.

TheRealFalcon (approved):
LGTM!

PR #461, opened 2025-01-29
fix(oracle): fix subnet selection logic

a-dubs:
When a subnet is not associated with an availability domain (AD),
we should not ignore the subnet just because it does not match our
configured AD. This change fixes the subnet selection logic to first
check if the subnet is associated with an AD before checking if it
matches the configured AD.

Before this change, the subnet selection logic was flawed and would
say it was using a certain subnet, but due to this bad AD logic, it
would actually not actually use the subnet it said it was using.

New unit test file was added that provides 100% coverage of the
get_subnet_id function!


uhryniuk (approved):

pycloudlib/oci/utils.py
```diff
@@ -89,23 +89,36 @@ def get_subnet_id(
         if len(vcns) > 1:
             raise PycloudlibError(f"Found multiple vcns with name: {vcn_name}")
         vcn_id = vcns[0].id
+        chosen_vcn_name = vcns[0].display_name
```
TheRealFalcon, line 92: Nit: Both the vcn id and chosen vcn name could be set after the if/else branches rather than being duplicated.

TheRealFalcon (approved):
One nit inline but otherwise LGTM! Nice test additions!

pycloudlib/oci/utils.py
a-dubs: hmmm fair enough. will do!

PR #472, opened 2025-02-20
various CI improvements

a-dubs:
## Description

- Add PR template to repository to make PRs more standardized and easier to review (this PR description is using it!)
- replaced tox `pytest` env in default list with new `py38` env that explicitly runs pytest using python 3.8 to ensure that code being tested does not introduce anything incompatible with our minimum supported version (3.8).
- replaced pylint with ruff's pylint implementation because the pylint env would take up to 30s, whereas the ruff check only takes a few seconds. this makes running CI locally much more enjoyable.


codyshepherd:
Can you describe which changes you consider breaking and which are not?

.github/pull_request_template.md
```diff
@@ -0,0 +1,69 @@
...
+    <_O_O_O_O_O_O_O_O_O_O_O_O_____/   \
+
+-->
+
+## PR Checklist
+
+To ease the process of reviewing your PR, do make sure to complete the following checklist **before** submitting a pull
+request.
+
+- [ ] I have added unit tests to cover the new behavior under ``tests/unit_tests/```
```
rpocase, line 15: would it be worthwhile to add [pre-commit](https://github.com/pre-commit/pre-commit) and include in the contributing guide?

pycloudlib/qemu/util.py
```diff
@@ -10,7 +10,7 @@
 
 def get_free_port():
     """Look for a free local port for SSHing to VM."""
-    global next_port  # pylint: disable=global-statement
+    global next_port  # noqa: PLW0603
```
codyshepherd, line 13: what does this code represent and why is it being skipped? some additional commentary here would help clarify.

tox.ini
```diff
@@ -1,14 +1,13 @@
 [tox]
 # As it may be undesired to make formatting changes, by default only check
-envlist = ruff, format-check, mypy, pytest, pylint
+envlist = ruff, format-check, mypy, py38
```
codyshepherd, line 3: Why not include py310 and py312 in this list?

codyshepherd (changes requested):
Can you help me understand how/where ruff's pylint has been subbed for traditional pylint? I see that the `pylint` command has been removed from the tox.ini line 66, is the `ruff check` command in line 78/84 already doing the ruff linting, and this PR just disabled the duplicate effort?

pycloudlib/qemu/util.py
a-dubs: I was just converting the ignore from pylint to ruff's equivalent so it respects this existing "noqa".

tox.ini
a-dubs: Good question. I felt it was somewhat redundant since 3.10 and 3.12 will be backwards compatible with 3.8 and didn't want to run essentially redundant tests, but to be fair the tests only take a handful of seconds to run so I would be happy to add py310 and py12 as default envs as well!

pyproject.toml
```diff
@@ -48,6 +48,9 @@ docstring-code-format = true
 extend-select = [
   "D",   # pydocstyle
   "I",   # isort
+  "PLC",  # Pylint conventions
+  "PLE",  # Pylint errors
+  "PLW",  # Pylint warnings
```
a-dubs, line 53: @codyshepherd this is how ruff is replacing pylint. By adding the 3 core sets of ruff rules for pylint.

a-dubs: yes that is correct! it is the `ruff check` command

.github/pull_request_template.md
a-dubs: I agree that this could be a very good value add. Given that this is not something we currently do as standard, I am going to capture this in a new GH issue so proper investigation can be done on this without holding up this simpler PR.

a-dubs:
@codyshepherd I have responded to all of your comments and also added a new commit that creates dedicated py38, py310, and py312 github action jobs for PRs to address your comment.

Ready for re-review!

uhryniuk (approved):
Looks like you addressed most of the comments, keep an eye after this gets merged to make sure the CI doesn't explode.

pycloudlib/qemu/util.py
codyshepherd: Can you please add additional commentary to this noqa that explains what is being skipped and why

tox.ini
```diff
@@ -22,33 +21,43 @@ envdir = {[common]envdir}
...
+    {envpython} -m ruff check {posargs:pycloudlib examples setup.py}
+    {envpython} -m ruff format --check {posargs:pycloudlib examples setup.py}
 
 [testenv:format]
 envdir = {[common]envdir}
 deps = {[common]deps}
 commands =
-    {envpython} -m ruff format -- .
-    {envpython} -m ruff check --fix -- pycloudlib examples setup.py
+    {envpython} -m ruff format {posargs:pycloudlib examples setup.py}
```
codyshepherd, line 59: Will this command change the formatting in place? i.e. will it make edits to the code if there are edits to be had?

codyshepherd (changes requested):

tox.ini
codyshepherd: Shall we just remove 3.10 and 3.12 then? what's the value of keeping them in if they aren't run?

tox.ini
a-dubs: yes, as is standard for most of our projects, the `format` env will fix linting issues and do formatting in place. for this project, the `ruff` command will run the same commands but will only check and will not make any changes.

codyshepherd: Ah, I see, thanks.

Issue #379, opened 2024-05-23
Use case: multi-nic ipv6-only launch


TheRealFalcon:
In cloud-init, we'd like to integration test scenarios that involve ipv6-only subnets and multinic already setup on first boot. We first need the plumbing in pycloudlib to make it happen.

PR #473, opened 2025-02-24
V11 - add networking config types

a-dubs:
## Description

This PR seeks to add the ability to uniformly identify and specify networking config across clouds, with initial sweeping implementation being made to Oracle cloud. 

The creation of a new `NetworkingType` enum and a `NetworkingConfig` dataclass should hopefully make it easy and standardized to represent networking for an instance or any of its NICs across clouds.

## Test Steps

- Ran the new integration tests I wrote:
  - `tox -e integration-tests -- tests/integration_tests/oracle/test_launch.py`
  - `tox -e integration-tests -- tests/integration_tests/oracle/test_utils.py`
- Ran the full `test_public_api` CI locally (`tox -e integration-tests-ci`)
- Added significant unit test coverage over modified oracle functions relating to networking configs
- Ran coverage locally and compared with my changed code and verified any new call sites have either direct unit or integration test coverage

### DIFF
--- a/pycloudlib/oci/cloud.py
+++ b/pycloudlib/oci/cloud.py
@@ -20,11 +20,13 @@ from pycloudlib.errors import (
 )
 from pycloudlib.oci.instance import OciInstance
 from pycloudlib.oci.utils import (
+    generate_create_vnic_details,
     get_subnet_id,
     get_subnet_id_by_name,
     parse_oci_config_from_env_vars,
     wait_till_ready,
 )
+from pycloudlib.types import NetworkingConfig
 from pycloudlib.util import UBUNTU_RELEASE_VERSION_MAP, subp
 
 
@@ -251,6 +253,7 @@ class OCI(BaseCloud):
             availability_domain=self.availability_domain,
             oci_config=self.oci_config,
             username=username,
+            vcn_name=self.vcn_name,
         )
 
     def launch(
@@ -263,6 +266,7 @@ class OCI(BaseCloud):
         username: Optional[str] = None,
         cluster_id: Optional[str] = None,
         subnet_name: Optional[str] = None,
+        primary_network_config: Optional[NetworkingConfig] = None,
         **kwargs,
     ) -> OciInstance:
         """Launch an instance.
@@ -273,12 +277,14 @@ class OCI(BaseCloud):
                 https://docs.cloud.oracle.com/en-us/iaas/Content/Compute/References/computeshapes.htm
             user_data: used by Cloud-Init to run custom scripts or
                 provide custom Cloud-Init configuration
-            subnet_name: string, name of subnet to use for instance.
             retry_strategy: a retry strategy from oci.retry module
                 to apply for this operation
             username: username to use when connecting via SSH
             vcn_name: Name of the VCN to use. If not provided, the first VCN
                 found will be used
+            subnet_name: string, name of subnet to use for instance.
+            primary_network_config: NetworkingConfig object to use for configuring the primary
+                network interface
             **kwargs: dictionary of other arguments to pass as
                 LaunchInstanceDetails
 
@@ -297,6 +303,7 @@ class OCI(BaseCloud):
                 self.compartment_id,
                 self.availability_domain,
                 vcn_name=self.vcn_name,
+                networking_config=primary_network_config,
             )
         metadata = {
             "ssh_authorized_keys": self.key_pair.public_key_content,
@@ -314,6 +321,10 @@ class OCI(BaseCloud):
             image_id=image_id,
             metadata=metadata,
             compute_cluster_id=cluster_id,
+            create_vnic_details=generate_create_vnic_details(
+                subnet_id=subnet_id,
+                networking_config=primary_network_config,
+            ),
             **kwargs,
         )
 
@@ -327,6 +338,32 @@ class OCI(BaseCloud):
         )
         self.created_instances.append(instance)
         return instance
+
+    def find_compatible_subnet(self, networking_config: NetworkingConfig) -> str:
+        """
+        Automatically select a subnet that is compatible with the given networking_config.
+
+        In this case, compatible means that the subnet can support the necessary networking type
+        (ipv4 only, ipv6 only, or dual stack) and the private or public requirement.
+        This method will select the first subnet that matches the criteria.
+
+        Args:
+            networking_config: NetworkingConfig object to use for finding a subnet
+
+        Returns:
+            id of the subnet selected
+
+        Raises:
+            `PycloudlibError` if unable to determine `subnet_id` for the given `networking_config`
+        """
+        subnet_id = get_subnet_id(
+            network_client=self.network_client,
+            compartment_id=self.compartment_id,
+            availability_domain=self.availability_domain,
+            vcn_name=self.vcn_name,
+            networking_config=networking_config,
+        )
+        return subnet_id
 
     def snapshot(self, instance, clean=True, name=None):
         """Snapshot an instance and generate an image from it.
--- a/pycloudlib/oci/instance.py
+++ b/pycloudlib/oci/instance.py
@@ -10,7 +10,13 @@ import oci
 
 from pycloudlib.errors import PycloudlibError
 from pycloudlib.instance import BaseInstance
-from pycloudlib.oci.utils import get_subnet_id, get_subnet_id_by_name, wait_till_ready
+from pycloudlib.oci.utils import (
+    generate_create_vnic_details,
+    get_subnet_id,
+    get_subnet_id_by_name,
+    wait_till_ready,
+)
+from pycloudlib.types import NetworkingConfig
 
 
 class OciInstance(BaseInstance):
@@ -27,6 +33,7 @@ class OciInstance(BaseInstance):
         oci_config=None,
         *,
         username: Optional[str] = None,
+        vcn_name: Optional[str] = None,
     ):
         """Set up the instance.
 
@@ -46,6 +53,7 @@ class OciInstance(BaseInstance):
         self.availability_domain = availability_domain
         self._fault_domain = None
         self._ip = None
+        self._vcn_name: Optional[str] = vcn_name
 
         if oci_config is None:
             oci_config = oci.config.from_file("~/.oci/config")  # noqa: E501
@@ -145,7 +153,8 @@ class OciInstance(BaseInstance):
             for vnic_attachment in vnic_attachments
         ]
         secondary_vnic_attachment = [vnic for vnic in vnics if not vnic.is_primary][0]
-        return secondary_vnic_attachment.private_ip
+        self._log.debug("secondary vnic attachment data:\n%s", secondary_vnic_attachment)
+        return secondary_vnic_attachment.private_ip or secondary_vnic_attachment.ipv6_addresses[0]
 
     @property
     def instance_data(self):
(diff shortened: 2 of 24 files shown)


pycloudlib/oci/utils.py
```diff
@@ -84,14 +82,126 @@ def get_subnet_id_by_name(
     return subnets[0].id
 
 
+def _get_subnet_features(
+    subnet: oci.core.models.Subnet,
+) -> Dict[str, Any]:
+    """
+    Get the availability domain, private, and networking type of the subnet.
```
codyshepherd, line 89: This could be reworded for clarity; e.g. "get the private of the subnet" doesn't make a ton of immediate sense.

pycloudlib/oci/utils.py
```diff
@@ -84,14 +82,126 @@ def get_subnet_id_by_name(
...
+    Returns:
+        A dictionary containing the following keys:
+        - availability_domain: The availability domain of the subnet.
+        - private: Whether the subnet is private.
+        - networking_type: The networking type of the subnet
+
+    """
+    availability_domain = subnet.availability_domain
+    private = subnet.prohibit_internet_ingress
+    has_ipv4_cidr_block = subnet.cidr_block not in (None, "<null>")
```
codyshepherd, line 103: It might be worth putting the string `"<null>"` into a named variable at module scope to provide a bit of self-documentation for where this particular string comes from; at first glance it looks like an erroneously-applied javascript convention. I'm guessing this comes from the way the OCI api produces text/json output, but there's an opportunity here to make that more immediately clear.

pycloudlib/oci/utils.py
```diff
@@ -84,14 +82,126 @@ def get_subnet_id_by_name(
...
+    availability_domain: str,
+    networking_config: NetworkingConfig,
+) -> bool:
+    """
+    Check if the subnet is compatible with the given restrictions.
+
+    For each restriction, the following must be true:
+    availability_domain:
+    - if the subnet has an availability domain, it must match the given availability domain
+    - otherwise, it is compatible
```
codyshepherd, line 136: It seems like there are more possibilities than are described here.

* the subnet has an AD
  * the subnet's AD matches the given AD --> compatible
  * the subnet's AD doesn't match the given AD -> incompatible
* the subnet does not have an AD -> compatible

It's a bit ambiguous which condition the word "otherwise" applies to -- having an AD, or matching the AD.

pycloudlib/oci/utils.py
```diff
@@ -84,14 +82,126 @@ def get_subnet_id_by_name(
...
+
+    Returns:
+        True if the subnet is compatible, False otherwise.
+    """
+    networking_type = networking_config.networking_type
+    private = networking_config.private
+
+    # to do this, lets get the subnet features
+    features = _get_subnet_features(subnet)
+    log.info("Subnet features: %s", features)
```
codyshepherd, line 163: Why is this logging as INFO when the other logs are logging as DEBUG?

pycloudlib/oci/utils.py
```diff
@@ -237,3 +337,38 @@ def parse_oci_config_from_env_vars() -> Optional[Dict[str, str]]:
...
+    Args:
+        subnet_id: The subnet id to use for the VNIC.
+        networking_config: The network configuration to use for the VNIC.
+
+    Returns:
+        vnic_details: The VNIC details object.
+    """
+    # default to IPv4 with public IP
+    # this will be used if networking_config is not provided or if set to AUTO
+    vnic_details = oci.core.models.CreateVnicDetails(  # noqa: E501
```
codyshepherd, line 358: it's helpful with `noqa` lines to add a brief justification for why the check is being skipped.

tests/unit_tests/oci/test_utils.py
```diff
@@ -137,15 +155,17 @@ def test_get_subnet_id_suceeds_with_vcn_name(self, setup_environment):
...
+            network_client=network_client,
+            compartment_id=compartment_id,
+            availability_domain=availability_domain,
             vcn_name="vcn_name",
         )
         assert result == "subnet_id"
         # Ensure that list_subnets is called with the specified VCN's ID
         network_client.list_subnets.assert_called_with(
-            compartment_id, vcn_id="vcn_id", retry_strategy=DEFAULT_RETRY_STRATEGY
+            compartment_id,
```
codyshepherd, line 166:
```suggestion
            compartment_id=compartment_id,
```

pycloudlib/oci/cloud.py
```diff
@@ -286,6 +295,7 @@ def launch(
             self.compartment_id,
             self.availability_domain,
             vcn_name=self.vcn_name,
+            networking_config=primary_network_config,
```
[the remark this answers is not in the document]
a-dubs: that is the idea yes. but now that you have brought this up, I went and looked at the Oracle interface for launching instances, and it is seems like in most cases they do NOT support attaching a secondary vnic before/at launch. So maybe it would be best to remove the `primary` from the variable name like you suggest, and then if you want a 2nd NIC, just call the `attach_network_interface` command after launch.

a-dubs:
@codyshepherd  thank you so much for such a detailed review. I did a lot more manual testing and added some pretty comprehensive integration tests for the oracle changes and I feel much better about the functionality of this PR now. 

Ready for re-review whenever you have the time.

pycloudlib/oci/cloud.py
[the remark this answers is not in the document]
a-dubs: okay well looking around some more, I found Pycloudlib issue #379 that requests that we add the infrastructure/plumbing to be able to specify multiple vnics at instance launch. Even if this is not something that we are able to do on Oracle, it sounds like other major clouds should have the ability to specify primary networking config and then any number of optional networking configs for additional nics that will be configured at launch.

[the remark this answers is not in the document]
a-dubs: so i think I am just going to leave this as is for now for clarity.

pycloudlib/oci/cloud.py
```diff
@@ -317,6 +332,32 @@ def launch(
...
+        Returns:
+            id of the subnet selected
+
+        Raises:
+            `PycloudlibError` if unable to determine `subnet_id` for the given `networking_config`
+        """
+        subnet_id = get_subnet_id(
+            self.network_client,
+            self.compartment_id,
+            self.availability_domain,
```
codyshepherd, line 355:
```suggestion
            network_client=self.network_client,
            compartment_id=self.compartment_id,
            availability_domain=self.availability_domain,
```

pycloudlib/oci/instance.py
```diff
@@ -304,13 +317,17 @@ def add_network_interface(
...
         vnic_data = self.network_client.get_vnic(vnic_attachment_data.vnic_id).data
+        self._log.debug(
+            "Newly attached vnic data:\n%s",
+            vnic_data,
+        )
         self._log.info(
-            "Added network interface with private IP %s to instance %s on nic #%s",
-            vnic_data.private_ip,
+            "Added network interface with IP %s to instance %s on nic #%s",
+            vnic_data.private_ip or vnic_data.ipv6_addresses[0],
```
codyshepherd, line 326: Not necessarily a blocker, but this index access runs the risk of becoming a runtime error; it assumes that if private_ip was not passed then ipv6_addresses will be populated. I don't know what kind of guarantees we have that that assumption will hold, which is why it's not a hard blocker. However, it could be an opportunity to check that assumption prior to attempting the index access.

Also, the index access is repeated; could be an opportunity to just throw that value into a variable, check it, and reuse it rather than repeating the index access. Just a though.

pycloudlib/oci/utils.py
```diff
@@ -84,14 +90,132 @@ def get_subnet_id_by_name(
...
+    For each restriction, the following must be true:
+    availability_domain:
+    - if the subnet is tied to an availability domain, it must match the given availability domain
+    - if the subnet is not tied to an availability domain, then it is automatically compatible
+
+    From the networking_config, we have the following restrictions:
+
+    private:
+    - if private is true, then `prohibit_internet_ingress` must be True
+    - if false (public), then `prohibit_internet_ingress` must be False
```
codyshepherd, line 153: It's  a bit odd to describe this restriction in terms of `prohibit_internet_ingress`, as no key by that name is ever used in this function. I see use of `features["private"]`, so maybe this docstring should be reworded?

pycloudlib/oci/utils.py
```diff
@@ -84,14 +90,132 @@ def get_subnet_id_by_name(
...
+    private_compatible = private == features["private"]
+    availability_domain_compatible = (
+        features["availability_domain"] is None
+        or features["availability_domain"] == availability_domain
+    )
+    compatible = (
+        networking_type_compatible and private_compatible and availability_domain_compatible
+    )
+    if not compatible:
+        log.debug(
```
codyshepherd, line 197: Would this be better as a Warning? I see warning used above when networking type can't be determined, and the caller logs INFO if it is compatible.

pycloudlib/oci/utils.py
```diff
@@ -105,12 +229,23 @@ def get_subnet_id(
...
         id of the subnet selected
     Raises:
-        `Exception` if unable to determine `subnet_id` for
-        `availability_domain`
+        `PycloudlibError` if unable to determine `subnet_id` for `availability_domain`,
+        or if no relevant VCNs are found in the compartment.
     """
+    if not networking_config:
+        networking_config = NetworkingConfig()
+        log.debug(
```
codyshepherd, line 242: This might also be appropriate as a warning

tests/unit_tests/oci/test_utils.py
```diff
@@ -402,3 +429,168 @@ def test_get_subnet_id_by_name_succeeds(self):
...
+            availability_domain,
+            vcn,
+        ) = setup_environment
+
+        # Create subnet mocks based on the parameters
+        ipv4_public_subnet = MagicMock()
+        ipv4_public_subnet.availability_domain = None
+        ipv4_public_subnet.prohibit_internet_ingress = False
+        ipv4_public_subnet.cidr_block = "10.0.0.0/24"
```
codyshepherd, line 477: These string literals could be put into variables to improve DRYness

codyshepherd (approved):
I've made some additional suggestions but none is a hard blocker. This is a lot of work, nice job.

pycloudlib/oci/utils.py
a-dubs: I think the idea here is that we would be looping through several potential subnets and expect only one or few to be compatible, whereas many will be incompatible.

tests/unit_tests/oci/test_utils.py
a-dubs: I agree with the sentiment, but after trying to do so, none of the IP addresses or id's are re-used so this would not improve dryness in this specific case. Really like the suggestion tho!

pycloudlib/oci/utils.py
a-dubs: thanks! great catch. this was leftover from when the code inside `_get_subnet_features()` lived in this function.

pycloudlib/oci/instance.py
a-dubs: agreed! added code to catch this index error at runtime AND moved it to a variable to improve dryness. :bird::bird: = :rock:

uhryniuk (approved):

codyshepherd (approved):

PR #474, opened 2025-02-25
feat(oracle): specify which subnet an instance must use

uhryniuk:


uhryniuk:
Tox is failing on `ruff check` but there are 96 files that need to be formatted? Not sure how that's possible but going to assume it's someone recently added that to the tox env.

uhryniuk:
Closing due to tox CI failing on strange ruff format error. Reopening on a new branch

PR #475, opened 2025-02-25
feat(oracle): specify which subnet an instance must use

uhryniuk:
The default behaviour for assigning subnets is to iterate over the VCN and take the first available subnet then assign it. The following changes allow the ability to specifically set which subnet in the VCN should be used for the instance. This is particularly helpful in a Kubernetes environment where many subnets exist on a single VCN.


a-dubs (approved):
Looks good! Be sure to get tox fixed and then lgtm!

codyshepherd (changes requested):
Can you add a description to the PR?

uhryniuk:
@codyshepherd Done :saluting_face:

codyshepherd:
LGTM except for the test failures

uhryniuk:
> LGTM except for the test failures

Finally resolved, I borked some context manager syntax and seems to be gouda now.

codyshepherd (approved):
thanks!

PR #484, opened 2025-05-07
fix: openstack needs to use neutron for floating ip association

blackboxsw:
## Description
Nova api microversion 2.44 doesn't support associating floating IPs via to compute API endpoint anymore.

Associating floating IPs to servers must be done via the neutron(network) API endpoints instead to avoid 404s during floating IP association to servers using the nova(compute) endoint.

Add an example/ostack.py basic lifecycle script to ease local testing of openstack instance lifecycle.

## PR Checklist

request

- [ ] I have added unit tests to cover the new behavior under ``tests/unit_tests/```
- [ ] I have run `tox -e format` locally to automatically format my code before submitting
- [ ] I have run `tox` locally ensuring that it passes before submitting
- [ ] (if applicable) I have added a reference to issues that this PR relates to in the PR message (Refs GH-1234, Fixes GH-1234)
- [ ] My commits are atomic and follow the convetional commit message format (https://www.conventionalcommits.org/en/v1.0.0/)

Otherwise, please leave the PR as a draft to indicate that it is still a work in progress.
-->

Upstream docs stating deprecated behavior for floating IPs in nova/compute
https://docs.openstack.org/api-ref/compute/#add-associate-floating-ip-addfloatingip-action-deprecated


examples/ostack.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
TheRealFalcon, line 1: Is there a reason not to call this file openstack.py?

TheRealFalcon (commented):
Change looks good to me!

I'm assuming your unit test today is for this PR?

examples/ostack.py
TheRealFalcon: Ahh, right. Import errors are fun. open_stack is fine, or maybe openstack_example.py. 

Whatever seems best to you.

TheRealFalcon (approved):
LGTM!

blackboxsw:
Dropped py38 from CI as ubuntu-20-04 environments are no longer supported in github workflows

PR #496, opened 2026-01-10
update cloud-init clean cmdline remove additional network, machine-id and cloud artifacts

blackboxsw:
- [ ] (if applicable) I have added a reference to issues that this PR relates to in the PR message (Refs GH-1234, Fixes GH-1234)

## Description
Add functionality to use new cloud-init clean params to remove stale cloud-init artifacts from previous boots.
Retain fallback for older versions of cloud-init which may not yet support the `-c all` param.

cloud-init version 23.1 introduced new commandline options to cloud-init clean:
1.  `--machine-id` to uninitialize /etc/machine-id
2. `-c all` which remove previous boot artifacts such as /etc/netplan/50-cloud-init.yaml and and cloud artifacts (generated by azure)


blackboxsw (commented):
Thank you for the review @hammerstefan and the context on CPC test matrix and image build process. We'll keep that in mind for breaking changes on ESM-based image streams such as Ubuntu Pro.   I've adjusted the cloud-init clean call to specify `--config` per your review.

pycloudlib/instance.py
```diff
@@ -232,15 +232,11 @@ def clean(self):
...
-        # bionic-pro cloud images.
-        #
-        # [1] https://github.com/canonical/cloud-init/commit/abfdf1d83995cc20e
-        self.execute("sudo cloud-init clean --logs")
-        self.execute("sudo echo 'uninitialized' > /etc/machine-id")
+        result = self.execute("sudo cloud-init clean --logs --machine-id --config all")
+        if result.failed and "unrecognized arguments" in result.stderr:
+            # Cope with cloud-init version < 23.4 which has no -c argument
+            # or version < 23.1 which has no --machine-id argument.
+            result = self.execute("sudo cloud-init clean --logs")
```
holmanb, line 239: Doesn't this change machine-id behavior?

tests/unit_tests/test_instance.py
```diff
@@ -15,31 +14,31 @@
 
 
 @pytest.fixture
-def concrete_instance_cls():
+def concrete_instance_cls(mocker):
```
holmanb, line 17: Mixing functional and non-functional changes complicates code review.

pycloudlib/instance.py
holmanb, line 239: This ignores failures.

holmanb (changes requested):

pycloudlib/instance.py
rpocase, line 239: would it be better to tailor the cloud-init command based on version range?

PR #507, opened 2026-06-26
chore: migrate dependency management to uv

rpocase:
## PR Checklist

request.

- [-] (if applicable) I have added a reference to issues that this PR relates to in the PR message (Refs GH-1234, Fixes GH-1234)

Otherwise, please leave the PR as a draft to indicate that it is still a work in progress.

## Description

this change migrates dependencies to be properly managed by uv instead of specified in requirements files,
upgrades mypy, and fixes newly found type hinting issues. this is in preparation for upgrading
openstack-sdk as netiface is orphaned and should be migrated to netiface2

this was triggered in part because running mypy was failing because of a missing typed-ast package.
at the same time, i've upgraded mypy and resolved all new type hinting issues.

## Test Steps

N/A? i think integration tests should be run, but i don't have a proper setup to facilitate this


rpocase:
@blackboxsw if you have a moment

holmanb (commented):
Why uv rather than pip?

rpocase:
> Why uv rather than pip?

1. i misread the repo and thought it was using uv to manage installation but wasn't using pyproject.toml for dependency management
2. more modern tooling that is become increasingly pervasive
3. we should look at integrating AGENTS.md and providing a sandboxing environment. i would go with workshops. the only supported path forward to easily bootstrap is with the uv plugin. it would have to be a lot more customized if we stuck with pip

pyproject.toml
```diff
@@ -1,6 +1,93 @@
...
+    "sphinx",
+    "sphinx_rtd_theme",
+]
+dev = [
+    "mypy>=1.13",
+    "ruff==0.5.0",
+    "types-mock>=5.0.0",
+    "types-pyyaml>=6.0.12",
+    "types-requests>=2.31.0",
+    "types-setuptools>=67.0.0",
```
blackboxsw: It seems we are reducing the minimum version of types-setuptools from 67.3.0 in main to 67.0.0.

```suggestion
    "types-setuptools>=67.3.0",
```

pyproject.toml
```diff
@@ -1,6 +1,93 @@
...
+    "myst-parser",
+    "sphinx",
+    "sphinx_rtd_theme",
+]
+dev = [
+    "mypy>=1.13",
+    "ruff==0.5.0",
+    "types-mock>=5.0.0",
+    "types-pyyaml>=6.0.12",
+    "types-requests>=2.31.0",
```
blackboxsw, line 70: Why are we reducing the minimum version floor here? main: `2.31.0.6`    this PR `2.31.0.` ?

```suggestion
    "types-requests>=2.31.0.6",
```

pyproject.toml
```diff
@@ -1,6 +1,93 @@
...
+    "packaging",
+]
+[dependency-groups]
+dev = [
+    "mypy>=1.13",
+    "ruff==0.5.0",
+    "types-mock>=5.0.0",
+    "types-pyyaml>=6.0.12",
+    "types-requests>=2.31.0",
+    "types-setuptools>=67.0.0",
```
blackboxsw: Double check reason for the reduction in dependency floor. We probably should keep at least the same floor as main if we are not increasing minimum versions in this PR.

pyproject.toml
```diff
@@ -1,6 +1,93 @@
...
+docs = [
+    "myst-parser",
+    "sphinx",
+    "sphinx_rtd_theme",
+]
+dev = [
+    "mypy>=1.13",
+    "ruff==0.5.0",
+    "types-mock>=5.0.0",
+    "types-pyyaml>=6.0.12",
```
blackboxsw, line 69: Now that we are unpinning specific versions, we should watch pycloudlib CI more often to ensure we aren't getting affected by changes in remote unpinnned dependencies. If this becomes a problem we can attempt to place a version ceiling to avoid external breaking changes where possible. I don't think we need a change to this PR, just an agreement to look at adding dependency ceilings if we start seeing breakage due to upstream external dependency changes.

docs/index.rst
```diff
@@ -49,7 +46,7 @@ for more information.
 Bugs
 ****
 
-File bugs on Launchpad under the `pycloudlib project <https://bugs.launchpad.net/pycloudlib/+filebug>`_.
+File bugs on `Launchpad <https://bugs.launchpad.net/pycloudlib/+filebug>`_.
```
blackboxsw: This project doesn't even exist in launchpad. Let's update this to https://github.com/canonical/pycloudlib/issues

pycloudlib/lxd/instance.py
```diff
@@ -494,13 +494,13 @@ def wait_for_state(self, desired_state: str, num_retries: int = 100):
             time.sleep(1)
         raise PycloudlibTimeoutError
 
-    def wait_for_stop(self):
+    def wait_for_stop(self, **kwargs):
```
blackboxsw: We're sprinkling a lot of **kwargs  to satisfy mypy warnings, but I think the absence of such **kwargs in the subclassed methods was intentional to show that those methods don't expect variadic arguments being passed through. I don't think we want to silently unintentionally call subclasses with ignored kwargs. We want an error in these cases. Can we instead use # noqa or something to avoid the mypy warnings here?

blackboxsw (commented):
Thank you @rpocase for this proposal this looks like a good direction to me. Let's iron out a few of these concerns before moving forward please.
 
1. I think we are missing changes to `docs/Makefile` related to `pip install -rrequirements.txt` which no longer exists. `cd docs; make deps` fails.2. I'd like to use your separate commit messages as the produce more useful  context for each commit, but the uv.lock file is introduced in the first commit and removed in the third commit. Can we drop that delta from both first as last commits
 
3.  I left inline comments about changes in the supported floor version of dependencies, can we ensure we aren't getting a lower/older supported minimum version than we already had in main?
4.  inline comment about using github issues instead of a launchpad project that doesn't exist.

5. I'm seeing with CI that is erroring  at the moment with this uv branch on 'new prs' is [the following lint errors](https://github.com/blackboxsw/pycloudlib/actions/runs/28548653877/job/84640244076?pr=1):```
pycloudlib/azure/instance.py:288: error: Unused "type: ignore" comment  [unused-ignore]
pycloudlib/azure/instance.py:311: error: Unused "type: ignore" comment  [unused-ignore]
pycloudlib/azure/instance.py:362: error: Unused "type: ignore" comment  [unused-ignore]
```

Note: I can confirm though both integration tests in ubuntu-pro-client and in cloud-init that this changeset doesn't introduce issues with CICD there.

Also, the general dep installation improvement time in cold/empty cache is a significant improvement:
  - uv-based cold : ~13.5 seconds
  - uv-based warm: ~1.5 seconds
  - pip-based cold:  ~64 seconds
  - pip-based warm: ~43 seconds

I like the improvements this represents for the project. thank you.

**Review by blackboxsw — CHANGES_REQUESTED** [review]

File: pyproject.toml [hunk 7]
```diff
@@ -1,6 +1,93 @@
 [build-system]
-requires = ["setuptools"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+dynamic = ["version"]
```
blackboxsw, line 6: missing a license declaration as we had in setup.cfg

pycloudlib/lxd/instance.py
rpocase: happy to revert for now. we should consider a better wait to handle this in the future. the inheritance scheme makes the lack of `**kwargs` feels like a broken contract system. `# noqa` isn't ideal, but at least doesn't change API until we revisit

pycloudlib/lxd/instance.py
```diff
@@ -530,17 +530,17 @@ def _wait_for_instance_start(self):
...
         """Run command in the instance."""
         if self.execute_via_ssh:
-            return super()._run_command(command, stdin)
+            return super()._run_command(command, stdin, get_pty=get_pty)
 
         if self.series == "xenial":
             self._log.warning(MISSING_AGENT_MSG, "lxc exec")
 
-        return super()._run_command(command, stdin)
+        return super()._run_command(command, stdin, get_pty=get_pty)
```
rpocase, line 541: @blackboxsw how do you want to handle this? this is another case where the superclass explicitly supports it and the subclass has broken the subclass contract.

.github/workflows/ci.yaml
```diff
@@ -10,49 +10,54 @@ jobs:
...
+            runner: ubuntu-24.04
+            tox-env: py312
+    runs-on: ${{ matrix.runner }}
     steps:
-      - name: Install dependencies
-        run: sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox
       - name: Git checkout
         uses: actions/checkout@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
blackboxsw: For all workflow uses clauses, please use full SHA commits instead of tag names to avoid future supply chain attacks.
v8.3.2 appears the be the latest release version so let's update to 

```suggestion
        uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990  # v8.3.2
```

blackboxsw: We can put AI up to updating all workflow pinning to git commit SHAs for other actions as a followup PR. But, since we are introducing new dependencies, let's use a pinning practice which avoids tags.

blackboxsw (commented):
After testing with both ubuntu-pro-client and cloud-init integration tests I feel comfortable that this tooling update isn't going to negatively impact CI in either case.

Also integration test runs locally appear to [work fine for new PRs in pycloudlib with these new workflow changes](https://github.com/blackboxsw/pycloudlib/pull/2).

While generally I wouldn't want to replace tooling unless dependency feature deprecations are imminent or projects are becoming abandoned I do see an easy path toward other tooling integration with workshop as @rpocase mentioned (as well as faster dep install time uv vs pip). Given that pycloudlib import as set is being used so broadly by CPC, cloud-init and ubuntu-pro-client CICD, that time/setup saving adds up fairly quickly.


will follow up on get_ply and lxd

pyproject.toml
blackboxsw: Also missing a `requires-python >=3.8` in [project]

pycloudlib/lxd/instance.py
blackboxsw: @rpocase Sorry, I'm not getting what you mean here.  The [superclass explicitly supports exactly this same call signature with get_pty param](https://github.com/rpocase/pycloudlib/blob/dce86a2f2d00d2f6f80fb64f145f1583bcce4a92/pycloudlib/instance.py#L246). How is the subclass breaking that contract? I must be misreading something.

blackboxsw (commented):
Additionally I think we need to:
1. Adapt the `deps` target from docs/Makefile to avoid pip references to requirements.txt which is now deleted.
2. drop *requirements.txt from MANIFEST.in

pycloudlib/lxd/instance.py
rpocase: the superclass supported the keyword argument, but the subclass didn't expose the argument. i wasn't sure if this was intentional or not.

rpocase:
@blackboxsw latest push addresses the commit pins and ensures that the recent azure-network-mgmt pin stays in place

pyproject.toml
```diff
@@ -1,6 +1,94 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+    "Topic :: Utilities",
+]
+dependencies = [
+    "azure-cli-core >= 2.21.0",
+    "azure-identity",
+    "azure-mgmt-compute >= 17",
+    "azure-mgmt-network <= 31",
```
blackboxsw: This should be `>= 16, < 31` not `<= 31`.

```suggestion
    "azure-mgmt-network >=16, < 31",
```

blackboxsw (changes requested):
Thanks @rpocase looks good with two exceptions:
1. pinning correction on azure-mgmt-network
2.  I don't think we want to commit `uv.lock` to this project repo do we? I may be mistaken, but thought when you have a library you are building, you may not want to provide uv.lock because that library may be used in other environments which may be created in other releases of Ubuntu where the python versions of dependencies may differ from our upstream repo.   Again, I may be mistaken as I haven't used uv too often with distributed libraries.

If you think we need to keep uv.lock, how do you envision cloud-init, cpc tests, ubuntu pro attempting to run those project test suites. do they perform a `uv sync` to get the same uv.lock version from pycloudlib prior to running tests?

The pinning suggestion doesn't matter for too long as I just put up https://github.com/canonical/pycloudlib/pull/516 which may help drop this pin in favor of the new Azure SDK model-based hybrid requests.  I'll rebase 516 after this PR lands to fix the pinning minor version and avoid delaying your work any longer.

rpocase:
> Thanks @rpocase looks good with two exceptions:
> 
> 1. pinning correction on azure-mgmt-network

ack, will wait on the rebase. 

> 2. I don't think we want to commit `uv.lock` to this project repo do we? I may be mistaken, but thought when you have a library you are building, you may not want to provide uv.lock because that library may be used in other environments which may be created in other releases of Ubuntu where the python versions of dependencies may differ from our upstream repo.   Again, I may be mistaken as I haven't used uv too often with distributed libraries.

uv.lock is for the developers of the project, not the consumers. the lock file ensures that the developer starting point is consistent. end-users installing the project will receive their dependency bounds from pyproject.toml.

> If you think we need to keep uv.lock, how do you envision cloud-init, cpc tests, ubuntu pro attempting to run those project test suites. do they perform a `uv sync` to get the same uv.lock version from pycloudlib prior to running tests?


see above

blackboxsw:
Thanks for the explanation @rpocase.

I think last things to resolve after rebase:
- align pinning (or lack thereof) for azure-mgmt-* dependencies
- drop or adapt any `pip install` references from docs/* to `uv pip install`

then let's ship it. Thanks again. I'll rebase the other PRs I have open after your 507 lands.

rpocase:
> * align pinning (or lack thereof) for azure-mgmt-* dependencies

done!

> * drop or adapt any `pip install` references from docs/* to `uv pip install`

the last reference should remain. end users can still use pip to install pycloudlib (or use their package manager from their project).

.github/workflows/ci.yaml
```diff
@@ -10,56 +10,61 @@ jobs:
   linting:
     runs-on: ubuntu-24.04
     steps:
-      - name: Install dependencies
-        run: sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox
       - name: Git checkout
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262  # v4
```
blackboxsw, line 14: I'm not sure why we are pinning an older image stream commit for checkout. Can we either retain v7.0.0 or bump this to v7.0.1?

.github/workflows/ci.yaml
```diff
@@ -10,56 +10,61 @@ jobs:
...
+          - python: "3.12"
+            runner: ubuntu-24.04
+            tox-env: py312
+    runs-on: ${{ matrix.runner }}
     steps:
-      - name: Install dependencies
-        run: sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox
       - name: Git checkout
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262  # v4
```
blackboxsw, line 33: Again we are reducing our workflow pinning from v7 to v4. Let's keep the recent releases unless there is justification for pinning and older release stream

.github/workflows/ci.yaml
```diff
@@ -10,56 +10,61 @@ jobs:
   linting:
     runs-on: ubuntu-24.04
     steps:
-      - name: Install dependencies
-        run: sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox
       - name: Git checkout
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262  # v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e  # v6
```
blackboxsw, line 16: Note astral-sh current release is v9. Any reason we don't want to pin against latest release of this action?

.github/workflows/ci.yaml
```diff
@@ -10,56 +10,61 @@ jobs:
...
         run: |
           sudo apt-get update -q
-          sudo apt-get install -qy distro-info tox
+          sudo apt-get install -qy distro-info
           sudo snap install lxd
       - name: Git checkout
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        uses: actions/checkout@a37ce9120846195fa4ece8f58b268e6043cb2f26  # v3
+      - name: Install uv
+        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e  # v6
```
blackboxsw, line 60: I see v3 instead of the earlier v4 for checkout here. Let's make sure we use the same version for all actions throughout our workflows.

pycloudlib/lxd/instance.py
blackboxsw: Ahh thanks for clarification and the correction. I think plumbing it into the subclass signature makes sense. This looks correct to me.

blackboxsw (approved):
LGTM, minor nits to update the pinned SHAs for the setup* actions. Let's ship this.

PR #508, opened 2026-06-26
feat: upgrade openstacksdk

rpocase:
## PR Checklist

request.

- [-] (if applicable) I have added a reference to issues that this PR relates to in the PR message (Refs GH-1234, Fixes GH-1234)

## Description
this migrates openstacksdk to a modern modern version. the intention is
to drop netiface from the dependency tree because the package is no                                                   
longer maintained   

this also drops python 3.8, 3.9 support. this started as a means to get to an openstack version that
did not include netiface as a dependency. after that, i noticed python 3.8 + 3.9 went EOL and are not tested in CI. to get to a version of openstacksdk without inetface it required dropping 3.8, 3.9.

this builds off of #507 

## Test Steps

N/A


pyproject.toml
```diff
@@ -37,11 +35,11 @@ dependencies = [
     "ibm-vpc >= 0.10, < 0.29.0",
     "knack >= 0.7.1",
     "oci >= 2.17.0",
-    "openstacksdk >= 1.1.0, < 1.5.0",
+    "openstacksdk >= 4.2.0, < 4.9.0",
```
blackboxsw, line 38: This is a huge bump in openstackSDK, can we provide a separate commit to all self.conn changes to align with latest openstacksdk API changes?

It looks to me like we'd expect to see methods hung off of separate `self.conn.compute`, `self.conn.network` or `self.conn.image`. For example,  instead of top-level `self.conn.create_floating_ip` we should be able to use the `self.conn.network.create_ip` method.

blackboxsw (changes requested):
Thank you @rpocase cloud-init is +1 on the breaking change w/ floor of python support at 3.10 for all our test harnesses.  With the bump in openstacksdk, I'd like to see us also adapt to the new [Compute](https://docs.openstack.org/openstacksdk/latest/user/proxies/compute.html), [Network](https://docs.openstack.org/openstacksdk/latest/user/guides/network.html)  connection objects to insulate pycloudlib from future SDK changes and eventual deprecation of legacy top-level conn methods.

blackboxsw (changes requested):
I believe this changeset may break openstacksdk as APIs have changed to Connection-based objects conn.network conn.compute etc.

Can we get a validation that the following works when you have a sourced valid .novarc?
```
openstack image list # determine the ID  your want to launch
IMAGE_ID=<IMAGE_ID>
tox -e pytest
source .tox/.testenv/bin/activate
python3 ./examples/openstack_example.py $IMAGE_ID
```

NOTE: I ran into issues that openstack doesn't like the keyname pycloudlib comes up with , which is our corparate-laptop username. I think we may need to replace any `@` and `.` with hyphens before trying to upload keys to openstack from pycloudlib.

NOTE: I also had to add a specific instance_type to examples to align with my current openstack instance I had access too. From: `openstack flavor list`

```diff
diff --git a/examples/openstack_example.py b/examples/openstack_example.py
index fe8657e..4458d18 100644
--- a/examples/openstack_example.py
+++ b/examples/openstack_example.py
@@ -14,7 +14,7 @@ REQUIRED_ENV_VARS = ("OS_AUTH_URL", "OS_PASSWORD", "OS_USERNAME")
 def basic_lifecycle(image_id: str):
     """Demonstrate basic set of lifecycle operations with OpenStack."""
     with pycloudlib.Openstack("pycloudlib-test") as os_cloud:
-        with os_cloud.launch(image_id=image_id) as inst:
+        with os_cloud.launch(instance_type="production-cpu2-ram4-disk20", image_id=image_id) as inst:
             inst.wait()
```

tox.ini
```diff
@@ -18,12 +18,6 @@ commands = {envpython} -m pytest --doctest-modules --cov=pycloudlib --cov-branch
 [testenv:py38]
 envdir = {[common]envdir}
 deps = {[common]deps}
-basepython = python3.8
```
blackboxsw, line 21: Can we retain an explicit  `[testenv:python10]` in tox.ini instead of just removing testenv:py38

rpocase:
@blackboxsw i could use some help testing. this is probably user error on my part, but i am hitting issues with networking finding an IP on the instance i create. i don't think there are actual source problems

rpocase:
@blackboxsw would you be able to help test these changes?

blackboxsw:
@rpocase. I'm having some issues with PS6 vm creation and launch ATM. Trying to reproduce this branch and ensure I can launch, snapshot and delete on OpenStack with this changeset. Initial instance launch appears to work, but the server that is launched doesn't appear to become accessible so I can validate the other shutdown/snapshot/delete operations. Will check on this tomorrow.

PR #518, opened 2026-07-22
dependabot batch: aggregate dependabot changes, setup-python, gh-action-pypi-publish, checkout

blackboxsw:
Consolidate all dependabot PRs #513 #514 and #515. Use individual commits

Will bump VERSION once we have approval.


blackboxsw:
Dropping in favor of #519

PR #524, opened 2026-07-31
fix(oci): raise ImageNotFoundError when daily image lookup has no match

blackboxsw:
## Description

Previously, daily_image raised an unhelpful IndexError when no matching image was found after filtering out aarch64 and GPU images. It now raises the more descriptive ImageNotFoundError carrying the release.

This issue was seen on July 31st before Canonical Ubuntu Oracle  Resolute images were made available. The resulting traceback was an unhelpful IndexError. This ImageNotFoundError will be easier to read and reason about in call sites.

## Test Steps
```
tox -e pytest
```


nryanl (approved):
Looks good. The change is clear and the tests appear to be enough for this change.

nryanl (changes requested):
Bump VERSION before merging, since another PR made it in first.

holmanb (approved):
Thanks!

nryanl (approved):

PR #528, opened 2026-09-08
fix(ibm): respect configured zone for custom VPC subnets

Gisaldjo:
Select a subnet in the explicit or configured zone, creating one there if needed. Preserve first-subnet selection when no zone is provided, and creation in <region>-1 when the VPC has no subnets.

## Description

Custom VPC subnet discovery currently selects the first subnet, even
when an explicit or configured zone is provided. In a multi-zone VPC,
this can select a subnet in a different zone from the requested
instance, causing instance creation to fail.

Select a subnet in the requested zone instead, creating one there if
none exists.

Preserve existing behavior when no zone is provided: select the first
subnet, or create one in `<region>-1` if the VPC has no subnets.

New zone-specific subnets include the zone in their names to avoid
collisions with existing subnets. Very long VPC names may exceed IBM's
subnet-name limit; name truncation is not addressed here.

## Test Steps

- All 34 IBM unit tests passed.
- Ruff lint and format checks passed.
- Scoped mypy, Python 3.10 tests, and the docs build passed before the
  final compatibility test was added.
- Live verification passed for:
  - Configured-zone selection when the first subnet is in another zone.
  - Instance launch and SSH access.
  - Omitted-zone first-subnet selection.
  - Empty-VPC subnet creation in <region>-1 when zone is omitted.
  - Creation in the configured zone when no matching subnet exists,
    preserving the existing subnet.
- Temporary live-test resources were deleted.
- Ran `tox`: Ruff passed and all 327 unit tests passed on Python 3.8.
  Mypy failed in unchanged Azure code at
  `pycloudlib/azure/instance.py:293` with a `begin_create_or_update`
  overload mismatch.


pycloudlib/ibm/instance.py
```diff
@@ -78,12 +78,20 @@ def from_default(cls, client: VpcV1, zone: str, vpc_id: str) -> "_Subnet":
...
             client.list_subnets,
             resource_name="subnets",
-            filter_fn=(lambda subnet: subnet["vpc"]["id"] == vpc_id),
+            filter_fn=(
+                lambda subnet: subnet["vpc"]["id"] == vpc_id
+                and (zone is None or subnet["zone"]["name"] == zone)
+            ),
         )
         if subnet is None:
             raise IBMException(f"No subnet associated to vpc found: {vpc_id}")
```
Chandan-M-N, line 97: Can we also print zone in the error message, if zone is not `None`? It will be easier for debugging.

pycloudlib/ibm/cloud.py
```diff
@@ -63,8 +63,9 @@ def __init__(
         )
         self._resource_group_id: Optional[str] = None
         self.region = str(region or self.config.get("region")).lower()
-        zone = zone or self.config.get("zone") or f"{self.region}-1"
-        self.zone = str(zone).lower()
+        configured_zone = zone or self.config.get("zone")
```
Gisaldjo, line 66: @blackboxsw I went back and forth on preserving the existing behavior where, without an explicit zone, pycloudlib selects the first available subnet. My current proposed change here would preserve existing behavior when no zone is explicitly provided by the user.

But I'm having second thoughts. I am leaning toward always filtering subnets based on self.zone. A subnet in a different zone cannot be used to create an instance in any case. When no zone is configured, pycloudlib already defaults self.zone to {region}-1; selecting a subnet in that zone seems more correct than selecting the first subnet returned by the API.

This would let us remove the _zone_provided / configured_zone workaround. What do you think?

PR #532, opened 2026-09-16
fix(azure): resolve mypy call-overload error on NIC creation

### LINKED ISSUE #531: ci lint errors seen against tip of main
Correct the following issues with CI against tip of main @ 3a97897232b89c9154e0175728807d2182cf80ef

```
pycloudlib/gce/cloud.py:318: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
pycloudlib/azure/instance.py:293: error: No overload variant of "begin_create_or_update" of "NetworkInterfacesOperations" matches argument types "Any", "str", "dict[str, dict[str, Collection[Collection[str]]] | str | None]", "str"  [call-overload]
pycloudlib/azure/instance.py:293: note: Possible overload variants:
pycloudlib/azure/instance.py:293: note:     def begin_create_or_update(self, resource_group_name: str, network_interface_name: str, parameters: NetworkInterface, *, content_type: str = ..., **kwargs: Any) -> LROPoller[NetworkInterface]
pycloudlib/azure/instance.py:293: note:     def begin_create_or_update(self, resource_group_name: str, network_interface_name: str, parameters: IO[bytes], *, content_type: str = ..., **kwargs: Any) -> LROPoller[NetworkInterface]
```

### Additional context
GH CI action run showing failure: https://github.com/canonical/pycloudlib/actions/runs/35011200328/job/104523299204?pr=530

8rulerstar:
## Description

`mypy pycloudlib examples` fails on one call: `network_interfaces.begin_create_or_update` is typed to take a `NetworkInterface` model or `IO[bytes]`, while this module builds and passes the ARM-shaped dict the REST API expects. The call is fine at runtime, so this narrows the ignore to that single overload error instead of changing what is sent.

I used `# type: ignore[call-overload]` to match the ignores already used a few lines above (268-270) in the same method. As mentioned in #531, I also have a `cast(NetworkInterface, default_config)` version ready — `NetworkInterface` is already imported here — so just say the word if you'd prefer that.

Verified locally with the versions CI uses (mypy 2.3.1, azure-mgmt-network 32.0.0, ruff 0.5.0):

- `mypy pycloudlib/azure/instance.py` — before: 1 error, after: `Success: no issues found`
- `ruff check pycloudlib examples` — `All checks passed!`
- `ruff format --check pycloudlib examples` — 64 files already formatted

One note on the issue text: `pycloudlib/gce/cloud.py:318` is a mypy *note*, not an error — it comes from `check_untyped_defs = false` for that module in the TODO override block — so this PR leaves it alone. Happy to drop that module from the override list in a separate PR if you want the bodies checked.

I did not run the unit tests locally (that needs the full cloud dependency set); this change adds no runtime behavior, so I left them to CI.

#### Issues

* Fixes GH-531


blackboxsw (changes requested):
Thank you for this submission and context related comment inline. I think what remains is bumping VERSION file so we can land this PR.

8rulerstar:
Bumped VERSION to 1!11.1.8. Thanks for the review!

blackboxsw (approved):
LGTM, awaiting  CI and then I'll merge this.

